PR #23, opened 2021-02-10
feat(application):window properties on linux

nklayman:
Waiting on #22, this adds support for all the config options I could get to work on linux.

### DIFF
--- a/src/application/gtkrs.rs
+++ b/src/application/gtkrs.rs
@@ -12,7 +12,9 @@ use std::{
 };
 
 use gio::{ApplicationExt as GioApplicationExt, Cancellable};
-use gtk::{Application as GtkApp, ApplicationWindow, ApplicationWindowExt};
+use gtk::{
+    Application as GtkApp, ApplicationWindow, ApplicationWindowExt, GtkWindowExt, WidgetExt,
+};
 
 pub struct Application<T> {
     webviews: HashMap<u32, WebView>,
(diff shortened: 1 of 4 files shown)


src/application/general.rs
```diff
@@ -68,6 +69,12 @@ impl<T> ApplicationExt<'_, T> for Application<T> {
         let window_attributes = WindowAttributes::from(&attributes);
         window_builder.window = window_attributes;
         let window = window_builder.build(&self.event_loop)?;
+        if attributes.x.is_some() && attributes.y.is_some() {
```
lucasfernog, line 72: When you use `is_some()` and then `unwrap()` inside the `if` condition, you're doing the same thing twice, which is checking whether the `Option` is `Some` or `None`. This would be more idiomatic if you wrote it like this:

```rust
match (attributes.x, attributes.y) {
    (Some(x), Some(y)) => {
		// set window properties here
	}
    _ => {}
}
```

nklayman: I didn't write that, it was part of #22. I used that as a base so it showed up here.

src/application/gtkrs.rs
```diff
@@ -73,9 +75,21 @@ impl<T> ApplicationExt<'_, T> for Application<T> {
         })
     }
 
-    fn create_window(&self, _attributes: AppWindowAttributes) -> Result<Self::Window> {
-        //TODO window config
+    fn create_window(&self, attributes: AppWindowAttributes) -> Result<Self::Window> {
+        //TODO window config (missing transparent, mix/max height/width, x/y position)
```
nothingismagick, line 79: Don't forget to update your todo.

src/application/general.rs
amrbashir: Thanks for the great advice, I barely know rust so my most of the time my code won't be idiomatic.

lucasfernog: most of the time my code isn't idiomatic either LOL some code reviews just destroy me

PR #26, opened 2021-02-11
fix(app/linux): close app when all windows are closed

lucasfernog:


nklayman:
Oh nice, I noticed that but never got around to fixing it.

nklayman:
Yeah it's much better than windows-rs lol

lucasfernog:
Yeah, you can know exactly how it works even by reading a C/Python/Go example.

PR #27, opened 2021-02-11
feat: more window configs: `icon` (winit/gtk-rs) and `fullscreen` (gtk-rs only)

amrbashir:
`load_icon` function is taken directly from winit's window icon [example](https://github.com/rust-windowing/winit/blob/master/examples/window_icon.rs) it uses [image](https://github.com/image-rs/image) to convert the image to `rgba Vec` so I added it as a dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,18 +21,19 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 thiserror = "1.0"
 url = "2.2"
+image = "0.23"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 webkit2gtk = { version = "0.11", features = ["v2_8"] }
 gio = "0.9"
 gtk = "0.9"
 gdk = "0.13"
+gdk-pixbuf = "0.9"
 glib = "0.10"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 windows = "0.2"
 winit = "0.24"
-image = "0.23.13"
 
 [target.'cfg(target_os = "windows")'.build-dependencies]
 cc = "1"
@@ -43,5 +44,3 @@ cocoa = "0.24"
 core-graphics = "0.22"
 objc = "0.2"
 winit = "0.24"
-image = "0.23.13"
-
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ glib = "0.10"
 [target.'cfg(target_os = "windows")'.dependencies]
 windows = "0.2"
 winit = "0.24"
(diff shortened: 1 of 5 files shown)


nklayman:
How would icon loading work in Tauri where all assets are bundled into the executable? Would we have to keep an icon file outside of the executable for it to work?

lucasfernog:
I think we should have two icon options: one taking a `Icon` instance and one taking the icon path.

lucasfernog:
Also, we should look on the other webview to see how to load the default app icon by default.

lucasfernog:
I think that's only a Windows thing: https://github.com/webview/webview/blob/ce27be3bc811688b04303fcab68fd76e4709f73a/webview.h#L1053

nklayman:
Isn't `Icon` platform specific? We would have to just pass a buffer of bytes or something since gtkrs uses the `PixBuff` type instead.

lucasfernog:
yeah I missed it - we should pass a Vec<u8> instead.

lucasfernog:
maybe we should define an `Icon` enum on the application layer:

```rust
enum Icon {
  File(PathBuf),
  Raw(Vec<u8>),
}

struct AppWindowAttributes {
  icon: Option<Icon>,
}
```

then each implementation (winit and gtk) handles the enum on its own way.

amrbashir:
Alright, will do it when I get back.

amrbashir:
There's something that i am not sure how to handle.
if the icon is of `Icon::Raw` how will we determine the width and height of the icon to pass it to winit's `Icon::from_rgba`.
>```
>pub fn from_rgba(
>    rgba: Vec<u8>,
>    width: u32,
>    height: u32
>) -> Result<Self, BadIcon>
>```
> 
> Creates an Icon from 32bpp RGBA data.
> 
> The length of rgba must be divisible by 4, and width * height must equal rgba.len() / 4. Otherwise, this will return a BadIcon error.
>
>The length of rgba must be divisible by 4, and width * height must equal rgba.len() / 4. Otherwise, this will >return a BadIcon error.

lucasfernog:
Just use the image crate too, it accepts raw byte array I believe.

amrbashir:
I can't find some thing that only accepts a raw byte, most of them require width and height too. I'll keep digging through the docs, maybe I missed it.

amrbashir:
also the gtk `Pixbuf` needs width and height. I guess I am stuck, if some knows which methods I should be using, that would be helpful.

lucasfernog:
I mean the `image` crate requires only the raw bytes, not the `Icon` struct from winit.

PR #28, opened 2021-02-11
fix(webview): callbacks should be tied to a window, add build CI

lucasfernog:


src/platform/macos.rs
```diff
@@ -102,6 +125,7 @@ impl InnerWebView {
             let cls = ClassDecl::new("WebViewDelegate", class!(NSObject));
             let cls = match cls {
                 Some(mut cls) => {
+                    decl.add_ivar::<WindowId>("_window_id");
```
wusyong, line 128:
```suggestion
                    cls.add_ivar::<WindowId>("_window_id");
```

PR #33, opened 2021-02-13
feat: add `skip_taskbar` window config

amrbashir:


amrbashir:
forgot to open it as a draft

amrbashir:
I added some code to close the hidden window when the visible one closes but I have no way to test if it actually works.

If someone knows a way to test it, let me know becuase TaskManager only show one window under `wry.exe` and it is the visible one. 

This piece of code  https://github.com/tauri-apps/wry/blob/ecb7a78aa915c0abb3a2fb4617bb350349efc518/src/application/general.rs#L168-L175 
will ensure to close the process when all windows are closed which is good but if I have two windows and I keep closing a window and open another, there would be a a lot of windows that won't be closed until the process is closed. 

The last commit should theoritically close it but as I said have no way of testing it.

wusyong (approved):

PR #35, opened 2021-02-13
Fix dpi scaling on macOS

wusyong:
@lucasfernog Could you test does this work on your mac mini?


lucasfernog:
I'll try it as soon as I get home, but it looks promising.

lucasfernog:
image

Seems to be working fine, but the right side has some padding for the scrollbar.

PR #36, opened 2021-02-14
feat: dynamic window creation

amrbashir:


src/application/general.rs
```diff
@@ -95,54 +101,31 @@ impl<T> ApplicationExt<'_, T> for Application<T> {
...
-        Ok(())
+    ) -> Result<WindowId> {
+        self.dispatcher()
+            .dispatch_message(Message::App(AppMessage::NewWindow(
+                window_attrs,
+                webview_attrs,
+                callbacks,
+            )))
+            .unwrap();
+        Ok(self.receiver.recv().unwrap())
```
amrbashir, line 128: The windows are all created fine and as expected without this line.

amrbashir, line 128: The windows are all created fine and as expected without this line.

wusyong (approved):

PR #55, opened 2021-02-21
chore: covector init

jbolda:
🎉 

Note: There is an added option that you can specify `housekeeping` and it will effectively no-op on it. This way we can always require a change file.

## TODO
- [ ] enter the `crate_token`

### DIFF
--- a/.github/workflows/covector-version-or-publish.yml
+++ b/.github/workflows/covector-version-or-publish.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: cargo login
-        run: cargo login ${{ secrets.crate_token }}
+        run: cargo login ${{ secrets.ORG_CRATE_TOKEN }}
       - name: git config
         run: |
           git config --global user.name "${{ github.event.pusher.name }}"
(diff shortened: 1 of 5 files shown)


nothingismagick:
@jbolda - we have it already at org level.
image

However, @wusyong will have to add the tauri-bot user to the crate or we won't be able to push.

.github/workflows/covector-version-or-publish.yml
```diff
@@ -0,0 +1,46 @@
...
+      change: ${{ steps.covector.outputs.change }}
+      commandRan: ${{ steps.covector.outputs.commandRan }}
+      successfulPublish: ${{ steps.covector.outputs.successfulPublish }}
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: cargo login
+        run: cargo login ${{ secrets.crate_token }}
```
jbolda, line 22: @nothingismagick I assume that means we need to switch this?

```suggestion
        run: cargo login ${{ secrets.ORG_CRATE_TOKEN }}
```

wusyong:
I've added tauri bot to the crate owner. This means we can merge it right?

wusyong (approved):

jbolda:
> I've added tauri bot to the crate owner. This means we can merge it right?

Aye, I think we are ready to see it fly.

PR #63, opened 2021-02-23
Use IIFE for injected javascript.

tmpfs:
Otherwise the global `name` variable is overriden by the last assigned callback and all function invocations target the last defined callback.

This fixes a bug whereby using multiple callbacks when calling `add_window` like this:

```rust
app.add_window(attrs, Some(vec![ipc, log_error, log_info, log_warn]))?;
```

Without this fix all javascript function invocations call the `Callback` named `log_warn` (the last one defined); after applying this fix each `Callback` should be invoked as expected.


lucasfernog (approved):

lucasfernog:
We should have added a change file on this PR, totally forgot that we have covector now :|

PR #65, opened 2021-02-23
Add custom protocol

wusyong:


Cargo.toml
```diff
@@ -22,20 +22,22 @@ serde_json = "1.0"
 thiserror = "1.0"
 url = "2.2"
 image = "0.23"
+mime_guess = "2.0.3"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 cairo-rs = "0.9"
 webkit2gtk = { version = "0.11", features = ["v2_8"] }
 gio = "0.9"
+glib = "0.10.3"
```
nklayman, line 31: Should I undo that move or does it not matter?

index.html
```diff
@@ -0,0 +1,5 @@
+<html>
+    <body>
+        WRY IS AWESOME
+    </body>
+</html>
```
nklayman, line 5: This file shouldn't have been committed, whoops.

src/platform/win.rs
```diff
@@ -92,6 +95,43 @@ impl WV for InnerWebView {
...
+                    custom_protocol_name = Some(protocol.0.clone());
+                    w.add_web_resource_requested_filter(
+                        &format!("file://custom-protocol-{}*", protocol.0),
+                        webview2::WebResourceContext::All,
+                    )?;
+                    w.add_web_resource_requested(move |_, args| {
+                        let uri = args.get_request().unwrap().get_uri().unwrap();
+                        // Remove leading custom protocol indicator
+                        let path = &uri[(23 + protocol.0.len())..];
+                        match protocol.1(path) {
```
nklayman, line 111: I should destructure the `if let` so there isn't any `.n` stuff.

nklayman (changes requested):
Feels odd to request changes on my own PR but I'm too tired to actually make them RN :joy:.

Cargo.toml
wusyong: no need, "0.10.3" actually means "0.10.3+" so I think it's the same effect to "0.10"

wusyong:
No worry. I'll make the change and then merge it.

src/platform/win.rs
wusyong: I'm outside without Windows. I'll make sure to do it back home.

wusyong: I'm outside without Windows. I'll make sure to do it back home.

PR #69, opened 2021-02-23
fix set_decorations

jussiniinikoski:


nothingismagick:
Thanks for noticing this! I just reviewed the file, and it seems there is one more place where its wrong:
```rust
    pub fn set_transparent(&self, resizable: bool) -> Result<()> {
        self.proxy.send_message(Message::Window(
            self.id,
            WindowMessage::SetResizable(resizable),
        ))
    }
```

jussiniinikoski:
Looks like setting transparency is currently only implemented using attributes..

jussiniinikoski:
I could try to fix this, perhaps a separate PR?

nothingismagick:
Let's wait for @wusyong to weigh in here.

wusyong:
@jussiniinikoski another PR is good to me. The actual working transparent option has to wait for next version anyway.

wusyong (approved):

wusyong:
@jussiniinikoski Sorry I forgot to tell you more context. Right now the transparent option can only be set on initialization so it shouldn't have a method in the proxy. We should remove those lines actually.

jussiniinikoski:
Ok, thanks for info.

PR #72, opened 2021-02-23
fix/examples

amrbashir:


nklayman (approved):
Lol I think I broke all these with the custom protocol PR. Thanks for fixing!

amrbashir:
No problem and I apparantly mixed two PR's LOL

PR #75, opened 2021-02-25
fix: update format of path passed to custom protocol resolver

nklayman:
now it is consistent cross-platform


nklayman:
Good thing we have tests since I apparently forgot how to code :facepalm:.

nklayman:
Ok hopefully that does it

wusyong (approved):

PR #82, opened 2021-02-26
Better mime handling. This time for real

wusyong:


nothingismagick (approved):
Tested winit example on macos m1

lucasfernog:
We should probably move this to a separate function in the future. But this is good enough for now.

lucasfernog (approved):

nothingismagick:
> We should probably move this to a separate function in the future. But this is good enough for now.

Do you mean since it is literally exactly the same code repetition in three files?

lucasfernog:
> > We should probably move this to a separate function in the future. But this is good enough for now.
> 
> Do you mean since it is literally exactly the same code repetition in three files?

Yeah :)

nklayman (approved):
Looks good but I do think this should be put in a separate function at some point, it'll be a good PR for new contributors to Wry.

Issue #66, opened 2021-02-23
Consider bi-directional RPC communication


tmpfs:
**Is your feature request related to a problem? Please describe.**
Not a problem but I think the ergonomics of the `Callback` API could be improved or potentially replaced with a higher-level RPC API.

**Describe the solution you'd like**
As I call `wry` callbacks from Javascript I don't want to define a new `Callback` for every function and I want to transparently return data to my Javascript code via Promises. I am currently writing a wrapper on top of the `Callback` API that achieves this using `JSON-RPC` but it feels icky because `wry` is already using `JSON-RPC` under the hood so I am effectively embedding a `JSON-RPC` message in the parameter of a `JSON-RPC` message.

I would like to propose a solution that uses a `Service` trait instead of a `Callback` function and RPC messages are handled transparently using a simple embedded Javascript client (`window.rpc` maybe?). The Javascript client would expose two public functions `call()` to call out to a service and receive a reply (via a `Promise`) and `send()` which would **not** include the `id` and therefore not receive a response from the `Service` implementation(s). 

**Describe alternatives you've considered**
I have a working prototype but I think this is such a common requirement that it should be part of `wry`.

**Would you assign yourself to implement this feature?**
- [x] Yes (Linux only)
- [ ] No

**Additional context**
The existing Rust `JSON-RPC` libraries are not a good fit for various reasons so I put together this [json-rpc2](https://docs.rs/json-rpc2/) crate that I am using for the service implementation.

Happy to share more if this interests you.

wusyong:
`Callback` and its implementation are inspired by the original webview repo. I can't say I'm happy with the interface either.
In general, I'm glad to see any improvement toward it as long as we still communicate via message handler from each webview platform.
I think the Service trait seems pretty promising, but I'm not the one eating the dog food. Perhaps @lucasfernog and @nklayman  you guys have more feedbacks about this?

tmpfs:
Great, I think this can be done with a view to augmenting the existing `Callback` mechanism. Maybe an enum for `Function`: 

```rust
let rpc = Callback {
  name: "rpc".to_owned(),
  function: Function::Rpc { /* Service handlers go here */ },
}
```

Or for a callback like the existing logic:

```rust
let rpc = Callback {
  name: "callback".to_owned(),
  function: Function::Call(/* boxed closure like we do now */),
}
```

Then when callback functions are invoked it's just a `match` and service handlers are invoked using the RPC request whilst `Function::Call` follows the existing code path.

[...]

PR #83, opened 2021-02-27
Rpc example (work in progress)

tmpfs:
Added an example app to test the work on the new RPC API.

Includes addition of `urlencoding` to the `dev-dependencies` so that we can embed simple test pages in the examples using `data:` URLs.

Very much work in progress but would like to get this merged now so rebasing is not too painful, cheers!


wusyong:
@tmpfs I'm going to turn this into a draft. After you feel comfortable about it, you can turn it back to PR and I'll start reviewing it.

You only need to implement it on Linux. It's fine to ignore other platforms for now.

tmpfs:
Hi @wusyong, just replied to the [original issue](https://github.com/tauri-apps/wry/issues/66#issuecomment-787637800) and I think there is enough here for you to review.

Would be great to get this merged if you approve!

Thanks! :+1:

src/platform/linux.rs
```diff
@@ -45,28 +52,94 @@ impl WV for InnerWebView {
...
-                                format!(
-                                    r#"window._rpc[{}].resolve("RPC call success"); window._rpc[{}] = undefined"#,
-                                    v.id, v.id
-                                )
+                        match serde_json::from_str::<FuncCall>(&js) {
+                            Ok(mut ev) => {
+                                // Use `isize` to conform with existing `Callback` API but should 
+                                // really be a `u64`. Note that RPC spec allows for non-numbers 
+                                // in the `id` field!
+                                let id: i32 = if let Some(value) = ev.payload.id.clone().take() {
```
wusyong, line 60: Let's make Callback use u64 then.

src/webview.rs
```diff
@@ -20,6 +21,9 @@ use winit::platform::windows::WindowExtWindows;
 #[cfg(not(target_os = "linux"))]
 use winit::window::Window;
 
+//pub type RpcProxyHandler = Box<dyn Fn(&Dispatcher, RpcRequest) -> Option<RpcResponse> + Send + Sync>;
```
wusyong, line 24: Remove this line

src/webview.rs
```diff
@@ -136,6 +152,59 @@ impl WebViewBuilder {
...
+                    if(this._promises[id]){
+                        this._promises[id].resolve(result);
+                        delete this._promises[id];
+                    }
+                }
+
+                // Call remote method and expect a reply from the handler
+                this.call = function(method) {
+                    const id = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
+                    const params = Array.prototype.slice.call(arguments, 1);
```
wusyong, line 182: Does this mean it only takes one argument to RPC handler?

wusyong (changes requested):
Hey I only left some minor change requests because I think the process of callback and rpc handler can still be the same. But this feels like another big change, I can merge and explore more once you make the below changes.

src/lib.rs
```diff
@@ -91,7 +91,8 @@ pub use application::{
     WindowMessage, WindowProxy,
 };
 pub use serde_json::Value;
-pub(crate) use webview::{Dispatcher, WebView, WebViewBuilder};
+pub use platform::{RpcRequest, RpcResponse};
+pub(crate) use webview::{Dispatcher, WebView, WebViewBuilder, RpcHandler};
```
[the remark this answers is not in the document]
tmpfs: Sure thing, I'll do that shortly.

src/platform/linux.rs
tmpfs: I can do this temporarily but I think I have a better idea. Bear with me - let's completely remove *all the calback code* and let people do it in Javascript:

```javascript
function someRpcMethod() {
  return window.rpc.call(arguments.callee.name, Arguments.prototype.slice.call(0));
}
```

That mimics what the `Callback` API gives us over the RPC channel! I think this is much more elegant.

As a convenience we could offer:

```rust
fn set_rpc_aliases( names: Vec<&str> );
```

Which would define those Javascript functions automatically.

src/webview.rs
tmpfs: Nope, that is all remaining arguments after the `method` argument.

src/platform/mod.rs
```diff
@@ -42,9 +44,62 @@ pub(crate) static CALLBACKS: Lazy<
     Mutex::new(m)
 });
 
+#[deprecated]
```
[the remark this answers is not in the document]
tmpfs: But if I do that it will break the Windows/Mac builds. Maybe leave it in for now and clean up later when we have arrived at a good solution?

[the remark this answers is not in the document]
wusyong: sure

src/platform/linux.rs
wusyong: That sounds pretty great!

src/lib.rs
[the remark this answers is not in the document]
tmpfs: @wusyong I wasn't entirely sure what you wanted here, but it looks like `platform` is a private module and `application` is for public facing code so I just moved the RPC related structs to the `application` module and re-export from there. Hope that is what you wanted!

tmpfs:
Hi @wusyong I am keen to get this completed. If you could merge this as is I will start work on clean up and removing all the obsolete stuff like `RPC`, °Callback`, `FuncCall` etc. I would also like to move the RPC handling code to a function that can be shared across the different platforms, it could just take the js string and the `RpcHandler` reference and return a `String` to be run as javascript. There is also a small bug where we need to ensure the promise is deleted when the RPC handler returns `None` if the request contains an `id`.

Thanks for your time and looking forward to getting this feature stable :+1:

wusyong (approved):
Ah sorry, I thought you mean to push more commits on this one. I can merge this real soon since the API isn't broke. 
I will create another branch to test some more and implement them on mac and windows. So maybe merge against it in next PR.

tmpfs:
Thanks @wusyong, that's create! Yeh, I am starting on a cleanup and improvement in a new branch.

Unfortunately, it looks like MacOS will require an `&'static` reference so we can get the rpc handler into the `fn did_receive()`. Going to do the clean up and then will migrate `RpcHandler` to use `OnceCell` for MacOS support.

PR #92, opened 2021-03-01
feat/ draggable-region

amrbashir:
I add a custom titlebar example mainly to test drag.

winit implementation will have to wait until this [PR](https://github.com/rust-windowing/winit/pull/1840) is merged.

### DIFF
--- a/.changes/draggable-regions.md
+++ b/.changes/draggable-regions.md
@@ -1,0 +1,5 @@
+---
+"wry": minor
+---
+
+Add draggable regions, just add `drag-region` class to the html element.
--- a/.changes/draggable-regions.md
+++ b/.changes/draggable-regions.md
@@ -1,5 +1,5 @@
 ---
-"wry": minor
+"wry": patch
 ---
 
 Add draggable regions, just add `drag-region` class to the html element.
--- a/src/application/gtkrs.rs
+++ b/src/application/gtkrs.rs
@@ -294,6 +294,9 @@ impl App for InnerApplication {
                 WindowMessage::EvaluationScript(script) => {
                   let _ = webview.dispatch_script(&script);
                 }
+                WindowMessage::BeginDrag { x, y } => {
+                  window.begin_move_drag(1, x as i32, y as i32, 0);
+                }
               }
             }
           }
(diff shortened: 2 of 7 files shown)


nklayman:
Awesome! This is super great to have.

src/application/gtkrs.rs
```diff
@@ -395,22 +399,39 @@ fn _create_webview(
...
-                        window_id,
-                    ),
-                    seq,
-                    req,
-                )
-            });
-        }
+
+    let begin_drag_callback = Callback {
+        name: "beginDrag".to_owned(),
```
nklayman: Should this be camel or snake case? On Tauri we decided that user-defined commands would be snake case, so it might be best to do that here as well.

nklayman:
I tested this out and it worked really well except for two things:
1. In the example, the maximize button won't unmaximize
2. When dragging the window, when I start moving the cursor the window jumps down by about 20px. For reference, I'm on Ubuntu 20.10 with Gnome 3.38.3 and nvidia drivers.

src/application/gtkrs.rs
amrbashir: yeah I think that's better, I also want to rename it to something that is unlikely to be overridden by users. should I rename it to `wry_invoke_begin_drag`

nklayman: Also, it might be best to name it `__WRY_BEGIN_DRAG__` as to not interfere with another `beginDrag` function, which doesn't seem all that uncommon to me.

amrbashir: yeah that's better

nklayman: Whoops, your comment didn't load till after I posted mine. Looks like we are thinking along the same lines though :joy:. Either name would be fine, but the general convention is to use `__` around all-caps I think.

amrbashir:
1. Since there is no getters implemented on the `WindowProxy` I can't get the window state (maximized or not to fire apropriate event)
2. I thought I fixed that last night, I will have to recheck it

amrbashir:
I just checked it again and the 20px jump doesn't happen to me, I am using AwesomeWM though not Gnome or Kde so that doesn't count I guess.

Not sure why it jumps 20px on your end,Last night I had the 20px jump but I managed to fix it by passing `clientX` and `clientY` from the mouse event to the callback then added these values on top of current window position

amrbashir:
@nklayman can you try with the latest commit ? hopefully it is fixed

nklayman:
Seems like that did it, nice work!

amrbashir:
Nice, we will have to wait for winit's PR to be merged tho.

nothingismagick:
Didn't the PR get merged already @amrbashir 
https://github.com/rust-windowing/winit/pull/1840

amrbashir:
Yeah it is merged but winit hasn't released a new version yet so I can't finish the PR unless they publish or wry uses winit from a git rev

wusyong:
I guess we can wait for next winit release? The PR can leave as is until then.

amrbashir:
Yeah it can wait, I just wanted to finish the PR while I can.

wusyong:
@amrbashir Could you update the winit dependency to the commit that has this feature?
I think we'll fork it and publish a non-aggressive crate version for tauri beta temporarily.

nothingismagick:
FYI: https://github.com/tauri-apps/winit

.changes/draggable-regions.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": minor
```
wusyong: LGTM! Could you update with `patch` since this is not a breaking change?

amrbashir: I thought `major` is for breaking changes and  `minor` for new features and `patch` for bugfixes.

wusyong: yeah, but that usually means [main version](https://semver.org/#spec-item-6) is > 0. With the main version under 1 in this crate, I wish to update the minor version only when there are breaking changes. This should be easier for users to work with without constantly updating their (minor) version

amrbashir: alright, let me change it

wusyong (approved):

PR #93, opened 2021-03-03
Rpc api cleanup

tmpfs:
This removes the obsolete callback API completely along with several other improvements and bug fixes.

Outstanding tasks to be done after this is merged:

1) Use `&'static` reference so `RpcHandler` can be exposed to the MacOS platform code
2) Add a `set_callbacks` function to alias Javascript functions to RPC method names (mimics the callback API)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,46 @@ cargo run --example multiwindow
 
 For more information, please read the documentation below.
 
+## Rust <-> Javascript
+
+Communication between the host Rust code and webview Javascript is done via [JSON-RPC][].
+
+Embedding code should call `set_handler()` on the application to register an incoming request handler and reply with responses that are passed back to Javascript. On the Javascript side the client is exposed via `window.rpc` with two public methods
+
+1. The `call()` function accepts a method name and parameters and expects a reply.
+2. The `notify()` function accepts a method name and parameters but does not expect a reply.
+
+Both methods return promises but `notify()` resolves immediately.
+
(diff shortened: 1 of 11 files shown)


tmpfs:
Hi @wusyong , sorry the history is a mess, I had problems rebasing after making these changes and had to resolve the conflicts manually. Hopefully, i have sorted it now.

I think this is more what we are aiming for with a lot of the old code removed and everything hanging off this thin RPC wrapper. The `rpc` example should work on Linux and if we are really lucky on Windows too; for MacOS support I need to make it `&'static` so it can be accessed in `did_receive()`. The logic has been moved to `platform::rpc_proxy()` so we can share it between platforms easily.

Look forward to your feedback, cheers :+1:

wusyong:
Hey sorry, I didn't realize your already start a draft. I'm going to add some changes regarding restructuring modules.
Could you clean up a bit on the commit history and merge against the rpc branch?
And yes currently it looks like rpc handler won't make it in mac since it's not a closure.
I'm looking for ways to achieve it.

tmpfs:
Ok @wusyong I am squashing the old branch and going to close this and open a new PR shortly.

Think I have a quick hack for the MacOS issue.

wusyong:
@tmpfs hey I update the signature of RPC handler quite a bit because I wish it can be more general. As for window proxy, we should add another type like `Callback` in the application module does.

Issue #39, opened 2021-02-17
Add examples


wusyong:
We want several examples to showcase how wry can do.

This can also help us writing mcve when there's a bug report issue.

PRs are welcome!

wusyong:
We probably also need a README in example directory to demonstrate each example.

wusyong:
Custom protocol feature just added! We need an example for it. Feel free to open a creative PR about it!

wusyong:
Closing it since I believe we have plenty of examples. But feel free to open any PR if anyone wants to add more examples.

PR #95, opened 2021-03-03
Implement draft RPC API.

tmpfs:
Remove old Callback mechanism.

* Remove obsolete Callback API
* Remove FuncCall and RPC
* Update README
* Rename set_rpc_handler() to set_handler()
* Use shared rpc_proxy() function for platform consistency
* Improve handling of promise cleanup
* Update README with RPC API info.
* Panic if webview handler is set after window creation.
* Improve rpc_proxy() logic, try to ensure any corresponding promise is always removed.
* Remove FuncCall wrapper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,46 @@ cargo run --example multiwindow
 
 For more information, please read the documentation below.
 
+## Rust <-> Javascript
+
+Communication between the host Rust code and webview Javascript is done via [JSON-RPC][].
+
+Embedding code should call `set_handler()` on the application to register an incoming request handler and reply with responses that are passed back to Javascript. On the Javascript side the client is exposed via `window.rpc` with two public methods
+
+1. The `call()` function accepts a method name and parameters and expects a reply.
+2. The `notify()` function accepts a method name and parameters but does not expect a reply.
+
+Both methods return promises but `notify()` resolves immediately.
+
+For example in Rust:
+
+```rust
+use wry::{Application, Result, RpcResponse};
+
+fn main() -> Result<()> {
+    let mut app = Application::new()?;
+    app.set_handler(Box::new(|proxy, mut req| {
+      // Handle the request of type `RpcRequest` and reply with `RpcResponse`
+      // Use the `WindowProxy` to modify the window, eg: `set_fullscreen` etc.
+      None
+    });
+    app.add_window(Default::default())?;
+    app.run();
+    Ok(())
+}
+```
+
+Then in Javascript use `call()` to call a remote method and get a response:
+
+```javascript
+async function callRemoteMethod() {
+  let result = await window.rpc.call('remoteMethod', param1, param2);
+  // Do something with the result
+}
+```
+
+See the `rpc` example for more details.
+
 ## [Documentation](https://docs.rs/wry)
 
 ## Platform-specific notes
@@ -84,3 +124,5 @@ WebView2 provided by Microsoft Edge Chromium is used. So wry supports Windows 7,
 
 ## License
 Apache-2.0/MIT
+
+[JSON-RPC]: https://www.jsonrpc.org
--- a/README.md
+++ b/README.md
@@ -50,26 +50,40 @@ For more information, please read the documentation below.
 
 Communication between the host Rust code and webview Javascript is done via [JSON-RPC][].
 
-Embedding code should call `set_handler()` on the application to register an incoming request handler and reply with responses that are passed back to Javascript. On the Javascript side the client is exposed via `window.rpc` with two public methods
+Embedding code should pass an `RpcHandler` to `add_window_with_configs()` to register an incoming request handler and reply with responses that are passed back to Javascript. On the Javascript side the client is exposed via `window.rpc` with two public methods
 
 1. The `call()` function accepts a method name and parameters and expects a reply.
 2. The `notify()` function accepts a method name and parameters but does not expect a reply.
 
-Both methods return promises but `notify()` resolves immediately.
+Both functions return promises but `notify()` resolves immediately.
 
 For example in Rust:
 
 ```rust
-use wry::{Application, Result, RpcResponse};
+use wry::{Application, Result, WindowProxy, RpcRequest, RpcResponse};
 
 fn main() -> Result<()> {
     let mut app = Application::new()?;
-    app.set_handler(Box::new(|proxy, mut req| {
-      // Handle the request of type `RpcRequest` and reply with `RpcResponse`
+    let handler = Box::new(|proxy: &WindowProxy, mut req: RpcRequest| {
+      // Handle the request of type `RpcRequest` and reply with `Option<RpcResponse>`, 
+      // use the `req.method` field to determine which action to take.
+      // 
+      // If the handler returns a `RpcResponse` it is immediately evaluated 
+      // in the calling webview.
+      // 
       // Use the `WindowProxy` to modify the window, eg: `set_fullscreen` etc.
+      // 
+      // If the request is a notification (no `id` field) then the handler 
+      // can just return `None`.
+      // 
+      // If an `id` field is present and the handler wants to execute asynchronous 
+      // code it can return `None` but then *must* later evaluate either 
+      // `RpcResponse::into_result_script()` or `RpcResponse::into_error_script()` 
+      // in the webview to ensure the promise is resolved or rejected and removed 
+      // from the cache.
       None
     });
-    app.add_window(Default::default())?;
+    app.add_window_with_configs(Default::default(), Some(handler), None)?;
     app.run();
     Ok(())
 }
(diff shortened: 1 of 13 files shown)


tmpfs:
@wusyong, this version has a tidy history. Would appreciate it if we could get this merged :+1: 

I tried and failed to fix the MacOS issue and hope you have more success. I tried using `OnceCell` to get an `&'static` reference to the `RpcHandler` but could not as it is not thread safe. It appears that ultimately it boils down to:

```
= help: within `[closure@src/application/general.rs:370:52: 380:10]`, the trait `Sync` is not implemented for `*mut winit::platform_impl::platform::observer::CFRunLoopSource`
```

Which is all the stuff inside the `WindowProxy` -> `ApplicationProxy` chain which I know very little about. Any idea how we can get the `RpcHandler` into `did_receive()` safely?

wusyong:
We probably need to declare another objc class and it's full of hacks. I can take another day to work it. But for now that's just make it work on linux.

tmpfs:
Thanks @wusyong, it works on Linux do you have a Windows machine to test on? It might work there too!

wusyong:
Sure. Would you mind I cherry pick this comment to `feat/rpc`? I wish I could work on it right away.

tmpfs:
Sure thing, let's put it in `feat/rpc` and merge it back into the main branch once we have all three platforms :+1: Let me know if there is anything else I can do to help. Maybe a tracking issue for progress on this feature?

wusyong:
We can keep the discussion on #66.
Currently I am thinking about making `Callback` in the application module work with rpc.
I will list more in the issue later.

tmpfs:
> Currently I am thinking about making Callback in the application module work with rpc.

Oh no, we may have misunderstood each other. I have removed `Callback` as I don't see the benefit it brings to have a separate Rust closure for each Javascript function. The new model is a single Rust closure which tests on the `method` of `RpcRequest`.

The benefit of the old callback model was that you could just call a vanilla Javascript function as I understood it. This can now be achieved like this:

```javascript
function someCallback() {
  return window.rpc.call('someRpcMethod', Array.prototype.slice.call(arguments, 0));
}

someCallback("param").then((res) => console.log(res));
```

What is still not done is a function that would set these *callback aliases* automatically. That would be the `set_rpc_aliases()` function I mentioned before.

wusyong:
I see. I do agree with this and it looks a lot cleaner.
Btw it turns out it's hell of chore to cherry pick 😅. Could you let me just add them manually?

tmpfs:
Hi @wusyong, sorry I didn't realize you have been working in the `feat/rpc` branch. I took a look at the commits and there are lots of conflicts with that I have done here. Particularly with regard to `CALLBACKS` and `Callback`.

What I attempted to do was be fully backwards compatible with the first PR that you merged and then when you said that you liked the idea of cleaning up `FuncCall` and `Callback` this PR was created that *completely removes any concept of callbacks* with the *exception of Javascript-defined callback functions* that wrap the RPC API.

Is it possible for me to take any important changes from your `feat/rpc` branch and merge them with this PR? Not sure which changes are important. Please advise.

wusyong:
Callback has to be here is because Tauri will need it. But I believe there's a way to build it on top of the RPC handler.
Let me try to merge your branch into `feat/rpc` and I'll list what should be done afterward in #66
We can remove callback which's fine, but it needs an alternative interface that could accomplish the same functionality.

wusyong:
Sorry I deleted the previous comment thinking it's not good for you to develop further.
I think let's just keep working on this branch and finish all the changes you would like to address (ie. RPC alias).
Forget about `feat/rpc`, I'll update them afterward once this PR is merged.

tmpfs:
Sure @wusyong, I'll do the callback aliases stuff and hope that it works out.

I looked at the [tauri code](https://github.com/tauri-apps/tauri/blob/adac93b9f10284e10bc58b436bb91284ed2cd132/tauri/src/app/utils.rs#L183) that configures the `Callback` and all they would need to do is change it to `Box<RpcHandler>` and call `set_handler()` when the application is set up. But without knowing the code well it is hard to say how much effort it would be.

Maybe we can get some people working on the main Tauri code to comment on how much this change will affect them?

wusyong:
Yeah we'll have a discussion about this.
Another thing I would like to address is maybe moves the handler to parameters of `Applicaton::create_window`.
I'm not a fan of requiring the correct order of calling methods (which we suffered a lot from original webview)

tmpfs:
> I'm not a fan of requiring the correct order of calling methods 

Agreed @wusyong . I did add a `panic!` for out of order calling but it's still a bit rubbish.

So then we remove `set_handler()` and `RpcHandler` is passed as the second argument to `add_window_with_configs()`? Replacing the old `Vec<Callback>`.

wusyong:
yeah, if that's possible.

tmpfs:
Sure thing @wusyong, that's done now and it feels more idiomatic :+1: 

However it does mean that for the *aliases* functionality that they would need to be passed in using the same mechanism so we probably want a wrapper type:

```rust
pub struct RpcConfig {
  handler: RpcHandler,
  callbacks: Vec<String>
}
```

Then we pass `Option<RpcConfig>` to `add_window_with_configs()` instead of `Option<RpcHandler>`; is that ok with you?

wusyong:
Current `Option<RpcHandler>` is cleaner imho. Let me check with other devs if we can actually replace callbacks first.

README.md
```diff
@@ -46,6 +46,46 @@ cargo run --example multiwindow
 
 For more information, please read the documentation below.
 
+## Rust <-> Javascript
```
wusyong, line 49: I left a comment in #39 that wish we could add descriptions to examples. Maybe you could start a `README.md` in `examples/`

examples/rpc.rs
```diff
@@ -38,7 +38,7 @@ async function getAsyncRpcResult() {
     };
 
     // NOTE: must be set before calling add_window().
```
wusyong, line 40: This comment could be removed right?

src/application/mod.rs
```diff
@@ -44,6 +43,7 @@ impl std::fmt::Debug for Callback {
             .finish()
     }
 }
+*/
```
wusyong, line 46: Let's kill these once and for all 🔪

src/application/mod.rs
```diff
@@ -293,13 +293,13 @@ pub enum WindowMessage {
 }
 
 /// Describes a general message.
-#[derive(Debug)]
+//#[derive(Debug)]
```
wusyong, line 296: It's not included in the error type anymore, so Debug trait is not required.

src/platform/linux.rs
```diff
@@ -1,17 +1,14 @@
...
 
 use std::rc::Rc;
-use std::sync::Arc;
 
-use serde_json::Value;
 use gdk::RGBA;
 use gio::Cancellable;
 use glib::{Bytes, FileError};
-use gtk::{ApplicationWindow as Window, ApplicationWindowExt, ContainerExt, WidgetExt};
+use gtk::{ApplicationWindow as Window, /* ApplicationWindowExt, */ ContainerExt, WidgetExt};
```
wusyong, line 11: You could remove it if not needed.

src/webview.rs
```diff
@@ -35,11 +34,12 @@ pub struct WebViewBuilder {
     initialization_scripts: Vec<String>,
     window: Window,
     url: Option<Url>,
+    // TODO: remove this field?
```
wusyong, line 37: yeah you could remove it since we just needed it for callbacks before.

wusyong (changes requested):
Great news! We could get rid of callbacks entirely! Just fix the following nits and I think it's good to go!

tmpfs:
Thanks @wusyong, excited to see this land! I have made those small changes you mentioned but I have a couple more points I would like to make:

1) We throw out the idea of *callback aliases* and just add documentation on how to do it in Javascript
2) I read the Tauri code some more and they do async stuff in the existing `Callback` code and we need to support that use case.

I have a good idea how we can support async flows it it just a small bit of code and some documentation. I will do it now and ask you to take a look, thanks :+1:

tmpfs:
Hi @wusyong, I think this can be merged if we can answer the following questions:

1. Does it work on Windows?
2. Do we have a clear path to get `RpcHandler` into `did_receive()` for MacOS support?

Hoping you are able to work out the MacOS issues! Cheers :+1:

examples/multiwindow.rs
```diff
@@ -1,63 +1,73 @@
...
+    await window.rpc.notify("openWindow", "https://i.imgur.com/x6tXcr9.gif");
+}
+</script>
+<p>Multiwindow example</p>
+<button onclick="openWindow();">Launch window</button>        
+"#;
+
     let attributes = Attributes {
-        url: Some("https://tauri.studio".to_string()),
+        url: Some(format!("data:text/html,{}", urlencoding::encode(html))),
```
wusyong, line 19:
```suggestion
        url: Some(format!("data:text/html,{}", html)),
```
It won't be rendered on windows if it's encoded. Looks like the web engine will do this for us.
Could you remove this dependency?

tmpfs: Sure thing, that's done! Strange they accept it un-encoded, I assume the browser is decoding before passing to the webview, all good :+1:

tmpfs:
Thanks for all your help @wusyong, shall we get this merged and then over to you for the platform-specific stuff?

wusyong (approved):
Sure, really appreciate the help!

PR #97, opened 2021-03-04
Use Rc for WindowProxy.

tmpfs:
Hi @wusyong,

I made a mistake passing `WindowProxy` by reference, could we pass an `Rc` instead so it is easy for calling code to pass off to services.

Hope you haven't started on the platform work so we can get this in. Thanks :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,11 @@ For example in Rust:
 
 ```rust
 use wry::{Application, Result, WindowProxy, RpcRequest, RpcResponse};
+use std::rc::Rc;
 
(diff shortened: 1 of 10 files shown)


wusyong:
Too late I already start it 😂
I removed the window proxy for now because I don't want to add more hacks on mac side.
I think we could do similar thing like callback type before.

tmpfs:
Cool. BTW, I tried removing `WindowProxy` too when trying to get the `RpcHandler` into `did_receive()` but failed, good luck :+1:

PR #98, opened 2021-03-04
Fix RPC handler on macOS

wusyong:


tmpfs:
Great stuff @wusyong, that was fast!

I have been thinking a little about the whole `WindowProxy` stuff and wonder whether we should be doing all communication over a channel that gets passed in to `Application::new()`. Something like this:

```
let (wtx, wrx) = std::sync::mpsc::channel::<WindowMessage>();
let app = Application::new(wrx)?;
let handler = Box::new(move |window_id: usize, mut req: RpcRequest| {
wtx.send(WindowMessage::Fullscreen(window_id, true));
});
// Set up the window with the handler and start the app running
```

Just tying to figure out how this could work without passing around `WindowProxy` all over the place. I think this pattern is fairly idiomatic.

wusyong:
WindowProxy is basically a channel for sending messages. It's more simple for users could just get one in the closure imho.
Anyway, it's back and mac port is also fixed. We could say that the RPC feature is complete!

tmpfs:
Wonderful, going to take it for a spin! Would it be ok to try to get my `Rc<WindowProxy>` update rather then passing the reference?

wusyong:
I think we don't need it. Proxy types are basically sender of the mpsc channel.

tmpfs:
Hi @wusyong, sorry to bother again, but could we at least do like the original implementation and pass a new `WindowProxy` for each invocation, passing the `&WindowProxy` is causing me all sorts of lifetime issues which I am unable to resolve.

PR #99, opened 2021-03-04
Pass WindowProxy not reference.

tmpfs:
So this is like the original implementation.

There is some small additional overhead on the extra `clone()`.


wusyong (approved):

tmpfs:
Fantastic @wusyong, I feel ready to crack on with my UI now :)

Shall we close #66 now?

wusyong:
The clone is fine since it's a reference counter on sender

PR #100, opened 2021-03-04
Fix rpc Javascript scope bug.

tmpfs:
Fix for a scope bug with the injected RPC code.

Suprised this didn't manifest earlier but after a colleague reported a bug on macOS I started debugging the RPC communication and came across this issue:

```
undefined:28:39: CONSOLE ERROR Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'this._promises[id] = {resolve, reject}')
```

Both `call` and `notify` must be `function` (not `() => {}`) so they can access `arguments` but because they are they don't benefit from inheriting the `this` scope so we fix this using the `const self`.

### DIFF
--- a/src/webview.rs
+++ b/src/webview.rs
@@ -85,6 +85,7 @@ impl WebViewBuilder {
     pub fn set_rpc_handler(mut self, handler: RpcHandler) -> Self {
         let js = r#"
             function Rpc() {
+                const self = this;
                 this._promises = {};
 
                 // Private internal function called on error
@@ -109,7 +110,7 @@ impl WebViewBuilder {
                     const params = Array.prototype.slice.call(arguments, 1);
                     const payload = {jsonrpc: "2.0", id, method, params};
                     const promise = new Promise((resolve, reject) => {
-                        this._promises[id] = {resolve, reject};
+                        self._promises[id] = {resolve, reject};
                     });
                     window.external.invoke(JSON.stringify(payload));
                     return promise;


tmpfs:
@wusyong, I think I only noticed this because my app has this code:

```javascript
window.onerror = function(message, filename, lineno, colno, error) {
    if (error != null) {
      console.error(error.stack);
    } else {
      console.error(`${message} ${filename} ${lineno}`);
    }
};
```

Is it worth considering injecting that by default? Happy to raise a new issue to track a feature for it. Maybe it could be a field on `Attributes` like `window_onerror`?

wusyong:
I would prefer for users to handle that themselves. Same reason I haven't added any log crate yet.

wusyong (approved):

tmpfs:
@wusyong, I think might be worth reverting the changes you made to use the `window.rpc` shortcut because it can be useful to intercept requests using `window.rpc`, for example I added this to my app to allow intercepting calls within the app for debugging:

```javascript
function RpcProxy() {

  this._result = function(id, result) {
    console.log("RpcProxy got result to resolve promise", id, result);
    window.external.rpc._result(id, result);
  }

  this._error = function(id, error) {
    console.log("RpcProxy got error to reject promise", id, error);
    window.external.rpc._error(id, error);
  }

  this.call = function() {
    let args = Array.prototype.slice.call(arguments)
    console.log("RpcProxy call: ", args);
    window.external.rpc.call.apply(window.external, args)
  }
}

window.rpc = new RpcProxy();
``` 

My thinking is the `wry` code should prefer to call directly to `window.external.rpc` and user code should prefer the `window.rpc` shortcut.

wusyong:
It didn't get the response on Windows when I reverted it back :(

tmpfs:
That's really interesting, thanks for letting me know. I wonder why that is.

tmpfs:
So Windows doesn't like the mutiple assignment style:

```
window.rpc = window.external.rpc = new Rpc();
```

wusyong:
Ah that's the reason. I'll just need to reorder the init scripts order

tmpfs:
I thought it might be, I have a little patch coming for super-defensive javascript if you want it!

PR #102, opened 2021-03-04
Restructure modules

wusyong:


src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! Wry is a Cross-platform WebView rendering library.
+//! Wry is a Cross-webview WebView rendering library.
```
nklayman, line 1: Cross-webview?

wusyong: ugh clion...

PR #120, opened 2021-03-17
feat/ allow webview2 (windows) to use optional user_data folder provided by the attributes.

lemarier:
Actually only supported on Windows.
Attention the folder need to exist or webview will failed to run.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

- On Windows, the default user data folder is the `exe path + .WebView2`.
- This causes problems with the packaged application as the `Program Files` is protected and can't be used for cached data.
- This allows us to set a different user data folder when needed.

This is only a demonstration of how we could fix it and feel free to make it better!

If you want me to do any changes let me know!

Also, this is NOT supported yet on macOS & Linux.

Thanks!

### DIFF
--- a/examples/custom_user_data.rs
+++ b/examples/custom_user_data.rs
@@ -1,12 +1,15 @@
 use std::path::PathBuf;
+use std::fs;
 use wry::{Application, Attributes, Result};
 
 fn main() -> Result<()> {
   let mut app = Application::new()?;
 
-  let test_path = PathBuf::from(env!("OUT_DIR"));
-
-  println!("Webview storage path: {:#?}", test_path);
+  // Use a sample directory at the root of the project
+  let test_path = PathBuf::from("./target/webview_data");
+  // The directory need to exist or the Webview will panic
+  fs::create_dir_all(&test_path)?;
+  println!("Webview storage path: {:#?}", &test_path);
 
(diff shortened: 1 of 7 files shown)


examples/custom_user_data.rs
```diff
@@ -0,0 +1,22 @@
+use std::path::PathBuf;
+use wry::{Application, Attributes, Result};
+
+fn main() -> Result<()> {
+  let mut app = Application::new()?;
+
+  let test_path = PathBuf::from(env!("OUT_DIR"));
```
wusyong: We should provide an example path imho. An environment variable is less clear.

lemarier: Something like `./target/webview_data` is good for you?

wusyong: sure!

lemarier: Perfect! Just tested and it works fine. 
Had to have an absolute path. 

So maybe it'll be important to tell it in the doc.

wusyong (approved):

PR #135, opened 2021-03-30
fix: spawn async event loop on gtk to prevent delayed messages (fix #132)

### LINKED ISSUE #132: Delay when processing javascript messages sent from the backend on Linux/GTK
**Describe the bug**
This bug occurs on the latest `dev` branch. I'm sending in a stream of `evaluate_script` commands into Wry as the backend listens to events from an external source. On the Winit backend this works perfectly, but on GTK these events are delayed in making it to the browser unless the user is actively engaged with the browser, such as clicking on something or moving their mouse. This happens because of the call to ` gtk::main_iteration()` on the main loop used by `src/application/gtkrs.rs`. `gtk::main_iteration()` is a blocking call, so once one iteration of the loop completes, no more messages sent to `self.event_loop_proxy_rx` will be processed until GTK thinks it needs to run the next iteration. If the user isn't actively using the browser, it appears GTK doesn't think it needs to update very often and only runs the next loop roughly once every second. As a result, every javascript callback sent into the queue gets batch executed roughly once per second instead of in real time.

**Steps To Reproduce**
I've created [an example here](https://github.com/aschey/wry/blob/server-push/examples/server_push.rs) that demonstrates this behavior. On Windows it behaves as expected, but on Linux it demonstrates the processing delay. The backend sends a counter increment to the UI every 500 milliseconds, but the UI only updates every second, so you should only see every other number displayed on the UI unless you focus on the window and start moving your mouse, then the UI should begin to update as expected. If you change the interval to something smaller, the batch delay becomes more apparent.

**Expected behavior**
I should be able to push events to the UI from the backend without user intervention and see the UI update without major processing delay.

OS: Arch Linux
Rustc: 1.51

- [x] Yes
- [ ] No

If I swap the call to `gtk::main_iteration()` with the nonblocking version `gtk::main_iteration_do(false)`, the code begins to process messages quickly and behave as expected, but now the entire loop is nonblocking and eats up a large amount of CPU, which is not great. The most obvious solution to me is to have the event loop for `self.event_loop_proxy_rx` run in a separate thread so it can process messages without waiting for gtk's event loop, however it appears that the gtk objects used in `self.webviews` are not threadsafe, so the thread that handles the event loop will also need complete ownership of the webviews.

I'd be happy to take a stab at this, but I'd like some input on my thoughts above as well as any other potential solutions first. Thanks!

aschey:
Resolves #132 

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
This change spawns an async event loop on GTK's main thread to process incoming messages from the backend instead of using the same loop that calls `gtk::main_iteration()`. This allows messages to be processed quickly even when the user isn't interacting with the app because messages won't be impeded while waiting for `gtk::main_iteration()` to complete.

### DIFF
--- a/src/application/gtkrs.rs
+++ b/src/application/gtkrs.rs
@@ -15,6 +15,7 @@ use std::{
   },
 };
 
+use async_channel::unbounded;
 use cairo::Operator;
 use gio::{ApplicationExt as GioApplicationExt, Cancellable};
 use gtk::{
@@ -23,7 +24,7 @@ use gtk::{
 
 pub type WindowId = u32;
 
-struct EventLoopProxy(Sender<Message>);
+struct EventLoopProxy(async_channel::Sender<Message>);
 
 impl Clone for EventLoopProxy {
   fn clone(&self) -> Self {
@@ -42,7 +43,7 @@ impl AppProxy for InnerApplicationProxy {
     self
       .proxy
       .0
-      .send(message)
+      .try_send(message)
       .map_err(|_| Error::MessageSender)?;
     Ok(())
   }
@@ -75,7 +76,7 @@ pub struct InnerApplication {
   webviews: HashMap<u32, WebView>,
   app: GtkApp,
   event_loop_proxy: EventLoopProxy,
-  event_loop_proxy_rx: Receiver<Message>,
+  event_loop_proxy_rx: async_channel::Receiver<Message>,
   event_channel: (Sender<WryEvent>, Arc<Mutex<Receiver<WryEvent>>>),
 }
 
@@ -88,7 +89,7 @@ impl App for InnerApplication {
     let cancellable: Option<&Cancellable> = None;
     app.register(cancellable)?;
 
-    let (event_loop_proxy_tx, event_loop_proxy_rx) = channel();
+    let (event_loop_proxy_tx, event_loop_proxy_rx) = unbounded::<Message>();
     let (tx, rx) = channel();
 
     Ok(Self {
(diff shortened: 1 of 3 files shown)


src/application/gtkrs.rs
```diff
@@ -88,7 +89,7 @@ impl App for InnerApplication {
     let cancellable: Option<&Cancellable> = None;
     app.register(cancellable)?;
 
-    let (event_loop_proxy_tx, event_loop_proxy_rx) = channel();
+    let (event_loop_proxy_tx, event_loop_proxy_rx) = unbounded::<Message>();
```
wusyong, line 92: Could it still use mpsc channel from std?

aschey: It has to be an async channel otherwise it blocks the main thread.

wusyong: Even we use `try_recv()` on `event_loop_proxy_rx`?

aschey: If we did that then we'd have to introduce an artificial sleep into the loop after each iteration and using the normal `thread::sleep` won't work since it'll still block the thread. We'd have to use an async sleep in that case.

wusyong: Okay, I saw it. It does block the thread.

wusyong (approved):

wusyong:
Oh I forgot to mention I tested on tauri example and it seems working fine.

aschey:
Awesome, thank you!

PR #164, opened 2021-04-19
Load html

FrankenApps:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

For standalone self-contained apps, that use html for the GUI it might be desirable to integrate the HTML into the binary. 

This is currently possible like so:
```rust
let window = app.add_window(Attributes{
        title: String::from("Hello Self-Contained"),
        url: Some(["data:text/html,", include_str!("../dist/index.html")].join("")),
        ..Default::default()
})?;
```

However there are multiple problems with this approach, the main one beeing, that URLs can only be of a [certain length](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers) and therefore this does not work for big files. Also the URL is checked, which is taking some time for larger strings.

This PR introduces an option to pass a `html` string like so:
```rust
let window = app.add_window(Attributes{
        title: String::from("Hello Self-Contained"),
        html: Some(include_str!("../dist/index.html").to_string()),
        ..Default::default()
})?;
```

**Other information:**

- The main open point is how this should behave, if `html` and `url` are set. Currently `html` would overwrite the `url` option. Specifying both just does not make sense to me.
- Maybe there should also be a `load_html(html: &str)` function somewhere, I am just not familiar enough with the API, so a hint, as to where this should be added would be appreciated

Apart from that I think it would be better to use a `&str` to remove the overhead, but I did not manage to use that in an `Option<>`.

I implemented this for myself first, but decided it might be worth to open a PR, to discuss this and maybe get some feedback.

### DIFF
--- a/src/application/attributes.rs
+++ b/src/application/attributes.rs
@@ -203,6 +203,12 @@ pub struct Attributes {
   ///
   /// The default is `None`.
   pub user_data_path: Option<PathBuf>,
+
+  /// A HTML String can be rendered with the webview window.
+  /// Use either this or load from a URL.
+  ///
+  /// The default is `None`.
+  pub html: Option<String>,
 }
 
 impl Attributes {
@@ -233,6 +239,7 @@ impl Attributes {
         url: self.url,
         initialization_scripts: self.initialization_scripts,
         user_data_path: self.user_data_path,
+        html: self.html,
       },
     )
   }
@@ -263,6 +270,7 @@ impl Default for Attributes {
       url: None,
       initialization_scripts: vec![],
       user_data_path: None,
+      html: None,
     }
   }
 }
@@ -293,4 +301,5 @@ pub(crate) struct InnerWebViewAttributes {
   pub url: Option<String>,
   pub initialization_scripts: Vec<String>,
   pub user_data_path: Option<PathBuf>,
+  pub html: Option<String>,
 }
--- a/src/application/general.rs
+++ b/src/application/general.rs
@@ -468,6 +468,11 @@ fn _create_webview(
     None => webview,
   };
 
+  webview = match attributes.html {
+    Some(html) => webview.load_html(html)?,
+    None => webview,
+  };
+
   let webview = webview.build()?;
   Ok(webview)
 }
--- a/src/application/gtkrs.rs
+++ b/src/application/gtkrs.rs
@@ -468,6 +468,11 @@ fn _create_webview(
     None => webview,
   };
 
+  webview = match attributes.html {
+    Some(html) => webview.load_html(html)?,
+    None => webview,
+  };
+
   for protocol in custom_protocols {
     webview = webview.register_protocol(protocol.name, protocol.handler);
   }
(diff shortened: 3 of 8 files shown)


nothingismagick:
While I think from a high level this totally makes sense, I am not sure how the mechanics should be implemented.

wusyong:
We do have API to load HTML string before. `navigate_to_string` is the leftover function which I planned to remove someday.
Right now the more proper way to load HTML data is to register a custom protocol for it. So this method was removed.
That being said, I have no preference for this.
@nothingismagick @lucasfernog wdyt?

nothingismagick:
Yeah, I know what you mean. I think that the custom protocol is actually the more secure of the two - and this `url` system is really quite useful for testing - but I can't see anyone really using this in production... Or is there a really valid use case that you have in mind @FrankenApps ?

FrankenApps:
No, not really. So did I understand the comment above correctly, that it is already possible to load html strings by registering a custom protocol?

nothingismagick:
For example how its being used in Tauri:

https://github.com/tauri-apps/tauri/blob/bf8213646689175f8a158b956911f3a43e360690/examples/multiwindow/src-tauri/src/main.rs

FrankenApps:
I see. This does however mean it's not possible to include the html inside the binary when using wry only? This is my current use case for this, but I do understand it is rather specific and I can therefore understand if this won't make it into wry.

amrbashir:
I recently was using wry with its custom protocol and [rust-embed](https://github.com/pyros2097/rust-embed) to do exactly what you want.

```rust
use rust_embed::RustEmbed;
use wry::{Application, Attributes, CustomProtocol, Result};

fn main() -> Result<()> {
  #[derive(RustEmbed)]
  #[folder = "public"] // relative path to your static files
  struct Assets;

  let mut app = Application::new()?;

  app.add_window_with_configs(
    Attributes {
      url: Some("wry://".to_string()), // navigate to your custom protocol
      ..Default::default()
    },
    None,
    vec![CustomProtocol {
      name: "wry".into(),
      handler: Box::new(|url| {
        match url.replace("wry:///", "").as_str() {
          // load index.html if url is empty after removing your custom protocol name from it
          "" => Ok(Assets::get("index.html").unwrap().into()),
          file => Ok(Assets::get(file).unwrap().into()),
        }
      }),
    }],
    None,
  )?;
  app.run();
  Ok(())
}


```

nothingismagick:
maybe a good idea to have an example of this approach @amrbashir & @wusyong ?

wusyong:
Closure in the custom protocol only cares about the returned bytes. So you could still hard code the HTML into it I believe.
We'll need a custom protocol example anyway. Feel free to open one for it.

FrankenApps:
@amrbashir Ok, thank you. Closing then.

PR #180, opened 2021-04-21
feat: allow resizing of borderless gtk windows

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

**Other information:**

it works fine excepts two things:
1. when the left mouse button is pressed, we call `begin_resize_drag` but that seems to stop the `notify-motion` signal from firing and the cursor style reverts back to normal. it could be that something wrong on my system (using a custom DE setup), so someone test and confirm please.

2. <s>when the webview is attached, the mouse click isn't passed through to the window widget, so the resize won't work. A possible solution is to detect the click on the webview dom and use the rpc handler to call `begin_resize_drag`.</s>. Fixed by attaching `button_press` event on the webview widget.

### DIFF
--- a/src/application/window.rs
+++ b/src/application/window.rs
@@ -4,6 +4,7 @@
 
 use std::{fmt, sync::mpsc::Sender};
 
+use gdk::{Cursor, CursorType, EventMask, WindowEdge, WindowExt};
 use gtk::{prelude::*, ApplicationWindow};
 use winit::{
   dpi::{PhysicalPosition, PhysicalSize, Position},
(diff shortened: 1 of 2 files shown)


wusyong:
Works for me. The `notify-motion` is kind of trivial for me. I think okay to skip at the moment.
Maybe add it to the comment as FIXME?

amrbashir:
omg I just realized I commited the FIXME note but didn't push 😅

wusyong (approved):

Issue #46, opened 2021-02-18
Custom menu bar


Milo123459:
Hey!

Feature request, a custom menu bar, ie, remove the windows default buttons and move it around with a custom system. Would be very cool!

nothingismagick:
The first part of this is possible already (the decorations boolean):
```
    let attributes = Attributes {
        decorations: false,
```

The second part is a bit more complicated (having a dragzone). I can confirm, however, on macos m1 it is possible today to resize the window as you would expect, but without that header bar there is no default dragzone - and therefore the undecorated window is stuck in place.

nothingismagick:
Here is a bit of research regarding this and a response over at the winit crate:
https://github.com/rust-windowing/winit/issues/427

nothingismagick:
And here is a partial solution for macos:
https://github.com/rust-windowing/winit/issues/1855

But it seems indeed like we should probably do this at Tauri level. @lucasfernog - what do you think?

nothingismagick:
And on a sidenote, this crate looks interesting (if rather outdated)
https://lib.rs/crates/fruitbasket

Milo123459:
Thanks, going to look into it!
Still would be cool if tauri could do it tho.

nothingismagick:
Don't get me wrong! We will absolutely HAVE to do this one way or the other.

nothingismagick:
Hmm - maybe I missed the point of your question. If you just want draggable menubar, we are just waiting for a new winit release and then having a draggable location and entirely custom area at the top are possible. I was researching into this kind of stuff:

<img width="512" alt="image" src="https://user-images.githubusercontent.com/35242872/111152457-083e8800-8591-11eb-8cfa-0f8a8b3011f7.png">

Because at the moment, Tauri does not have any means to create `Edit | View | History | ...` etc. This is totally important

Milo123459:
I meant as in these (on windows): ![e](https://salvage.is-inside.me/l3Cwpza3.png)

Then of course like that on other platforms, but of course that thing you said as well would be cool!

lucasfernog:
I think wry might need some changes to export the proper types on the proxy.

Milo123459:
I wouldn't know, but, probably.

wusyong:
It's better to implement in wry imho. The philosophy of this crate is to provide a safe interface. 
I'll start with windows to explore since it's the most difficult platform usually.
Anyone who would like to contribute is also welcome.

[...]

Issue #184, opened 2021-04-21
macOS Event Loop / Focus Bug


lemarier:
**Describe the bug**
This is the main issue to track all bugs related to the macOS Event Loop / Focus weird behaviors.

To resume the looks like the event loop is not working great with the focus.

On the first draw, the [activation hack](https://github.com/rust-windowing/winit/blob/b9f3d333e41464457f6e42640793bf88b9563727/src/platform_impl/macos/activation_hack.rs#L1) seems to disable some elements and makes the window unusable.

We also lost the focus on the Window / Webview causing different problems.

**Steps To Reproduce**
N/A

**Expected behavior**
Should work exactly like a native window.

**Screenshots**
N/A

**Platform and Versions (please complete the following information):**
OS: macOS

**Additional context**
Please refer to linked issues.

wusyong:
Closed by #220

PR #187, opened 2021-04-22
feat(macOS): Implement basic menubar

lemarier:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

Related to #46


lemarier:
https://user-images.githubusercontent.com/22237916/115715392-11680500-a346-11eb-9b15-1901cb8d60c2.mov

lemarier:
The menu does not work on the first render.

We need to focus in / focus out of the window.

Also, it seems to cause an issue with the keyboard shortcut, sometimes it works, sometimes it doesn't.

It's related to focus issue with Winit / Webview probably related to https://github.com/tauri-apps/wry/issues/184

lemarier:
https://user-images.githubusercontent.com/22237916/115722813-5ba0b480-a34d-11eb-941f-00f1b8ccd5f4.mov

Known bug by winit for focus on first draw

https://github.com/rust-windowing/winit/pull/1903

src/webview/macos/menu_bar.rs
```diff
@@ -0,0 +1,253 @@
...
+    selector("handleCustomMenu:"),
+    handle_menu as extern "C" fn(&Object, _cmd: Sel, item: id),
+  );
+  decl.add_ivar::<*mut c_void>("eventProxy");
+}
+
+// Safety: objc runtime calls are unsafe
+pub(crate) unsafe fn create_menu(window: &Window) {
+  // make sure our raw window is available
+  if let RawWindowHandle::MacOS(_window) = window.raw_window_handle() {
```
wusyong: Could we check it with the method from `WindowExtMacOS` trait? Like checking if it's not null.

lemarier: Thanks I'll check!

lemarier: So with my tests, we don't even require to check `RawWindowHandle::MacOS(_window) = window.raw_window_handle()`, I switched to the `sharedApplication` method and worked fine and forgot to test without the raw_window_handle

It may be required for the custom implementation of the menu handler.

Or we could also check to do something like you've done with file_drop with callback Fn but I think the custom menu will be another scope. Because I don't want to register a method for each menu item as it makes no sense.

wusyong: yeah custom menu should be another scope. Let's leave it to future us to worry.

lemarier:
Closing in favor of https://github.com/tauri-apps/tao/pull/12

PR #223, opened 2021-04-29
feat(gtk): add mouse/cursor events

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

**Other information:**

### DIFF
--- a/src/application/event_loop.rs
+++ b/src/application/event_loop.rs
@@ -35,7 +35,9 @@ use winit::{
 };
 
 use super::{
-  event::{Event, StartCause, WindowEvent},
+  event::{
+    DeviceId, ElementState, Event, ModifiersState, MouseButton, StartCause, TouchPhase, WindowEvent,
+  },
   window::{WindowId, WindowRequest},
 };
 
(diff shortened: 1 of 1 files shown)


wusyong (approved):

src/application/event_loop.rs
```diff
@@ -429,6 +431,115 @@ impl<T: 'static> EventLoop<T> {
               }
               Inhibit(false)
             });
+
+            let tx_clone = event_tx.clone();
+            window.connect_enter_notify_event(move |_, _| {
+              if let Err(e) = tx_clone.send(Event::WindowEvent {
+                window_id: id,
+                event: WindowEvent::CursorEntered {
+                  // FIXME: currently we use a dummy device id, find if we can get device id from gtk
```
wusyong, line 440: Is it possible to `get_device` from closure's parameter and then get either vendor id or product id?

src/application/event_loop.rs
```diff
@@ -429,6 +431,115 @@ impl<T: 'static> EventLoop<T> {
...
+              if let Some(cursor) = display
+                .get_device_manager()
+                .and_then(|device_manager| device_manager.get_client_pointer())
+              {
+                let (_, x, y) = cursor.get_position();
+                if let Err(e) = tx_clone.send(Event::WindowEvent {
+                  window_id: id,
+                  event: WindowEvent::CursorMoved {
+                    position: PhysicalPosition::new(x as f64, y as f64),
+                    // FIXME: currently we use a dummy device id, find if we can get device id from gtk
```
wusyong, line 464: Ditto and the rest below.

src/application/event_loop.rs
amrbashir: yeah it is possible to get the device vendor id or product id but it returns a `GString` and I have no idead what is that and winit expects a `u32`

wusyong: The field of DeviceID is not public. I think you are free to change to any type you want. Be sure to implement Copy and Clone trait. I know this is usually not recommended but compatibility with winit API is more important to us.

wusyong: Or we can hash the string into an integer.

amrbashir: `get_product_id` and `get_vendor_id` seems to be behind a feature flag `#[cfg(any(feature = "v3_16", feature = "dox"))]`

and btw winit is also passing an empty `DeviceID` on some of their backends like windows so I think we should be fine.

PR #225, opened 2021-04-29
fix(gtk): correct `drag_window` signature, closes #224

### LINKED ISSUE #224: `drag_window` API is not consistent
On Linux, it returns `()`: https://github.com/tauri-apps/wry/blob/cea79fb7098cf870274b24512b117bd220ca7f34/src/application/window.rs#L717
But the winit interface returns a Result: https://docs.rs/tauri-winit/0.24.1/winit/window/struct.Window.html#method.drag_window

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**

### DIFF
--- a/src/application/window.rs
+++ b/src/application/window.rs
@@ -714,13 +714,14 @@ impl Window {
     self.maximized.load(Ordering::Acquire)
   }
 
-  pub fn drag_window(&self) {
+  pub fn drag_window(&self) -> Result<(), ExternalError> {
     if let Err(e) = self
       .window_requests_tx
       .send((self.window_id, WindowRequest::DragWindow))
     {
       log::warn!("Fail to send drag window request: {}", e);
     }
+    Ok(())
   }
 
   pub fn set_fullscreen(&self, fullscreen: Option<Fullscreen>) {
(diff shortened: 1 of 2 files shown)


src/application/window.rs
```diff
@@ -714,13 +714,14 @@ impl Window {
     self.maximized.load(Ordering::Acquire)
   }
 
-  pub fn drag_window(&self) {
+  pub fn drag_window(&self) -> Result<(), ExternalError> {
     if let Err(e) = self
```
lucasfernog, line 718: should we propagate this error instead?

amrbashir: why ? `ExternalError` indicates its either a OS error, or just not a supported method and the error here is `SendError`.

lucasfernog: yeah, and also i think the SendError should never happen right

wusyong: SendError occurs only when the receiver is disconnected first. This depends on how gtk handles its threads and tasks.

wusyong (approved):

PR #266, opened 2021-05-16
Add screenshot features for linux and win32

Beanow:
Sharing for reference. Hopefully can serve as a starting point, not sure I can figure out much more myself.

- Linux
  - [x] Visible
  - [x] Full document
- Win32
  - [x] Visible
  - [ ] Full document: no luck

 https://github.com/MicrosoftEdge/WebView2Feedback/issues/529#issuecomment-772540317
- WinRT: haven't looked at.
- Macos
  - [x] Visible
  - [ ] Full document: not sure yet


wusyong:
I updated the snapshot handler on mac and now it can handle the snapshot now.
The next thing to do is to convert UIImage to bytes.

wusyong:
okay, now basic mac support is also done.
While we are exploring full document on mac, I feel like we should think about the API too.
It only supports PNG at the moment. It feels weird to pass a vector of bytes.
Perhaps the parameter should just be a path to create the file?

Beanow:
Great stuff @wusyong 

> I feel like we should think about the API too.
> It only supports PNG at the moment. It feels weird to pass a vector of bytes.
> Perhaps the parameter should just be a path to create the file?

The current API was a dirty hack to get some image from A to B :smile:.
For many browser implementations, they offer saving to file, but also copy to clipboard. And some apps may want to do their own processing of the image? Being able to grab some image buffer could have some utility.
So that's why I thought about passing some platform independent image handle to the callback.
Bonus that the platform specific screenshot implementations don't need to include file IO.

wusyong:
If that's so, I think we could pass an image type we define instead of vector then.
It could provide methods to create file, copy to clipboard and support many other file formats.

Beanow:
Yeah that sounds good to me.
Possibly we could get rid of the cairo png dependency for Linux with that as well at least in the screenshot stage.

PR #281, opened 2021-05-20
webdriver support

chippers:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Issue #___
- [ ] No

**Other information:**
I marked this as a draft cause the interfaces are not final. If a release is not planned soon, then this can be merged and the `Application` interface can be worked on before the next release. LMK if the non-final interface is okay to merge for now and I'll mark it for review.

Overall this was less lines of code than I thought, but putting it together correctly in the right order was important. I guess since we are heavily leveraging existing webdriver servers (`WebKitWebDriver` and `msedgedriver`) it was mostly about enabling proper support for them.

Some notes on the changes:
* `webkit2gtk` starts webdriver support on `v2_18` so I bumped it up to that.
* `unix::ApplicationExt` allows setting the automation flag for `Application` because there is only allowed to be 1 automation context at a time. The other platforms have similar limitations I imagine (just from how webdriver works typically), but `webkit2gtk` semi-enforces this through the api.
    * Note: By default the context is an automation context. This just means it _allows_ automation and not that it requires it. 
* `wry::Application` is back for the above reason. It should be thought of/acts like a "browser" where there might be multiple windows and/or tabs but it refers to the same context. Like how incognito mode is a single context that shares multiple windows/tabs. It doesn't really do anything other than represent some context specific data.

I'm thinking that `wry::Application` can be made into a builder, and start adding items that are getting passed to `InnerWebView::new()` to it that make sense. So far, I've only moved over `data_directory`.

Also, because we now have `wry` specific stuff in application (`Application` and `mod unix`), it gets mixed in with the tao re-exports. Maybe we should move the tao re-export to the crate root and maybe also add root re-exports for items expected to be used by most wry applications (`EventLoop`, `WindowBuilder`, `Application`, `WebviewBuilder`, ect...)

_only linux + windows for now_


chippers:
@wusyong tagging you in a comment since I don't want to mark this as Review until we know how we want to expose `Application` (I'm thinking a builder).

wusyong:
I prefer to move all the setup of `Application` type into `WebViewBuilder`.
I looked up the `ApplicationInner` for Linux and Windows. And it should be possible to pass those fields into `WebViewInner` with `WebViewBuilder` I believe.

chippers:
> 
> 
> I prefer to move all the setup of `Application` type into `WebViewBuilder`.
> I looked up the `ApplicationInner` for Linux and Windows. And it should be possible to pass those fields into `WebViewInner` with `WebViewBuilder` I believe.

I'll look into this approach

src/webview/web_context.rs
```diff
@@ -0,0 +1,171 @@
...
+
+impl WebContextData {
+  /// A reference to the data directory the context was created with.
+  pub fn data_directory(&self) -> Option<&Path> {
+    self.data_directory.as_deref()
+  }
+}
+
+#[cfg(not(target_os = "linux"))]
+#[cfg_attr(doc_cfg, doc(cfg(not(target_os = "linux"))))]
```
wusyong: We don't need doc_cfg, the doc should create the right one according to `#[cfg(not(target_os = "linux"))]`

src/webview/web_context.rs
```diff
@@ -0,0 +1,171 @@
+use std::path::{Path, PathBuf};
+
+/// A context that is shared between multiple [`WebView`]s.
```
wusyong, line 3:
```suggestion
/// A context that is shared between multiple [`WebView`]s.
/// [`WebView`]: crate::webview::WebView
```

wusyong (changes requested):

src/webview/web_context.rs
chippers: oops that was leftover from when i merged the windows and mac one into a single unit struct

chippers:
right now the `unix::WebContextImpl` struct is behind a private crate module, since it's only used internally for now. Realistically this information isn't problematic to expose publicly, should the `web_context` module be `pub` (and therefore the unix module and `unix::WebContextExt` also)?

wusyong (dismissed):

wusyong (approved):

PR #289, opened 2021-06-06
feat: expose `eval` method

geieredgar:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
This allows executing javascript code without having to use the `mpsc::channel` and helps to simplify `handle_event_loop` in `tauri-runtime-wry` (see tauri-apps/tauri#1962).

### DIFF
--- a/.changes/eval.md
+++ b/.changes/eval.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Expose `eval` method in the `webview` module.
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -354,7 +354,10 @@ impl WebView {
     Ok(())
   }
 
-  /// Evaluate javascript code immediately.
+  /// Evaluate and run javascript code. Must be called on the same thread who created the
+  /// [`WebView`]. Use [`EventLoopProxy`] and a custom event to send scripts from other threads.
+  ///
+  /// [`EventLoopProxy`]: crate::application::event_loop::EventLoopProxy
   pub fn eval(&self, js: &str) -> Result<()> {
     self.webview.eval(js)
   }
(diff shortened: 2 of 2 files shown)


wusyong:
The problem is this method is not thread-safe. It should be called on the main thread only.
Mpsc channel could help you send the script on any thread.
I think one workaround is we could call `evaluate_script` immediately after `dispatch_script` if it's on main thread (ie. in the event loop)
@lucasfernog wdyt?

geieredgar:
Yes, calling `evaluate_script` immediately after `dispatch_script` is also what I did in my PR [tauri-apps/tauri#1963](https://github.com/tauri-apps/tauri/pull/1963) to fix [tauri-apps/tauri#1962](https://github.com/tauri-apps/tauri/issues/1962). I just think that solution is a bit ugly, which is why I opened this PR. In my opinion it would be a lot cleaner if wry did only provide the `eval` method and no `evaluate_script`, `dispatch_script` and `Dispatcher`, with a doc comment stating that `eval` should only be called on the main thread (I think this information should also be added to `evaluate_script`) and maybe a hint that users could use channels if they want to send scripts from any thread. I actually think that using the `EventLoopProxy` to send scripts from any thread is a better solution than using an mpsc channel because it always wakes up the event loop, which would not happen if a script is send via `Dispatcher`, where we cannot call `evaluate_script` afterwards and would have to send a message to the event loop anyway.

wusyong:
Yeah, I think we could do that. If @lucasfernog also agrees (since tauri will need some custom event on EventLoopProxy), let's remove the dispatcher and all its related methods and expose this instead.

geieredgar:
FYI, `tauri` already uses a `WebviewMessage::EvaluateScript` custom event and AFAICT doesn't use the `Dispatcher`.

geieredgar:
Closed in favor of #291.

Issue #276, opened 2021-05-19
Ability to set custom User Agent


aditsachde:
**Is your feature request related to a problem? Please describe.**
It'd be useful to set a custom UA for identifying desktop apps. Additionally, some sites require UA spoofing to work.

**Describe the solution you'd like**
Add a way to allow setting a custom UA for a webview.

**Describe alternatives you've considered**
N/A

**Would you assign yourself to implement this feature?**
- [x] Yes
- [ ] No

**Additional context**
N/A

I'd like to make an attempt at adding this feature, the main question I have is whether it's reasonable to implement it as a property on `WebViewBuilder` and `InnerWebView::new`, or if there is a better place.

shirakaba:
Exposing it on WebViewBuilder may make sense for convenience, but ideally it should be possible to update UA on the WebView at any moment post-construction, because it is essentially a dynamic property rather than something you might only want to set once.

For example, you might have an app that augments a user's YouTube and Netflix experiences, and therefore might have to set a different UA on each site to unlock different features (e.g. text subtitles rather than vector subtitles on Netflix). Ideally the API needs to be flexible enough to allow a user to update the UA just before navigation.

_Side-note:_ I'm not sure what the design philosophy behind WebViewBuilder is. For iOS/macOS WKWebView, for example, the only things that need to be set up at construction time are the `frame` and the [WKWebViewConfiguration](https://developer.apple.com/documentation/webkit/wkwebviewconfiguration), and the latter can't be updated later (there is no setter). So the fact that WebViewBuilder lets you set things like URL suggests that it also offers offers APIs just out of convenience, rather than strictly construction-time necessity.

PR #280, opened 2021-05-19
Fix: 276 - Add ability to set custom User Agent

aditsachde:
It'd be useful to set a custom UA for identifying desktop apps. Additionally, some sites require UA spoofing to work.

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes. Issue #___
- [x] No

**The PR fulfills these requirements:**

- x ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).

If adding a **new feature**, the PR's description includes:
- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**
- The Win32 Webview2 API does not support overriding the user agent yet. 
- The linux and winrt implementations still have to be tested.
- @shirakaba mentioned in #276 that ideally it should be possible to update the user agent on the fly. This makes sense, but I do not know the best way to implement this functionality.
- I still need to add a change file.

PR #464, opened 2022-01-05
Fix: #276 - Add ability to set custom User Agent

jhonatn:
It'd be useful to set a custom UA for identifying desktop apps. Additionally, some sites require UA spoofing to work.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix #___, #___`)

### Other information
- Based on the [draft PR 280](https://github.com/tauri-apps/wry/pull/280) from @aditsachde 
- The Windows implementation still has to be tested


src/webview/webview2/mod.rs
```diff
@@ -471,6 +471,13 @@ impl InnerWebView {
       }
     }
 
+    // Set user agent
+    if let Some(user_agent) = attributes.user_agent {
+      unsafe {
+        webview.Settings()?.SetUserAgent(String::from(user_agent.as_str()));
```
wusyong: Looks like I don't have the right to push your PR.
Could you add these lines for me?
```suggestion
        let settings: ICoreWebView2Settings2 = webview.Settings()?
          .cast()
          .map_err(webview2_com::Error::WindowsError)?;
        settings.SetUserAgent(String::from(user_agent.as_str()))?;
```

wusyong (changes requested):

src/webview/webview2/mod.rs
wusyong: and probably run `cargo fmt` after this

lucasfernog:
We need to check if this has security implications, if so we need to push it into v2.

src/webview/webview2/mod.rs
jhonatn: Done!

wusyong:
Should be fine if tauri don't expose it on v1 I suppose. I'll merge this in wry first.

PR #500, opened 2022-02-17
Changed NSDragOperation from link to copy

icewind:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

Consistency improvement. 
Currently there is inconsistency between implementations of webview for different platforms. For example in `webview2` the drag handler returns `DROPEFFECT_COPY` while in webkit it was NSDragOperation.link. 
In addition there is another handler in winit/tao which also returns NSDragOperation.copy and currently the cursor in MacOS flickers between two states when dragging onto the application border and the webview itself.

- [x] No

Since there is no ability to define a custom callback for d&d events, this PR affects only UX

### Checklist

### DIFF
--- a/src/webview/wkwebview/file_drop.rs
+++ b/src/webview/wkwebview/file_drop.rs
@@ -19,7 +19,7 @@ use crate::{application::window::Window, webview::FileDropEvent};
 
 pub(crate) type NSDragOperation = cocoa::foundation::NSUInteger;
 #[allow(non_upper_case_globals)]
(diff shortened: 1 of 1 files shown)


amrbashir (approved):
Thanks, I will defer the merge to @wusyong

wusyong (approved):

PR #507, opened 2022-02-21
Fix rustdoc generation of Windows and Mac on docs.rs, fix #503

### LINKED ISSUE #503: Doc.rs of macOS and Windows aren't built
The docs of Mac and Windows are missing after 0.13.
This is probably because of `thiserror`. Its recent patch version only generates linux target.
And on Windows, it's also because of sys-info. Since it's only used for detecting windows 7, we could add our own helper function instead.

wusyong:

- [x] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,16 +13,16 @@ documentation = "https://docs.rs/wry"
 categories = [ "gui" ]
 
 [package.metadata.docs.rs]
-features = [ "dox" ]
-default-target = "x86_64-unknown-linux-gnu"
+default-features = false
+features = [ "dox", "file-drop", "protocol" ]
 targets = [
-  "x86_64-pc-windows-msvc",
   "x86_64-unknown-linux-gnu",
+  "x86_64-pc-windows-msvc",
   "x86_64-apple-darwin"
 ]
 
 [features]
-default = [ "file-drop", "protocol" ]
+default = [ "file-drop", "objc", "protocol" ]
 file-drop = [ ]
 protocol = [ ]
 dox = [ "tao/dox" ]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -78,5 +78,5 @@ sys-info = "0.9"
 [target."cfg(any(target_os = \"ios\", target_os = \"macos\"))".dependencies]
 cocoa = "0.24"
 core-graphics = "0.22"
-objc = { version = "0.2", features = [ "exception" ] }
+objc = "0.2"
```
lucasfernog, line 82: I really think you should keep the `exception` as a default feature flag, disabled on docs.rs, this was part of the audit

lucasfernog:
Are you sure it builds without the default features? I thought we only needed to use the objc/exception feature as a default feature.

wusyong:
No, it's not. That's why I set the features need to use in docs.rs
That being said, I update the scope of default features like you said.

Cargo.toml
```diff
@@ -78,5 +78,5 @@ sys-info = "0.9"
 [target."cfg(any(target_os = \"ios\", target_os = \"macos\"))".dependencies]
 cocoa = "0.24"
 core-graphics = "0.22"
-objc = { version = "0.2", features = [ "exception" ] }
+objc = { version = "0.2", features = ["exception"] }
```
lucasfernog: you still need to remove the feature here right? since it's enabled with default features now

lucasfernog (approved):

PR #512, opened 2022-03-01
fix(macOS): enable tab navigation on all elements, fixes #406

### LINKED ISSUE #406: Tab navigation not working in MacOS
**Describe the bug**
When navigating through the TAB button in a form, some elements are skipped:
- input radio
- input checkbox
- button

**Steps To Reproduce**
Clone the `snapshot/tabbing-behavior` branch from [repo](https://github.com/gigioSouza/consolidAcao/tree/snapshot/tabbing-behavior)

Install front-end dependencies with `npm install` and run the following commands in different terminal windows:
```
// runs the fron-end devServer
npm run dev

// runs the tauri app in dev mode
npm run tauri:dev
```

Follow this gif to setup the first steps which you'll just need to select a directory where you want the app to store your data, like sqlite database.
It will create a directory called `consolidAcao` in the directory you choose.
You may skip this step, first skip the gif.
tauri-tabindex-4

To skip the apps first steps, open the devtools in the console tab and run the following command in it:
```js
location.href = 'http://localhost:8080/home'
```

You'll be redirected to the following page
image

Now you'll only need to press the TAB button from your keyboard and check that the input radio, input checkbox and button will never get focused.
(I added a CSS to outline in red the focused element).

> All elements in this sample are using the atribute `tabindex` with a number greater than `0`

**Expected behavior**
Be able to interact with any form control using TAB navigation.

**Screenshots**
Tabbing in Tauri instance
tauri-tabindex-1

Tabbing in Brave Browser
tauri-tabindex-2

Tabbing in Safari, after [allowing tab navigation through System Preferences](https://stackoverflow.com/questions/1848390/safari-ignoring-tabindex)
tauri-tabindex-3

OS: Mac OS, version 11.5.0 X64
Rustc: 1.54.0
Node: v16.6.1

- [ ] Yes
- [x] No

**Additional context**
None

FabianLars:
I forgot the changefile as always, but i'll add it once i'm back and if someone can confirm that this actually works (it did work in my mac vm).

Furthermore we might want to solve this differently, like giving the user an option to choose if this should be enabled or not, because apparently the default shortcut for this is command+tab instead of tab :shrug: (i'd prefer to force this behavior for cross-platform consistency)

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/src/webview/wkwebview/mod.rs
+++ b/src/webview/wkwebview/mod.rs
@@ -261,6 +261,9 @@ impl InnerWebView {
         let _: id = msg_send![_preference, setValue:_yes forKey:dev];
       }
 
+      #[cfg(target_os("macos"))]
+      let _: () = msg_send![_preference, setValue:_yes forKey:NSString::new("tabFocusesLinks")];
+
       #[cfg(feature = "transparent")]
       if attributes.transparent {
         let no: id = msg_send![class!(NSNumber), numberWithBool:0];
--- a/src/webview/wkwebview/mod.rs
+++ b/src/webview/wkwebview/mod.rs
@@ -261,8 +261,8 @@ impl InnerWebView {
         let _: id = msg_send![_preference, setValue:_yes forKey:dev];
       }
 
-      #[cfg(target_os("macos"))]
-      let _: () = msg_send![_preference, setValue:_yes forKey:NSString::new("tabFocusesLinks")];
+      #[cfg(target_os = "macos")]
+      let _: id = msg_send![_preference, setValue:_yes forKey:NSString::new("tabFocusesLinks")];
 
       #[cfg(feature = "transparent")]
       if attributes.transparent {
--- a/.changes/fix-tab-focus.md
+++ b/.changes/fix-tab-focus.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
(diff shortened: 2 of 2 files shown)


src/webview/wkwebview/mod.rs
```diff
@@ -261,6 +261,9 @@ impl InnerWebView {
         let _: id = msg_send![_preference, setValue:_yes forKey:dev];
       }
 
+      #[cfg(target_os("macos"))]
+      let _: () = msg_send![_preference, setValue:_yes forKey:NSString::new("tabFocusesLinks")];
```
FabianLars: should this be `let _: id = `?

wusyong: You are right with this one. But the syntax of cfg should be `#[cfg(target_os = "macos")]`

FabianLars: Ah whoops, you're right

wusyong:
Just runned some website login pages and it seems working

wusyong (approved):

PR #514, opened 2022-03-02
Fix resizing issue custom titlebar example

JasperVanEsveld:
It was impossible to click and resize using the top edge of the window in the custom titlebar example
Instead the window would be dragged.
By moving the `drag-region` to the title element and using `grid` instead of `flex` this is fixed.
Given that people often start working from a certain example I think it could save people quite some debugging time.

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/examples/custom_titlebar.rs
+++ b/examples/custom_titlebar.rs
@@ -27,8 +27,8 @@ fn main() -> wry::Result<()> {
 
   let url = r#"data:text/html,
         <body>
-          <div class='drag-region titlebar'>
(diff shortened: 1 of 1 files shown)


amrbashir (approved):
This is really awesome, thank you.

PR #519, opened 2022-03-07
feat: Implement navigation event and cancellation, closes #456

### LINKED ISSUE #456: Event or method to intercept page redirects/navigation
I'd like to be able to detect or monitor when the URL of a webview changes, for instance, when clicking on a link. This can be helpful for implementing things like custom download behaviour.

Either a new event describing the URL the current webview or a new webview has navigated to OR a function that can be called on a the WebviewBuilder which allows the registration of a callback which is executed whenever navigation ocurrs.

You could register HTTP/HTTPS as custom protocols, do the fetch yourself in rust then pass the resulting data through, though this seems a bit extreme if you want to just say log page navigations.

You could also potentially insert an `onbeforeunload` handler using the `with_initialization_script` builder function, though that might not catch certain navigations events or popups (which may themselves be another problem altogether). There could be a number of other ways to do this through pure javascript too.

- [x] Yes
- [ ] No

**Additional context**
There may be an obvious way to do this that I haven't seen, but I can't find it.

atlanticaccent:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
- [ ] There are likely code style issues with this
- [ ] This is only fully implemented for Windows

Notable: the design approach taken is to:
- Make WebContexts generic over user event types so that they can (potentially) store an EventLoopProxy
  - This is non-breaking, as a type alias of `WebContextGeneric<()>` is provided, named `WebContext` which effectively hides any generic functionality from consumers that do not need it.
- WebContextBuilders are necessarily also made generic over the user event type, however this should have minimal impact
- The EventLoopProxy is utilised to send a user defined event whenever a navigation occurs. 
- An "interface" is effectively defined by `WebContextGeneric::with_navigation_event` which requires as argument a closure that can construct the user event that is emitted whenever a navigation occurs, thus ensuring that any user defined events also include an event for navigation
- Navigation events cannot be cancelled in "real time", ie: within the event loop. A lambda must instead be provided that is executed within the navigation event handler itself - what difference this makes is unclear, however it does not appear to be a major drawback from what I have seen
- A basic example is included - it demonstrates some notable drawbacks, including: not preventing javascript that effectively rewrites a page (most noticeable when navigating to Twitter and then opening any other user's feeds) and it also does not prevent popups (a future PR or enhancement to this one could prevent this, on Windows at least)

### DIFF
--- a/src/webview/web_context.rs
+++ b/src/webview/web_context.rs
@@ -11,7 +11,7 @@ use crate::webview::webkitgtk::WebContextImpl;
 #[cfg(any(target_os = "macos", target_os = "ios"))]
 use crate::webview::wkwebview::WebContextImpl;
 
-use std::path::{Path, PathBuf};
+use std::{path::{Path, PathBuf}, sync::Arc};
 
 /// A context that is shared between multiple [`WebView`]s.
 ///
@@ -24,10 +24,10 @@ use std::path::{Path, PathBuf};
 /// interact with them.
 ///
 /// [`WebView`]: crate::webview::WebView
-#[derive(Debug)]
 pub struct WebContextGeneric<T: 'static> {
   data: WebContextData,
   event_loop_proxy: Option<EventLoopProxy<T>>,
+  navigation_event: Option<Arc<dyn Fn(String) -> T>>,
   #[allow(dead_code)] // It's not needed on Windows and macOS.
   pub(crate) os: WebContextImpl,
 }
@@ -43,7 +43,12 @@ impl<T: 'static> WebContextGeneric<T> {
   pub fn new(data_directory: Option<PathBuf>) -> Self {
     let data = WebContextData { data_directory };
     let os = WebContextImpl::new(&data);
-    Self { data, event_loop_proxy: None, os }
+    Self {
+      data,
+      event_loop_proxy: None,
+      navigation_event: None,
+      os
+    }
   }
 
   /// A reference to the data directory the context was created with.
@@ -62,6 +67,7 @@ impl<T: 'static> WebContextGeneric<T> {
   pub fn with_event_loop_proxy<U: 'static>(self, proxy: EventLoopProxy<U>) -> WebContextGeneric<U> {
     WebContextGeneric {
       event_loop_proxy: Some(proxy),
+      navigation_event: None,
       data: self.data,
       os: self.os
     }
@@ -71,13 +77,27 @@ impl<T: 'static> WebContextGeneric<T> {
     self.event_loop_proxy.as_ref()
   }
 
+  pub fn with_navigation_event(mut self, event: impl Fn(String) -> T + 'static) -> Self {
+    self.navigation_event = Some(Arc::new(event));
+
+    self
+  }
+
+  pub fn navigation_event(&self) -> Option<&Arc<dyn Fn(String) -> T>> {
+    self.navigation_event.as_ref()
+  }
 }
 
 impl<T: 'static> Default for WebContextGeneric<T> {
   fn default() -> Self {
     let data = WebContextData::default();
     let os = WebContextImpl::new(&data);
-    Self { data, event_loop_proxy: None, os }
+    Self {
+      data,
+      event_loop_proxy: None,
+      navigation_event: None,
+      os
+    }
   }
 }
 
@@ -105,4 +125,10 @@ impl WebContextImpl {
   }
 
   fn set_allows_automation(&mut self, _flag: bool) {}
+}
+
+impl<T: 'static> std::fmt::Debug for WebContextGeneric<T> {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    f.debug_struct("WebContextGeneric").field("data", &self.data).field("event_loop_proxy", &self.event_loop_proxy).field("os", &self.os).finish()
+  }
 }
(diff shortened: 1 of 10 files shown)


wusyong:
I think this is a really great feature to be added!
Both linux and mac have similar event callback iirc.
I can help adding them later on.

Could this just be the config of Webview? Webcontext usually describes shared behaviour (like webdriver setup).
And I think it should let users handle event loop proxy themselves. I believe the proxy can be moved into navigation call backs by users.
This should give more freedom for them.

atlanticaccent:
I could move this code into the webview fairly easily I think, including using the same type aliasing trick to reduce breakage. The reason I picked the WebContext is because it was already being used for things like the webview data directory override, which has a similar execution.

I do think that the `EventLoopProxy` needs to be explicitly passed in - otherwise the feature is somewhat pointless as a user could write the callback and no event will be emitted. An alternative approach could be to eschew emitting events entirely, and simply take a callback that will be executed when the platform specific event occurs.

atlanticaccent:
I've done some thinking about your suggestions and reworked the feature's implementation.
There `WebContext` is no longer generic over users events. As you suggested, users should now just move an event proxy into the callback closure.
`WebView` and `WebViewBuilder` can both now add the navigation callback.
There is only one callback, which is where users can define behaviour to execute when the event happens, and also whether to cancel the navigation or not - there didn't seem to much point in having the two closures as separate.
I'm still using WebContext to hold the callback, as holding it as a field on the `WebView` would still mean adding an additional parameter to all platform webview implementations, when webcontexts are already available to the WebView and are passed into all platform implementations.

src/webview/web_context.rs
```diff
@@ -63,17 +71,32 @@ impl Default for WebContext {
...
+pub trait NavCallback: Fn(String, bool) -> bool + 'static {}
+
+impl<T: Fn(String, bool) -> bool + 'static> NavCallback for T {}
+
 /// Data that all [`WebContext`] share regardless of platform.
-#[derive(Debug, Default)]
+#[derive(Default)]
 pub struct WebContextData {
   data_directory: Option<PathBuf>,
+  nav_callback: Option<Arc<dyn NavCallback>>
```
amrbashir, line 82: The navigation handler should live in `Webview` and not in `WebContextDat`, same as `ipc_handler`.

amrbashir (changes requested):
Looks good so far except the handler shouldn't be tied to `WebContext` at all. Instead, it should be on `Webview` same as the ipc handler.

atlanticaccent:
I've removed all modifications to WebContext and instead stored the navigation handler in `WebViewAttributes`

src/webview/web_context.rs
```diff
@@ -9,7 +9,7 @@ use crate::webview::webkitgtk::WebContextImpl;
 #[cfg(any(target_os = "macos", target_os = "ios"))]
 use crate::webview::wkwebview::WebContextImpl;
 
-use std::path::{Path, PathBuf};
+use std::{path::{Path, PathBuf}};
```
amrbashir:
```suggestion
use std::path::{Path, PathBuf};
```
Unnecessary parentheses.

src/webview/webview2/mod.rs
```diff
@@ -299,6 +299,49 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
...
+              }
+              
+              Ok(())
+            })),
+            &mut token
+          )
+          .map_err(webview2_com::Error::WindowsError)?;
+      }
+    }
+
```
amrbashir: This should be in another PR because it is not the same thing.

amrbashir (changes requested):
Webview2 implementation Looking good so far except two small comments below.

src/webview/webview2/mod.rs
atlanticaccent: That's fair, even though I can see users wanting pretty similar behaviour from both. I'll open a PR based on these changes that basically adds the above but using a separate `new_window` handler

atlanticaccent:
https://github.com/tauri-apps/wry/pull/519/commits/0500e74cce6b324f1b3f691a83778ad6284565ed 
Implements WebkitGtk navigation callback handling. Tested under WSL so should presumably work everywhere else

wusyong:
Here's the one for macos but I suppose it won't be easy to implement if you are not familiar with c/objc ffi.
https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455641-webview?language=objc
I could do this after I'm done with ios/android support.

atlanticaccent:
I don't really have any experience with objc ffi so if you could lend a hand after the Android port effort that would be great. I'll still try and give it a shot in my own time but I don't expect much

wusyong:
Just add the macos and it's working!
Now there are two remaining changes I would like to request:
One is switch the condition on returning `bool`. I think it's more intuitive for true to allow the url and false for not.
The other is make the callback be `Box` instead of `Rc`. Is there a place we still need multiple ownership?

I just need the agreement and I'll help pushing the commits. I know some might be busy on work days.

amrbashir:
> One is switch the condition on returning `bool`. I think it's more intuitive for true to allow the url and false for not.

Seems okay but I could see it the other way around too so we gonna have to document this. and If we want more clarity, we can use an enum
```rs
enum Navigation {
  Allow,
  Deny
}
```

> The other is make the callback be `Box` instead of `Rc`. Is there a place we still need multiple ownership?

`Box` is fine.

amrbashir (dismissed):

wusyong:
I'm very happy with the result. Thank you @atlanticaccent for the help!

PR #530, opened 2022-03-27
feat: Add download started and download completed callbacks

atlanticaccent:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Windows works completely - the callbacks fire and the file downloads. The download location can even be changed through the crude API I've implemented (a `&mut String` passed to the download started callback). One note on changing download location paths is that Universal Naming Convention file paths do not work (they start with slashes and a question mark) - these are generated when calling `canonicalize` so an alternative solution is necessary.

Linux seems to be working to some extent - both callbacks fire, however controlling where the file is saved seems to have no effect and causes the download to stall indefinitely. Not setting the download path causes bizarre behaviour where webkitgtk either claims that the download has succeeded (to where is unclear) or that the path (which is full of invalid characters) is occupied.

~~I attempted a macOS implementation however seem to have failed completely. I have a `DownloadDelegate` being created and set, and a couple of what I think are necessary methods implemented on the `WKWebview`, but nothing seems to get the actual delegate functions to fire.~~
The macOS implementation is effectively on par with the linux implementation, in that all the callbacks now fire correctly, but setting the download location causes the entire download to fail, and not setting it does not seem to actually write out a download/file.


src/webview/mod.rs
```diff
@@ -126,6 +126,19 @@ pub struct WebViewAttributes {
...
+  /// The first closure takes two parameters - the first is a `String` representing the url being downloaded from and and the 
+  /// second is a mutable `String` reference that (possibly) represents where the file will be downloaded to. The latter 
+  /// parameter can be used to set the download location by assigning a new path string to it - the assigned path _must_ be 
+  /// absolute, and (on Windows) cannot include a UNC prefix. The closure returns a `bool` to allow or deny the download.
+  /// The second closure is fired when the download completes, with a `String` representing the path to where the download was
+  /// saved and a `bool` indicating if the download succeeded.
+  pub download_handlers: Option<(
+    Box<dyn FnMut(String, &mut String) -> bool>,
+    Box<dyn Fn() -> Box<dyn Fn(String, bool) + 'static> + 'static>
+  )>,
```
amrbashir: Should be split into two handlers, `download_handler` and `download_completed_callback`

src/webview/webkitgtk/web_context.rs
```diff
@@ -198,6 +207,16 @@ impl WebContextExt for super::WebContext {
...
   }
+
+  fn register_download_handler(
+    &mut self,
+    handler: (
+      Box<dyn FnMut(String, &mut String) -> bool>,
+      Box<dyn Fn() -> Box<dyn Fn(String, bool) + 'static> + 'static>,
+    ),
+  ) {
+    actually_register_download_handler(self, handler);
```
amrbashir: just inline the function here, no need to extract the logic in a separate function

src/webview/mod.rs
```diff
@@ -126,6 +126,19 @@ pub struct WebViewAttributes {
...
+  ///
+  /// The first closure takes two parameters - the first is a `String` representing the url being downloaded from and and the 
+  /// second is a mutable `String` reference that (possibly) represents where the file will be downloaded to. The latter 
+  /// parameter can be used to set the download location by assigning a new path string to it - the assigned path _must_ be 
+  /// absolute, and (on Windows) cannot include a UNC prefix. The closure returns a `bool` to allow or deny the download.
+  /// The second closure is fired when the download completes, with a `String` representing the path to where the download was
+  /// saved and a `bool` indicating if the download succeeded.
+  pub download_handlers: Option<(
+    Box<dyn FnMut(String, &mut String) -> bool>,
+    Box<dyn Fn() -> Box<dyn Fn(String, bool) + 'static> + 'static>
```
amrbashir: why does this need to be a function that returns a function?

src/webview/mod.rs
```diff
@@ -126,6 +126,19 @@ pub struct WebViewAttributes {
...
+  /// Set a download handlers to manage incoming downloads.
+  ///
+  /// The first closure takes two parameters - the first is a `String` representing the url being downloaded from and and the 
+  /// second is a mutable `String` reference that (possibly) represents where the file will be downloaded to. The latter 
+  /// parameter can be used to set the download location by assigning a new path string to it - the assigned path _must_ be 
+  /// absolute, and (on Windows) cannot include a UNC prefix. The closure returns a `bool` to allow or deny the download.
+  /// The second closure is fired when the download completes, with a `String` representing the path to where the download was
+  /// saved and a `bool` indicating if the download succeeded.
+  pub download_handlers: Option<(
+    Box<dyn FnMut(String, &mut String) -> bool>,
```
amrbashir: Probably better to use the `Path` type
```suggestion
    Box<dyn FnMut(String, &mut Path) -> bool>,
```

amrbashir (changes requested):

src/webview/mod.rs
atlanticaccent: I'm probably not going to explain this well, but I'll give it a shot:

If I just use a regular function, we run into the issue that (I think) we need a mutable copy of the download completed callback _every time the download started function is called_. To do that with a "raw" closure would mean cloning the closure, but I can't find a way to define the closure as cloneable when storing it - `Fn` closures are cloneable if all their captured variables are cloneable, but I can't see a way of specifying that.

An alternative solution (again, I think) would be to wrap the "raw" closure in an `Rc`, which would make it as cloneable as we like.

I can't actually tell if this is the same stories on other platforms, but the fact it's the case on windows basically makes it unavoidable.

amrbashir: Yeah `Rc` should work, it is better than having this level of indirection. cc @wusyong

atlanticaccent: Alternatively this https://stackoverflow.com/a/65204236 may be a viable solution, though it means implementing a relatively complex custom trait that may make the API more opaque to users

src/webview/mod.rs
atlanticaccent: Switched to using `&mut PathBuf` - `Path` size can't be known at compile time so we can't take a mutable reference.

src/webview/mod.rs
atlanticaccent: I'll use the `Rc` solution as it seems the simplest

src/webview/mod.rs
amrbashir: Yeah that's fine too.

keiya01:
What is this PR status?

lucasfernog:
What's the status here?

atlanticaccent:
I'm attempting to get the work I had done fixed up so it's up to date with master, however the outstanding issues in linux (and I assume macOS) still remain. Specifically, downloads don't complete and modifying the target file path makes no difference (or causes the download to fail I haven't narrowed it down).
I may be able to get the macOS version up to date soon, but macOS still isn't my primary environment

atlanticaccent:
The macOS implementation is now fixed, and the entire PR should be up to date with master (barring any mistakes made while merging). I have also managed to fix the macOS issue where downloads would spuriously fail. I still don't have a proper linux setup though, so can't test that platform fully.

There's a couple things that still need done. Completed handlers still need to be completely decoupled from started handlers on all platforms so that they fire regardless of whether a started handler is present. I also think that the completed handler should have the URI of the downloaded resource added as a parameter, as otherwise there is no reliable way to tell what download the handler is actually being fired for.

atlanticaccent:
@lucasfernog @amrbashir

amrbashir:
> I still don't have a proper linux setup though, so can't test that platform fully.

don't worry, we can test it later once you're finished with the PR

> I also think that the completed handler should have the URI of the downloaded resource

yeah that would be a good addition if it is possible

atlanticaccent:
> > I also think that the completed handler should have the URI of the downloaded resource

> yeah that would be a good addition if it is possible

I'm actually thinking that it might even be worth generating some kind of unique identifier to attach to downloads (like a GUID) as URIs are not actually a completely unique identifier for download operations - say, for example, two downloads of the same resource are triggered in quick succession. In the example, it's also not sufficient to assume the first download to complete was the download that was started first, as it may have stalled for any number of reasons whilst the download initiated second completed first. An alternative to GUIDs would be something like attaching the system time the download was initiated at.

amrbashir:
GUIDs is a bit overkill. I think these download handlers could be used to spawn a dialog for the user to choose the file path and so it should be the download handler responsibility to avoid conflicts

src/webview/mod.rs
```diff
@@ -160,6 +160,31 @@ pub struct WebViewAttributes {
...
+  ///
+  /// The closure takes two parameters - the first is a `String` representing the url being downloaded from and and the
+  /// second is a mutable `PathBuf` reference that (possibly) represents where the file will be downloaded to. The latter
+  /// parameter can be used to set the download location by assigning a new path to it - the assigned path _must_ be
+  /// absolute. The closure returns a `bool` to allow or deny the download.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows:** The assigned path cannot include a UNC (Universal Naming Convention) prefix. This prefix is
+  /// sometimes included when canonicalizing paths in order to differentiate between networked filesystems.
```
amrbashir: We can just use `dunce` crate to strip UNC from the path after it has been modified by the handler

amrbashir (changes requested):

atlanticaccent:
@amrbashir I believe I've addressed your comments, plus cleaned some things up. Would you be able to give it another look over?

src/webview/mod.rs
```diff
@@ -160,6 +160,26 @@ pub struct WebViewAttributes {
...
+  ///
+  /// The closure is fired when the download completes, whether it was successful.
+  /// The closure takes a `String` representing the URL of the original download request, a `String` representing the
+  /// filesystem path the file was downloaded to (if successful), and a `bool` indiciating if the download succeeded.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **macOS**: The second parameter indicating the path the file was saved to is always empty, due to API
+  /// limitations.
+  pub download_completed_handler: Option<Rc<dyn Fn(String, String, bool) + 'static>>,
```
amrbashir: could we change this also to provide `PathBuf`
```suggestion
  pub download_completed_handler: Option<Rc<dyn Fn(String, PathBuf, bool) + 'static>>,
```

src/webview/mod.rs
```diff
@@ -416,6 +438,38 @@ impl<'a> WebViewBuilder<'a> {
...
+  /// The closure takes a `String` representing the URL of the original download request, a `String` representing the
+  /// filesystem path the file was downloaded to (if successful), and a `bool` indiciating if the download succeeded.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **macOS**: The second parameter indicating the path the file was saved to is always empty, due to API
+  /// limitations.
+  pub fn with_download_completed_handler(
+    mut self,
+    download_completed_handler: impl Fn(String, String, bool) + 'static,
```
amrbashir: ditto
```suggestion
    download_completed_handler: impl Fn(String, PathBuf, bool) + 'static,
```

amrbashir (changes requested):
Other than these two nit-picks, the Linux/Windows implementation looks good, as for macOS, I will leave it for @wusyong or @lucasfernog to approve.

amrbashir (dismissed):
LGTM now for Windows and Linux

src/webview/wkwebview/mod.rs
```diff
@@ -329,66 +336,240 @@ impl InnerWebView {
...
+      let (navigation_decide_policy_ptr, download_delegate) = if attributes
+        .navigation_handler
+        .is_some()
+        || attributes.new_window_req_handler.is_some()
+        || attributes.download_started_handler.is_some()
+      {
+        let cls = match ClassDecl::new("UIViewController", class!(NSObject)) {
+          Some(mut cls) => {
+            cls.add_ivar::<*mut c_void>("function");
+            cls.add_ivar::<*mut c_void>("HasDownloadHandler");
```
amrbashir, line 435: Can we extract these two strings into some consts so it is easier to track/change in the future? Actually any ivar used more than one time should be in a const

amrbashir (changes requested):
Sorry for taking so long to merge this PR, our macOS expert is currently on vacation

src/webview/wkwebview/mod.rs
wusyong: There are many more ivar like this. I think we should do this in another PR.

wusyong (dismissed):

amrbashir (approved):

wusyong:
Thanks for this amazing PR @atlanticaccent !

PR #556, opened 2022-04-24
fix: reset background process when webview is closed, closes #536

### LINKED ISSUE #536: Even if I delete it, WebView (or other?) remains.
**Describe the bug**
Let's say you have a program that creates a window to play videos.
And when I created multiple of them and then deleted some, the video in the browser that I had deleted would still be playing.

**Steps To Reproduce**
Prepare code of `wry/examples/multi_window.rs`.
Add a below snipet (YouTube embed) to [here str](https://github.com/tauri-apps/wry/blob/313eaea0ff123bddbc8b5c337ded05d464d3dfaa/examples/multi_window.rs#L52).
```html
<iframe class="background" id="background" 
src="https://www.youtube.com/embed/JUGHD_YWOe4?iv_load_policy=3&amp;loop=1&amp;playlist=JUGHD_YWOe4&amp;control=0&amp;fs=0&amp;mute=On&amp;end=111&amp;modestbranding=1&amp;start=11&amp;autoplay=1" 
frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; 
picture-in-picture" allowfullscreen="" style="width: 1191px; height: 678px;"></iframe>
```
Create about five windows every almost few seconds.
After a while, when it gets too noisy, close the window a little at a time and try to get only one.
Then it remains noisy.
So the closed window is still in motion.

**Expected behavior**
To be successfully closed.

**Screenshots**
Nothing

OS: macOS 12.3
Rustc: 1.59.0

- [ ] Yes
- [x] No

**Additional context**
Nothing

keiya01:
I fixed issue which didn't stop video when multiple window is closed.
I think this issue was caused by [WKProcessPool](https://developer.apple.com/documentation/webkit/wkprocesspool?language=objc) is singleton.
`WKProcessPool` is not reset even if wkwebview is deallocated.
And I could not find a feature to resolve this problem.
So I think we need to override the process.
But I think this is workaround. If you know a correct way, I will adopt your way.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/webview/wkwebview/mod.rs
+++ b/src/webview/wkwebview/mod.rs
@@ -621,6 +621,11 @@ impl Drop for InnerWebView {
         }
       }
 
+      // WKWebview has a single WKProcessPool to manage web contents.
+      // The WKProcessPool is not reset even if WKWebview is deallocated.
+      // So we need to override the process by navigating to `about:blank`.
+      self.navigate("about:blank");
+
       let _: Id<_> = Id::from_ptr(self.webview);
       #[cfg(target_os = "macos")]
       let _: Id<_> = Id::from_ptr(self.ns_window);
--- a/.changes/stop-background-process.md
+++ b/.changes/stop-background-process.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix to reset process on MacOS when webview is closed, closes #536.


src/webview/wkwebview/mod.rs
```diff
@@ -621,6 +621,8 @@ impl Drop for InnerWebView {
         }
       }
 
+      self.navigate("about:blank");
```
wusyong, line 627: I couldn't find any other method exposed by wkwebview either.
This one looks good to me.
Could you add some comments here? So we could understand why it has to navigate to blank page before dropping.

keiya01: You are right. I added comment.
Thank you for reviewing.

wusyong (approved):

PR #558, opened 2022-04-25
fix: disalbe bounce option for macos, closes #557

### LINKED ISSUE #557: Set scroll bounce on WKWebView
MacOS has this bounce effect when you scroll. I would like to disable it. I think it is `webView.scrollView.bounces = NO;`. I think the default should be `false`, like it is in for example Electron.

Tried to do this in CSS, but I couldn't find a working solution so far. I've read that adding a certain class to all of the elements in the DOM fixes it, or applying some JS, but it is a real hassle. Most people suggest just setting the bounces property to NO.

- [ ] Yes
- [x] No

**Additional context**
None

yuta24:
I fixed https://github.com/tauri-apps/wry/issues/557.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

For compatibility, bounce is On by default.

### DIFF
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -65,6 +65,8 @@ pub struct WebViewAttributes {
   pub visible: bool,
   /// Whether the WebView should be transparent. Not supported on Windows 7.
   pub transparent: bool,
+  /// Whether the WebView should be bounce. Supported on macOS only.
+  pub bounce: bool,
   /// Whether load the provided URL to [`WebView`].
   pub url: Option<Url>,
   /// Whether load the provided html string to [`WebView`].
@@ -160,6 +162,7 @@ impl Default for WebViewAttributes {
       user_agent: None,
       visible: true,
       transparent: false,
+      bounce: true,
       url: None,
       html: None,
       initialization_scripts: vec![],
@@ -206,6 +209,12 @@ impl<'a> WebViewBuilder<'a> {
   /// Sets whether the WebView should be transparent.
   pub fn with_visible(mut self, visible: bool) -> Self {
     self.webview.visible = visible;
+    self
+  }
+
+  /// Sets whether the WebView should be bounce. Supported on macOS only.
+  pub fn with_bounce(mut self, bounce: bool) -> Self {
+    self.webview.bounce = bounce;
     self
   }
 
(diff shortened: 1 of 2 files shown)


probablykasper:
Does bouncing mean rubber band scrolling?

nothingismagick:
I thought there was a CSS way to do this too...
```css
body {
  height: 100%;
  padding: 0;
  overflow: auto;
  margin: 0;
  -webkit-overflow-scrolling: touch;
}
```

nothingismagick:
or this (if you manage scrolling internally in the layout):

```css
body {
  position: fixed;
}
```

wusyong:
Thanks you for the PR! But I dont think this will work.
WkWebview on macOS doesn't have scroll view somehow.
https://stackoverflow.com/questions/39929552/disable-bounce-scrolling-for-wkwebview-in-macos

Did it solve the issue on your side?

yuta24:
> Does bouncing mean rubber band scrolling?

Yes, If webview's content size is over the window size, rubber band scrolling is good. But, the app using wry always enable rubber band scrolling. I thought it would be better if wry had an enable/disable option.

> I thought there was a CSS way to do this too...
> body {
>  height: 100%;
>  padding: 0;
>  overflow: auto;
>  margin: 0;
>  -webkit-overflow-scrolling: touch;
> }

It wasn't working on my project.

> WkWebview on macOS doesn't have scroll view somehow.
> https://stackoverflow.com/questions/39929552/disable-bounce-scrolling-for-wkwebview-in-macos
> 
> Did it solve the issue on your side?

Sorry. This PR's solution works in specific situations. I'll close this PR.

nothingismagick:
don't be hasty - Tauri has become a collection of edge-cases, and some might find it useful as an option.

probablykasper:
I get horizontal rubber band scrolling, and it's always an annoyance with no benefit. The CSS above doesn't fix it, but the PR does. `-webkit-overflow-scrolling` isn't supported for me on macOS 10.15.

Is there any reason to actually keep `bounce` on by default?

evoactivity:
For future googlers, you also need to disable overflow on the html element, eg

```css
html {
  height: 100vh;
  width: 100vw;
  overflow: hidden;
}

body {
  height: 100%;
  padding: 0;
  overflow: auto;
  margin: 0;
  -webkit-overflow-scrolling: touch;
}
```

PR #587, opened 2022-05-15
fix: prevent memory leak on macOS, closes #536

### LINKED ISSUE #536: Even if I delete it, WebView (or other?) remains.

keiya01:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/webview/wkwebview/mod.rs
+++ b/src/webview/wkwebview/mod.rs
@@ -645,15 +645,8 @@ impl Drop for InnerWebView {
         }
       }
 
-      // WKWebview has a single WKProcessPool to manage web contents.
-      // The WKProcessPool is not reset even if WKWebview is deallocated.
-      // So we need to override the process by navigating to `about:blank`.
-      self.navigate("about:blank");
-
-      let _: Id<_> = Id::from_ptr(self.webview);
-      #[cfg(target_os = "macos")]
-      let _: Id<_> = Id::from_ptr(self.ns_window);
-      let _: Id<_> = Id::from_ptr(self.manager);
+      let _: Id<_> = Id::from_retained_ptr(self.webview);
+      let _: Id<_> = Id::from_retained_ptr(self.manager);
     }
   }
 }
--- a/.changes/memleak.md
+++ b/.changes/memleak.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Prevent memory leak on macOS.


lucasfernog:
I think you meant removing the script message handler for the IPC? https://github.com/tauri-apps/wry/blob/494a11057f9ddd2bf4bcecdc96b43ed95c5bd08e/src/webview/wkwebview/mod.rs#L307

keiya01:
This PR is based on https://stackoverflow.com/questions/58454042/wkwebview-instances-and-memory-leak.
I'm sorry I don't know detail about this solution but we need to attach webview to scriptMessageHandler. And we can reduce memory usage when I add and remove window(webview).
But maybe this improvement is occurred only on my macOS environment, so please check it on your environment.

keiya01:
Maybe we can also use [removeAllScriptMessageHandlers](https://developer.apple.com/documentation/webkit/wkusercontentcontroller/3585114-removeallscriptmessagehandlers?language=objc).

lucasfernog:
Sounds like exactly what we need :)

src/webview/wkwebview/mod.rs
```diff
@@ -476,6 +476,10 @@ r#"Object.defineProperty(window, 'ipc', {
         let ns_window = window.ns_window() as id;
         let _: () = msg_send![ns_window, setContentView: webview];
 
+        // To prevent memory leak, we need to remove webview by script message handler.
+        let interop = NSString::new("interop");
+        let _: () = msg_send![manager, addScriptMessageHandler:webview name:interop];
```
wusyong: I feel like this should be the ipc [handler](https://github.com/tauri-apps/wry/blob/632d5439f8e1717cd86cb3d5bb49bb07a3d13545/src/webview/wkwebview/mod.rs#L302). The type of `addScriptMessageHandler` is `id<WKScriptMessageHandler>`. This is exact protocol WebViewDelegate has.

src/webview/wkwebview/mod.rs
```diff
@@ -650,6 +654,8 @@ impl Drop for InnerWebView {
       // So we need to override the process by navigating to `about:blank`.
       self.navigate("about:blank");
 
+      let _: () = msg_send![self.manager, removeAllScriptMessageHandlers];
```
FabianLars: This is pretty similar to what i tried, but it has the same problem. If you try to close a window the app straight up panics
> unrecognized selector sent to instance

Idk from my testing it seemed like doing it in `drop` was too late :thinking:

keiya01: hmm, this is not occurred in my environment...
Could you show me to reproduce code?

FabianLars: You can use this exact code (a fresh clone of your fork panics on my end too).
The easiest to use is wry's multi_window example and spawning and closing a window immediately causes a panic with a reference to this line of code.

FabianLars: Aaahh i see why. This api was added in macos 11 and i primarily tested in 10.15

FabianLars: Soo changing it back to `removeScriptMessageHandlerForName` doesn't panic anymore (it did in my first tests, guess i was doing something wrong.). But neither your PR nor changing it to remove `"ipc"` has an effect.

keiya01: Oh I see.
Thank you for your description :)
I will fix it.

src/webview/wkwebview/mod.rs
```diff
@@ -645,15 +645,8 @@ impl Drop for InnerWebView {
...
-      // WKWebview has a single WKProcessPool to manage web contents.
-      // The WKProcessPool is not reset even if WKWebview is deallocated.
-      // So we need to override the process by navigating to `about:blank`.
-      self.navigate("about:blank");
-
-      let _: Id<_> = Id::from_ptr(self.webview);
-      #[cfg(target_os = "macos")]
-      let _: Id<_> = Id::from_ptr(self.ns_window);
-      let _: Id<_> = Id::from_ptr(self.manager);
+      let _: Id<_> = Id::from_retained_ptr(self.webview);
```
keiya01, line 648: I found the `Id::from_ptr` is not working well, so id object is not dropped.
I fixed these code to use `Id::from_retained_ptr` with reference to [rust-objc-id example](https://github.com/SSheldon/rust-objc-id).
If we use `Id::from_retained_ptr`, navigation to `about:blank` is not needed :)

keiya01 (commented):
I changed all of commit because I found new solution 🙇

src/webview/wkwebview/mod.rs
```diff
@@ -645,15 +645,8 @@ impl Drop for InnerWebView {
...
       }
 
-      // WKWebview has a single WKProcessPool to manage web contents.
-      // The WKProcessPool is not reset even if WKWebview is deallocated.
-      // So we need to override the process by navigating to `about:blank`.
-      self.navigate("about:blank");
-
-      let _: Id<_> = Id::from_ptr(self.webview);
-      #[cfg(target_os = "macos")]
-      let _: Id<_> = Id::from_ptr(self.ns_window);
```
keiya01, line 655: I think ns_window is dropped in tao layer, so we don't need to drop it.

wusyong: We control its drop here iirc. Other `Rc<Window>`s are held by all hander/callbacks which we just free above. cc @lucasfernog

src/webview/wkwebview/mod.rs
FabianLars: Yep, tested it and it indeed cleans up the process we were talking about correctly! nice :partying_face: 

What it still doesn''t cleanup is the Networking process, but since it does the same on Linux it looks like a webkit problem :thinking:

src/webview/wkwebview/mod.rs
lucasfernog: tauri doesn't hold the rc IIRC btw, do we need to do something?

src/webview/wkwebview/mod.rs
keiya01: I think network connection is closed but network memory is still remained. I think this is because networking process is single process, networking memory is shared between multiple webview 👀

src/webview/wkwebview/mod.rs
wusyong: No, I just think we still need to convert `self.ns_window` here IMHO.
`WebView` should own the `Window` and also release it when it's dropped.

src/webview/wkwebview/mod.rs
keiya01: > This process is responsible for handling network requests as well as storage management. All WebContent processes in a single session (default vs. private browsing) share a single networking session in the networking process.
> https://github.com/WebKit/WebKit/blob/main/Introduction.md

This is webkit description.
I think we can [remove cache](https://developer.apple.com/documentation/webkit/wkwebsitedatastore/1532938-removedataoftypes?language=objc) but all of webview cache will be removed. It will bring to bad page loading performance.

keiya01: I will try to use [this gist](https://gist.github.com/dakeshi/6915b6681cbc29327dcc9b27cef263b8) to remove cache in another PR.

src/webview/wkwebview/mod.rs
wusyong: Just tested with removing this line and it seems fine. But I remembered we added these because of weird behavior in some combination when closing windows. Maybe take an eye on Tauri after this PR.

wusyong (dismissed):

wusyong (approved):
@keiya01 LGTM. Do you want to merge this now?
I suppose networking process will be another PR right? Could you choose an issue and make it the main tracking issue?
Also you can just add a branch in this repo next time.

keiya01:
Yes, I want to merge this PR now.
I will try to resolve https://github.com/tauri-apps/wry/issues/590 in another PR.

PR #592, opened 2022-05-20
Update windows-rs to 0.37.0 and webview2-com to 0.16.0 to match

wravery:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Starting as a draft PR because I still have a patch override in `Cargo.toml` to pull the matching version of `tao` in https://github.com/tauri-apps/tao/pull/400.

As in that PR, updating to this version of `windows-rs` raises the MSRV on Windows to 1.61, which was just released to the `stable` channel today. 🎉

### DIFF
--- a/.changes/windows-0.37.0.md
+++ b/.changes/windows-0.37.0.md
@@ -2,6 +2,6 @@
 "wry": patch
 ---
 
-Update the windows-rs crate to the latest 0.37.0 release and webview2-com to 0.16.0 to match.
+Update the `windows` crate to the latest 0.37.0 release and `webview2-com` to 0.16.0 to match.
 
-This version of windows-rs depends on rustc version 1.61 for some `const` generic support which was just stabilized, so on Windows the MSRV is effectively 1.61 now.
+The `#[implement]` macro in `windows-implement` depends on `const` generic features which were just stabilized in `rustc` version 1.61, so this change also adds a `rust-version` attribute to the manifest setting the MSRV to 1.61.
--- a/.changes/windows-0.37.0.md
+++ b/.changes/windows-0.37.0.md
@@ -2,6 +2,8 @@
 "wry": patch
 ---
 
-Update the windows-rs crate to the latest 0.37.0 release and webview2-com to 0.16.0 to match.
+Update the `windows` crate to the latest 0.37.0 release and `webview2-com` to 0.16.0 to match.
 
-This version of windows-rs depends on rustc version 1.61 for some `const` generic support which was just stabilized, so on Windows the MSRV is effectively 1.61 now.
+The `#[implement]` macro in `windows-implement` and the `implement` feature in `windows` depend on some `const` generic features which stabilized in `rustc` 1.61. The MSRV on Windows targets is effectively 1.61, but other targets do not require these features.
+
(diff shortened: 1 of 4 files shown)


amrbashir (dismissed):

wravery:
The reason the last CI failed is just that I removed the `[patch.crates-io]` table in https://github.com/tauri-apps/wry/pull/592/commits/309fffa76734c2ac32e87f07c1fd43472bd265a1 to prepare this PR for merging. Once https://github.com/tauri-apps/tao/pull/400 is merged and a new version of `tao` is published, CI for this PR should start passing again.

wravery:
The changes in https://github.com/tauri-apps/wry/pull/592/commits/b63db59bb2361c3bc30fa0a469046cab9826a10e and https://github.com/tauri-apps/wry/pull/592/commits/181a8272b16120b0b5111914180964a6b267f7ed should make it so the next time we update `windows`, we don't need to stage the PRs like this. Once `tao` no longer requires a matching `HWND` type (in https://github.com/tauri-apps/tao/commit/793c1e6515988f1bbd17aac955a11a26c805cdf3) and `wry` includes all of the features it needs from `windows` instead of inheriting some of them from `tao`, they should be able to build with independent versions of `windows`.

wravery:
I undid the changes in https://github.com/tauri-apps/wry/pull/592/commits/9855a8510bd787555b3041e9b82b44af397ba679 (bumping the MSRV for `wry` on all platforms) and documented it in the change log: https://github.com/tauri-apps/wry/pull/592/commits/6e4e0ee5100d9879688487a2397512497e1fe0c2.

amrbashir (dismissed):

wusyong:
@wravery  We just published tao 0.9!

lucasfernog:
The change looks ok, but the file drop path reader is wrong. The `DragQueryFileW(hdrop, i, &mut path_buf)` call is accessing random memory it seems. cc @wravery

wravery:
> The change looks ok, but the file drop path reader is wrong. The `DragQueryFileW(hdrop, i, &mut path_buf)` call is accessing random memory it seems. cc @wravery

Ah, I think this is because it's not actually sizing the `Vec` until after the call (trying to avoid initializing the out-param). The signature change where it takes a slice instead of a pointer + length means that the slice is length 0.

It should probably pass `std::mem::transmute(path_buf.spare_capacity_mut())` instead of `&mut path_buf`. I need to check if the equivalent pattern exists anywhere else in here or in the corresponding `tao` changes.

lucasfernog:
I can confirm the `std::mem::transmute(path_buf.spare_capacity_mut())` change fixes it.

lucasfernog (dismissed):

wravery:
> I need to check if the equivalent pattern exists anywhere else in here or in the corresponding `tao` changes.

I started a [bugfix PR](https://github.com/tauri-apps/tao/pull/401) in `tao` for the same issue. Other than that, it looks like there are no more references to `with_capacity` -> `&mut slice` in either project, so I think it was just this one line that was affected in each of them.

PR #593, opened 2022-05-22
feat: add find-in-page, closes #585

### LINKED ISSUE #585: Is there a way to find string in web content？
I want to find sowe string in web content like [this api](https://developer.apple.com/documentation/webkit/wkwebview/3650493-find)

I thik we can use [this api](https://developer.apple.com/documentation/webkit/wkwebview/3650493-find) on macOS.
I will search about other os.

Implemet it with rust or javascript by myself.

- [x] Yes
- [ ] No

**Additional context**

keiya01:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/add-find-in-page.md
+++ b/.changes/add-find-in-page.md
@@ -2,4 +2,4 @@
 "wry": minor
 ---
 
-Add find-in-page feature.
+Add `WebView::find_in_page`.
--- a/.changes/add-find-in-page.md
+++ b/.changes/add-find-in-page.md
@@ -1,0 +1,5 @@
+---
+"wry": minor
+---
+
+Add find-in-page feature.
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -551,7 +551,11 @@ pub struct FindInPageOption {
   pub case_sensitive: bool,
   /// Wrap around to the other side of the page.
   pub wraps: bool,
-  /// For Linux only, Maximum number of search strings to highlight.
+  /// Maximum number of search strings to highlight.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// **Windows / macOS / Android / iOS**: Unsupported.
   pub max_match_count: u32,
 }
 
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -542,7 +542,7 @@ impl WebView {
   }
 }
 
-/// A configuration for `find_in_page`.
+/// A configuration for [`WebView::find_in_page`].
 #[derive(Default)]
 pub struct FindInPageOption {
   /// Indicate the search direction.
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -517,6 +517,20 @@ impl WebView {
     self.webview.zoom(scale_factor);
   }
 
+  /// Searches for the specified string in the WebView content.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Windows / Android**: Not supported.
+  /// - **macOS**: available on macOS 10.15.4+ only.
+  /// - **iOS**: available on iOS 13.4+ only.
+  pub fn find_in_page<F>(&self, string: String, option: FindInPageOption, f: F)
+  where
+    F: Fn(bool) + 'static,
+  {
+    self.webview.find_in_page(string, option, f);
+  }
+
   #[cfg(target_os = "android")]
   pub fn run(self, env: JNIEnv, jclass: JClass, jobject: JObject) -> jobject {
     self.webview.run(env, jclass, jobject).unwrap()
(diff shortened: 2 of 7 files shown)


amrbashir (commented):
I don't know how this feature might be useful for desktop apps tbh. It only returns whether a match has been found or not, it doesn't return the matched word nor the html element that contains it. I can't think of a use case for it at all.

keiya01:
This feature highlight the search text automatically.
Also electron has [this feature](https://www.electronjs.org/docs/latest/api/web-contents#contentsfindinpagetext-options).
Searching word on browser by using JavaScript is too heavy process. And I think we should provide this feature from wry if it is possible.

amrbashir:
Ah I see, I didn't know it highlights it too. I see that electron has this event https://www.electronjs.org/docs/latest/api/web-contents#event-found-in-page with information about the matched text, can we provide some of these?

keiya01:
I want to provide these information but I can't find API to do it :(

keiya01:
Additionally `wkwebview.findString` API has restriction that text is not highlighted only on macOS (but text is selected like `window.find`).
This is the issue in webkit: https://bugs.webkit.org/show_bug.cgi?id=140210.

.changes/add-find-in-page.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": minor
+---
+
+Add find-in-page feature.
```
amrbashir:
```suggestion
Add `WebView::find_in_page`.
```

src/webview/mod.rs
```diff
@@ -517,6 +517,20 @@ impl WebView {
     self.webview.zoom(scale_factor);
   }
 
+  /// Searches for the specified string in the WebView content.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Windows / Android**: Not supported.
```
```suggestion
  /// - **Windows / Android**: Unsupported.
```

src/webview/mod.rs
```diff
@@ -528,6 +542,19 @@ impl WebView {
   }
 }
 
+/// A configuration for `find_in_page`.
```
```suggestion
/// A configuration for [`WebView::find_in_page`].
```

src/webview/mod.rs
```diff
@@ -528,6 +542,19 @@ impl WebView {
...
+#[derive(Default)]
+pub struct FindInPageOption {
+  /// Indicate the search direction.
+  pub backwards: bool,
+  /// Match the search string in a case-sensitive.
+  pub case_sensitive: bool,
+  /// Wrap around to the other side of the page.
+  pub wraps: bool,
+  /// For Linux only, Maximum number of search strings to highlight.
```
```suggestion
  /// Maximum number of search strings to highlight.
  ///
  /// ## Platform-specific:
  ///
  /// **Windows / macOS / Android / iOS**: Unsupported.
```

src/webview/webkitgtk/mod.rs
```diff
@@ -390,6 +399,40 @@ impl InnerWebView {
...
+    }
+    if option.wraps {
+      flags |= FindOptions::WRAP_AROUND;
+    }
+
+    self
+      .find_controller
+      .search(&string, flags.bits(), option.max_match_count);
+
+    let handler = Rc::new(WithClosure { f: Box::new(f) });
```
amrbashir: Why wrap in a struct? can't you just do?
```suggestion
    let handler = Rc::new(Box::new(f));
```

amrbashir (changes requested):
I sill don't see much value in this API but I will leave that for @wusyong to decide and I will just do a review.

src/webview/webkitgtk/mod.rs
keiya01: Thank you for your review.
I fixed your review.

amrbashir (approved):

wusyong:
> Also electron has [this feature](https://www.electronjs.org/docs/latest/api/web-contents#contentsfindinpagetext-options).

If electron has it, it should be possible for webview2 perhaps (since they are both chromium)? This method will be really awkward if it doesn't support Windows which is the major desktop most users use

keiya01:
Webview2 builtin these feature with search box. But [webview2 doesn't provide the API to find text in page](https://github.com/tauri-apps/wry/issues/585#issuecomment-1126718493) :(

amrbashir:
Maybe worth opening a feature request at webview2 feedback repo and hold this PR  until it happens?

rhysd:
How about using [`window.find`](https://developer.mozilla.org/en-US/docs/Web/API/Window/find) through `evaluate_script` as a fallback for WebView2?

> I sill don't see much value in this API

Many projects are using the same functionality in Electron.
https://github.com/search?l=TypeScript&q=findInPage+electron&type=Code

This functionality is useful for some document viewer to search its content.

Issue #608, opened 2022-06-14
Is it necessary to destroy window when dropping wry::webview::WebView ?


bg5hfc:
I noticed that the WebView struct implements the Drop trait, and a destroy method is called on `window` when the webview drops(in wry/src/webview/mod.rs).

Is it really necessay because tao::Window has already do the destruction properly on both Linux/mac and Windows platform.

I'm asking this because I wanna build a window with multiple `WebView`s on it (something like the browser with many tabs), but the Drop trait of WebView hinders me.

wusyong:
We stick to one window with one webview for simplicity.
We can evaluate multiple webviews support.

keiya01:
I agree.
I think window sould be dropped myself in tao layer.

keiya01:
If we need to drop window in wry, I think we need to check if window has multi webview🤔

amrbashir:
As it stated above, our current api is towards one window with one webview. I am gonna close this as duplicate of #677 as it will be the same fix in the future.

PR #613, opened 2022-06-15
feat: impl `Deref` for `WebView`

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -48,7 +48,7 @@ use webview2_com::Microsoft::Web::WebView2::Win32::ICoreWebView2Controller;
 #[cfg(target_os = "windows")]
 use windows::{Win32::Foundation::HWND, Win32::UI::WindowsAndMessaging::DestroyWindow};
 
-use std::{path::PathBuf, rc::Rc};
+use std::{ops::Deref, path::PathBuf, rc::Rc};
 
 use url::Url;
 
@@ -410,6 +410,13 @@ impl Drop for WebView {
     unsafe {
       DestroyWindow(HWND(self.window.hwnd() as _));
     }
+  }
+}
+
+impl Deref for WebView {
+  type Target = Window;
+  fn deref(&self) -> &Self::Target {
+    &self.window
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/webview/mod.rs
```diff
@@ -413,6 +413,13 @@ impl Drop for WebView {
   }
 }
 
+impl Deref for WebView {
+  type Target = Window;
```
keiya01, line 417: I think this is not intuitive.
But is there something context?

amrbashir: the webview takes ownership of the window, and accessing the window methods requires calling `.window()` method on the webview, so this is deref implementation is just for convenience and since the webivew and the window is 1:1 related, it should be ok.

keiya01: Is exporting different type used usually?
I have not ever seen this pattern.
But I think it makes sense if there is usecase for user.

amrbashir: > Is exporting different type used usually?

Yes, it is actually a common pattern, some examples in rust std: `Arc`, `Rc`, `MutexGuard` and `Box`, they all implement `Deref` or some times `DerefMut` into the inner type.

keiya01: Sure. But although these types have inner type explicitly like T of Rc<T>, webview type has window as struct field.
These are only a little difference, so it may not be carefully😅

wusyong: There's an issue #608 about multiple webview in one window. Do we decide it's going to be one webview one window forever?

amrbashir: This only for current implementation, once we implement multiple webview per window, we can reevaluate this.

wusyong: I'm sorry but I still don't think this is a good practice. `Deref` trait is used for dereferencing. `Window` and `WebView` are still two different types IMHO.
I think in prepare of multi webview in the future, we could probably propose a `WebViewWindow` type and it then has valid point to use `Deref` to `Widnow`.

amrbashir: sounds reasonable

PR #640, opened 2022-07-25
fix: set system language to webview on windows, closes #442

### LINKED ISSUE #442: better l18n support in webview (navigator.language)
> This issue has been upstreamed from https://github.com/tauri-apps/tauri/issues/2735 

I want to rely on `navigator.language` or `navigator.languages` in the webview to determine which language a user prefers and to then translate the app.

On WSL 2 / Ubuntu 20 the language is just "c", on Windows 10 with WebView2 it does not respect language settings on the system or browser (Edge) level, on native Ubuntu 20 it had a language that was actually more accurate then the system setting, but did not respect the system setting as well. I don't have access to a macos device.

Ideally `navigator.language` or `navigator.languages` contain languages that match the system level user preferences.

One could call system level APIs (different for every platform, i guess) from Rust to determine the preferences and expose those to Rust and finally the webview. Without changing  `navigator.language` or `navigator.languages` this approach is not ideal, because some libraries rely on those fields. Maybe an approach where those preferences are exposed on the `navigator` and additionally have exposed to Rust and Webview via functions.

keiya01:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [n] No

### Checklist

### Other information

Currently this is not working on ARM windows but this is doing on x86 windows.

### DIFF
--- a/.changes/i18n-windows.md
+++ b/.changes/i18n-windows.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix i18n on windows
--- a/.changes/i18n-windows.md
+++ b/.changes/i18n-windows.md
@@ -2,4 +2,4 @@
 "wry": patch
 ---
 
-Fix i18n on windows
+Set webview2 language to match the OS language. This makes i18n functions like `new Date().toLocaleStrin()` behave correctly. 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,8 @@ windows-implement = "0.37.0"
   "Win32_System_SystemInformation",
   "Win32_System_SystemServices",
   "Win32_UI_Shell",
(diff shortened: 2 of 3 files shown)


.changes/i18n-windows.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix i18n on windows
```
amrbashir:
```suggestion
Set webview2 language to match the OS language. This makes i18n functions like `new Date().toLocaleStrin()` behave correctly. 
```

amrbashir (changes requested):

amrbashir (approved):

PR #664, opened 2022-08-17
feat(mobile): configure URLs to be allowed to load with self signed cert

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

On Android, it requires this change on RustWebClient.kt:

```
@SuppressLint("WebViewClientOnReceivedSslError")
    override fun onReceivedSslError(view: WebView?, handler: SslErrorHandler, error: android.net.http.SslError) {
      if (allowSslError(error.url)) {
        handler.proceed()
      } else {
        handler.cancel()
      }
    }

    private external fun allowSslError(url: String): Boolean
```

### DIFF
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -166,7 +166,7 @@ pub struct WebViewAttributes {
   ///
   /// ## Platform-specific
   ///
-  /// - **macOS / Windows / Linux**: Not implemented.
+  /// - **Windows / Linux**: Not implemented.
   pub allowed_self_signed_cert_urls: Vec<Url>,
 }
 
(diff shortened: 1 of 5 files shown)


src/webview/mod.rs
```diff
@@ -386,6 +393,17 @@ impl<'a> WebViewBuilder<'a> {
     self
   }
 
+  /// Configures the given URL to be allowed to load with a self-signed certificate.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **macOS / Windows / Linux**: Not implemented.
```
wusyong: mac should also get the support I assume?

wusyong (approved):

src/webview/mod.rs
lucasfernog: Fixed.

PR #665, opened 2022-08-19
Revert "feat(mobile): configure URLs to be allowed to load with self signed cert"

lucasfernog:
Reverts tauri-apps/wry#664


lucasfernog:
@wusyong i'm reverting the entire PR but the change to use `MyUiViewController` as class name is really needed, the class is never registered otherwise.

wusyong (approved):

PR #671, opened 2022-08-24
feat(android): generate kotlin files at build time

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -23,6 +23,10 @@ fn main() {
       ))
     }
 
+    println!("cargo:rerun-if-env-changed=WRY_ANDROID_REVERSED_DOMAIN");
+    println!("cargo:rerun-if-env-changed=WRY_ANDROID_APP_NAME_SNAKE_CASE");
+    println!("cargo:rerun-if-env-changed=WRY_ANDROID_KOTLIN_FILES_OUT_DIR");
+
     let reversed_domain = env_var("WRY_ANDROID_REVERSED_DOMAIN");
     let app_name_snake_case = env_var("WRY_ANDROID_APP_NAME_SNAKE_CASE");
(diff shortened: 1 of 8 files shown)


src/webview/android/kotlin/TauriActivity.kt
```diff
@@ -0,0 +1,71 @@
+package {{app-domain-reversed}}.{{app-name-snake-case}}
```
lucasfernog, line 1: should this one be generated by tao?

amrbashir: iirc, tao generates functions using `android_binding!` macro, which takes the app_domain and app_name as input.

lucasfernog: Yeah but the TauriActivity stuff is still related to the tao code, so in this setup if tao changes, we'd need to change wry too.

build.rs
```diff
@@ -9,4 +9,37 @@ fn main() {
...
     println!("cargo:rustc-link-lib=framework=WebKit");
   }
+
+  let is_android = std::env::var("TARGET")
+    .map(|t| t.ends_with("-android"))
+    .unwrap_or_default();
+  if is_android {
+    use std::{fs, path::PathBuf};
+
+    fn env_var(var: &str) -> String {
```
lucasfernog, line 19: we need to add a `println!("cargo:rerun-if-env-changed={var}");` here

lucasfernog (approved):

PR #674, opened 2022-08-25
fix(android): inject initialization scripts in HTML

lucasfernog:
The Android WebView does not have a reliable way to inject the initialization scripts. Using `onPageStarted` has some issues when running a production app, the script is executed too late. Until we can find a better solution, we'll inject the scripts in the HTML via the custom protocol.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -89,6 +89,10 @@ pub struct WebViewAttributes {
   /// Initialize javascript code when loading new pages. When webview load a new page, this
   /// initialization code will be executed. It is guaranteed that code is executed before
   /// `window.onload`.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Android:** The initialization scripts are only executed on custom protocol URLs.
   pub initialization_scripts: Vec<String>,
   /// Register custom file loading protocols with pairs of scheme uri string and a handling
   /// closure.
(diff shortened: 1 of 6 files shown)


src/webview/mod.rs
```diff
@@ -89,6 +89,10 @@ pub struct WebViewAttributes {
   /// Initialize javascript code when loading new pages. When webview load a new page, this
   /// initialization code will be executed. It is guaranteed that code is executed before
   /// `window.onload`.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Android:** The initialization scripts are only executed on custom protocol URLs.
```
wusyong: Okay, I've searched react native and ionic. They both seem to set script in html too.
I think we should document the behavior here too. So users can expect how is it done.

src/webview/mod.rs
```diff
@@ -224,6 +228,10 @@ impl<'a> WebViewBuilder<'a> {
...
```
wusyong: ditto

wusyong (approved):

PR #682, opened 2022-09-03
feat: allow setting webview bg color, closes #197

### LINKED ISSUE #197: Allow setting the browser's background color during attributes
Allow the user to set a background color for the webview window to render while loading content. This is typically the step between the os window initializing the webview, and the webview initializing the application.
https://github.com/tauri-apps/tauri/issues/1564 ([my comment](https://github.com/tauri-apps/tauri/issues/1564#issuecomment-825871338))

Having a `background_color` field on `wry::Attributes` (I think that is the proper place for it).
I don't know which format we should accept the color in...
* hex string?
* rgb struct?
* rgb tuple?
* other?

none yet

- [ ] Yes
- [x] No (I am not familiar with the macOS code and don't have access to an macOS machine)

**Additional context**
The linux part seems easy enough once a format for accepting the color is decided on.  Some pseudo-code to replace:
https://github.com/tauri-apps/wry/blob/765fe5ae413a1c13ad99802b49f3af859a2445d5/src/webview/linux/mod.rs#L138-L146
with
```rust
let alpha = (!transparent) as i8 as f64;
webview.set_background_color(&RGBA {
  red: bg_color.red,
  green: bg_color.green,
  blue: bg_color.blue,
  alpha,
});
```

assuming `bg_color` is some sort of struct with red, blue, green fields of `f64`.

I assume webview2 and macOS has a way of setting the background color but I'm not familiar with them yet

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

## TODOs:
- [x] android
- [ ] macOS

### DIFF
--- a/src/webview/android/main_pipe.rs
+++ b/src/webview/android/main_pipe.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
+use crate::webview::RGBA;
 use crossbeam_channel::*;
 use once_cell::sync::Lazy;
 use std::os::unix::prelude::*;
@@ -40,12 +41,13 @@ impl MainPipe<'_> {
     let activity = self.activity.as_obj();
     if let Ok(message) = CHANNEL.1.recv() {
       match message {
-        WebViewMessage::CreateWebView {
-          url,
-          devtools,
-          background_color,
-          transparent,
-        } => {
+        WebViewMessage::CreateWebView(attrs) => {
+          let CreateWebViewAttributes {
+            url,
+            devtools,
+            transparent,
+            background_color,
+          } = attrs;
           // Create webview
           let rust_webview_class = find_my_class(
             env,
@@ -157,7 +159,7 @@ fn find_my_class<'a>(
 fn set_background_color<'a>(
   env: JNIEnv<'a>,
   webview: JObject<'a>,
-  background_color: (u8, u8, u8, u8),
+  background_color: RGBA,
 ) -> Result<(), JniError> {
   let color_class = env.find_class("android/graphics/Color")?;
   let color = env.call_static_method(
@@ -177,12 +179,15 @@ fn set_background_color<'a>(
 
 #[derive(Debug)]
 pub enum WebViewMessage {
-  CreateWebView {
-    url: String,
-    devtools: bool,
-    background_color: Option<(u8, u8, u8, u8)>,
-    transparent: bool,
-  },
+  CreateWebView(CreateWebViewAttributes),
   Eval(String),
-  SetBackgroundColor((u8, u8, u8, u8)),
+  SetBackgroundColor(RGBA),
+}
+
+#[derive(Debug)]
+pub struct CreateWebViewAttributes {
+  pub url: String,
+  pub devtools: bool,
+  pub transparent: bool,
+  pub background_color: Option<RGBA>,
 }
--- a/src/webview/android/main_pipe.rs
+++ b/src/webview/android/main_pipe.rs
@@ -40,7 +40,12 @@ impl MainPipe<'_> {
     let activity = self.activity.as_obj();
     if let Ok(message) = CHANNEL.1.recv() {
       match message {
-        WebViewMessage::CreateWebView(url, devtools) => {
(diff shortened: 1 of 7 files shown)


src/webview/mod.rs
```diff
@@ -61,8 +61,24 @@ pub struct WebViewAttributes {
...
+  /// ## Platform-specific:
+  ///
+  /// **Android**: Not implemented.
+  /// **Windows 7**: Not supported.
   pub transparent: bool,
+  /// Specify the webview background color. This will be ignored if `transparent` is set to `true`.
+  ///
+  /// The color uses the RGBA format.
+  ///
+  /// ## Platfrom-specific:
```
lucasfernog:
```suggestion
  /// ## Platform-specific:
```

src/webview/mod.rs
```diff
@@ -61,8 +61,24 @@ pub struct WebViewAttributes {
...
+  ///
+  /// - **Android**: Not implemented..
```
lucasfernog:
```suggestion
  /// - **Android**: Not implemented.
```

src/webview/mod.rs
```diff
@@ -61,8 +61,24 @@ pub struct WebViewAttributes {
...
+  /// - **Windows**:
+  ///   - on Windows 7, transparency is not supported and the alpha value will be ignored.
```
lucasfernog:
```suggestion
  ///   - On Windows 7, transparency is not supported and the alpha value will be ignored.
```

src/webview/mod.rs
```diff
@@ -61,8 +61,24 @@ pub struct WebViewAttributes {
...
+  ///   - on Windows higher than 7: translucent colors are not supported so any alpha value other than `0` will be replaced by `255`
```
lucasfernog:
```suggestion
  ///   - On Windows higher than 7: translucent colors are not supported so any alpha value other than `0` will be replaced by `255`
```

src/webview/mod.rs
```diff
@@ -535,6 +572,20 @@ impl WebView {
...
+
+  /// Specify the webview background color.
+  ///
+  /// The color uses the RGBA format.
+  ///
+  /// ## Platfrom-specific:
+  ///
+  /// - **Android**: Not implemented.
+  /// - **Windows**:
+  ///   - on Windows 7, transparency is not supported and the alpha value will be ignored.
```
lucasfernog:
```suggestion
  ///   - On Windows 7, transparency is not supported and the alpha value will be ignored.
```

src/webview/mod.rs
```diff
@@ -535,6 +572,20 @@ impl WebView {
...
+  ///   - on Windows higher than 7: translucent colors are not supported so any alpha value other than `0` will be replaced by `255`
```
lucasfernog:
```suggestion
  ///   - On Windows higher than 7: translucent colors are not supported so any alpha value other than `0` will be replaced by `255`
```

lucasfernog:
Works on Linux and Windows. Does not work on macOS though :(

EDIT: Ahh I just saw you removed it :P

amrbashir:
yeah I removed the macOS implementation since I can't test it and from my research it needs a lot more work

src/webview/android/main_pipe.rs
```diff
@@ -136,8 +154,35 @@ fn find_my_class<'a>(
...
+  env.call_method(webview, "setBackgroundColor", "(I)V", &[color])?;
+  Ok(())
+}
+
 #[derive(Debug)]
 pub enum WebViewMessage {
-  CreateWebView(String, bool),
+  CreateWebView {
+    url: String,
+    devtools: bool,
```
lucasfernog: we should take the entire webviewattributes struct here

amrbashir: I don't think we can, there is bunch of handlers in WebViewAttributes that aren't send+sync

lucasfernog: well just create a dedicated struct then, otherwise this enum value will soon have 10 fields

PR #683, opened 2022-09-04
fix(android): delay WebView#loadUrl call

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

In older webview versions, the `WebViewClient::shouldInterceptRequest` method is not invoked for the initial URL load. This is fixed if we delay the loadUrl() call by posting a message to the main thread. This was reported in Discord by a user testing on an emulator with API 28.

### DIFF
--- a/src/webview/android/kotlin/RustWebView.kt
+++ b/src/webview/android/kotlin/RustWebView.kt
@@ -6,11 +6,19 @@ package {{app-domain-reversed}}.{{app-name-snake-case}}
 
 import android.webkit.*
 import android.content.Context
+import android.os.Handler
+import android.os.Looper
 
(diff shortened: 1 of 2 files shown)


lucasfernog:
We could also add an option for a custom error page validating the webview version like what Capacitor does: https://github.com/ionic-team/capacitor/pull/5723 (though emitting an event is more like our way of doing things)

amrbashir (approved):

PR #689, opened 2022-09-06
feat(android): JNI execution handle

jhutchins:

- [ ] Bugfix
- [X] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist

### Other information
Add the ability to execute JNI commands on the webview object.

### DIFF
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -15,6 +15,8 @@ pub mod prelude {
   pub use super::android::{binding::*, setup};
 }
 #[cfg(target_os = "android")]
+pub use android::JniHandle;
+#[cfg(target_os = "android")]
 use android::*;
 #[cfg(any(
   target_os = "linux",
@@ -659,6 +661,19 @@ impl WebviewExtMacOS for WebView {
 
   fn ns_window(&self) -> cocoa::base::id {
     self.webview.ns_window.clone()
+  }
+}
+
+#[cfg(target_os = "android")]
+/// Additional methods on `WebView` that are specific to Android
+pub trait WebviewExtAndroid {
+  fn handle(&self) -> JniHandle;
+}
+
+#[cfg(target_os = "android")]
+impl WebviewExtAndroid for WebView {
+  fn handle(&self) -> JniHandle {
+    JniHandle
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/webview/mod.rs
```diff
@@ -662,6 +664,19 @@ impl WebviewExtMacOS for WebView {
   }
 }
 
+#[cfg(target_os = "android")]
+/// Additional methods on `WebView` that are specific to Android
+pub trait WebviewExtAndroid {
+  fn handle(&self) -> JniHandle;
+}
```
amrbashir, line 671: I think this should be something like to match other platforms and allow a bit more flexibility.

```suggestion
pub trait WebviewExtAndroid {
  fn jni_env(&self) -> &JNIEnv; // or *mut JNIEnv
  fn activity(&self) -> &GlobalRef;
  fn webview(&self) -> &GlobalRef;
}
```

amrbashir (changes requested):

src/webview/mod.rs
jhutchins: I can make that change, but we still need some method to execute code in the context of the looper used in the android setup. The issue being that methods on the webview can only be called from that thread associated with the looper. Calls from any other thread just result in errors.

I'm not sure, but it's possible we could expose the ForeignLooper through this mechanism as well and allow addition callbacks to be attached and that might work to run the code from the right thread. I haven't thought all the way through it yet.

amrbashir: we could also expose a ptr to the looper where they can use https://docs.rs/ndk/latest/ndk/looper/struct.ForeignLooper.html#method.from_ptr but your approach is probably safer.

amrbashir (dismissed):

src/webview/mod.rs
jhutchins: Also, there's the actual difficulty of simply getting those references. They're owned by the MainPipe and there is not a ready reference to that object anywhere.

lucasfernog (dismissed):

amrbashir (approved):

PR #695, opened 2022-09-16
fix(windows): Focus webview when window starts moving or resizing.

### LINKED ISSUE #1911: <select> opens outside window on first try
`<select>` elements open outside the application window. When opened a second time, they open in the correct position. The problem occurs again when moving the window to a new location.

OS: Windows, version 10.0.19043 X64
Node: 15.14.0
Rustc: 1.52.1

### LINKED ISSUE #5135: [bug] The drop-down menu of "select" item does not change the position as the application window moves
### Describe the bug

I run the 'Navigation' example and click “select” item to bring up the drop-down menu：
图片

however, after I moved the application window, the drop-down menu remained in its original position:
图片

### Reproduction

1、run the 'Navigation' example and click “select” item to bring up the drop-down menu：
图片

2、after I moved the application window, the drop-down menu remained in its original position:
图片

### Expected behavior

_No response_

### Platform and versions

Environment
  › OS: Windows 10.0.22000 X64
  › Webview2: 105.0.1343.25
  › MSVC:
      - Visual Studio Community 2022
  › Node.js: 16.14.0
  › npm: 8.3.1
  › pnpm: Not installed!
  › yarn: Not installed!
  › rustup: 1.24.3
  › rustc: 1.61.0
  › cargo: 1.61.0
  › Rust toolchain: stable-x86_64-pc-windows-msvc

Packages
  › @tauri-apps/cli [NPM]: 1.0.3
  › @tauri-apps/api [NPM]: 1.0.1 (outdated, latest: 1.0.2)
  › tauri [RUST]: 1.0.2,
  › tauri-build [RUST]: 1.0.2,
  › tao [RUST]: 0.12.1,
  › wry [RUST]: 0.19.0,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › devPath: ../dist

App directory structure
  ├─ dist
  ├─ node_modules
  └─ src-tauri

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
Fixes https://github.com/tauri-apps/tauri/issues/1911#issuecomment-1078252299, https://github.com/tauri-apps/tauri/issues/5135

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/fix-webview-focus.md
+++ b/.changes/fix-webview-focus.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Focus webview when window starts moving or resizing on Windows to automatically close `<select>` dropdowns.
--- a/.changes/fix-webview-focus.md
+++ b/.changes/fix-webview-focus.md
@@ -2,4 +2,4 @@
 "wry": patch
 ---
 
-Focus webview when window starts moving or resizing on Windows to automatically close `<select>` dropdowns.
+Focus webview when window starts moving or resizing on Windows to automatically close `<select>` dropdowns. Also notify webview2 whenever the window position/size changes which fixes the `<select>` dropdown position
--- a/src/webview/webview2/mod.rs
+++ b/src/webview/webview2/mod.rs
@@ -192,13 +192,13 @@ impl InnerWebView {
     // background color
     if !attributes.transparent {
       if let Some(background_color) = attributes.background_color {
-        set_background_color(&controller, background_color)?;
+        set_background_color(controller, background_color)?;
       }
     }
 
     // Transparent
     if attributes.transparent && !is_windows_7() {
-      set_background_color(&controller, (0, 0, 0, 0))?;
+      set_background_color(controller, (0, 0, 0, 0))?;
     }
 
     // The EventRegistrationToken is an out-param from all of the event registration calls. We're
@@ -616,12 +616,12 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
           });
         }
 
-        win32wm::WM_ENTERSIZEMOVE | win32wm::WM_SETFOCUS => {
+        win32wm::WM_SETFOCUS | win32wm::WM_ENTERSIZEMOVE => {
           let controller = dwrefdata as *mut ICoreWebView2Controller;
           let _ = (*controller).MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
         }
 
-        win32wm::WM_EXITSIZEMOVE => {
+        win32wm::WM_WINDOWPOSCHANGED => {
           let controller = dwrefdata as *mut ICoreWebView2Controller;
           let _ = (*controller).NotifyParentWindowPositionChanged();
         }
--- a/src/webview/webview2/mod.rs
+++ b/src/webview/webview2/mod.rs
@@ -616,7 +616,7 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
           });
         }
 
-        win32wm::WM_SETFOCUS => {
+        win32wm::WM_ENTERSIZEMOVE | win32wm::WM_SETFOCUS => {
           let controller = dwrefdata as *mut ICoreWebView2Controller;
           let _ = (*controller).MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
         }
--- a/src/webview/webview2/mod.rs
+++ b/src/webview/webview2/mod.rs
@@ -621,6 +621,11 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
           let _ = (*controller).MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
         }
 
+        win32wm::WM_MOVE | win32wm::WM_MOVING => {
+          let controller = dwrefdata as *mut ICoreWebView2Controller;
+          let _ = (*controller).NotifyParentWindowPositionChanged();
+        }
+
         win32wm::WM_DESTROY => {
           Box::from_raw(dwrefdata as *mut ICoreWebView2Controller);
         }
--- a/src/webview/webview2/mod.rs
+++ b/src/webview/webview2/mod.rs
@@ -621,7 +621,7 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
           let _ = (*controller).MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
         }
 
-        win32wm::WM_EXITSIZEMOVE=> {
+        win32wm::WM_EXITSIZEMOVE => {
           let controller = dwrefdata as *mut ICoreWebView2Controller;
           let _ = (*controller).NotifyParentWindowPositionChanged();
         }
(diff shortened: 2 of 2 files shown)


FabianLars:
btw this is basically how the chrome browser behaves, with the difference that chrome actually blurs the `select` element while it will still look focused (minus the open dropdown) in wv2.

src/webview/webview2/mod.rs
```diff
@@ -616,7 +616,7 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
           });
         }
 
-        win32wm::WM_SETFOCUS => {
+        win32wm::WM_ENTERSIZEMOVE | win32wm::WM_SETFOCUS => {
```
amrbashir: This will focus only once when the resize/move starts, is it better to just use `WM_EXITSIZEMOVE`?

amrbashir (changes requested):

src/webview/webview2/mod.rs
FabianLars: No it needs to run when it starts moving/resizing, otherwise the window will move but the drop-down menu will stay in it's current position.

amrbashir: I just tried both and it doesn't matter both failed. When I start moving the window, the webview is focused and it will close the select dropdown but when I click on it, the drop down appears in the previous location. What version of webview2 are you on?

FabianLars: 105, but for me it only opened in the old location once really randomly.
And to be clear, this PR is to close the drop-down automatically. After that it's up to the webview to move the actual drop-down position (which MS mostly fixed) - so it's more about the second issue i linked

FabianLars: Oh btw i only tested this in a tauri app, not in a wry example

FabianLars: ah damn i can repro it. When the dropdown is already closed before moving it opens at the wrong location. if it was open and auto-closed it works as expected.

amrbashir: I am on 105 as well and the only difference this PR made is that it closes the dropdown but the location bug still exists

amrbashir: I still see the bug even when it is open and gets auto-closed

FabianLars: guess i tested it in a really weird way then idk. It also seems to fix its position on hover change which is kinda funny and probably what tricked me here idk.

amrbashir: when it comes to Windows and Microsoft, everything is funny

FabianLars:
shit i was a little too trigger happy here. i think i got it

src/webview/webview2/mod.rs
```diff
@@ -616,11 +616,16 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
...
           let controller = dwrefdata as *mut ICoreWebView2Controller;
           let _ = (*controller).MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
         }
 
+        win32wm::WM_MOVE | win32wm::WM_MOVING => {
```
amrbashir: this works and actually we can optimize this and only call it when moving/resizing is finished
```suggestion
        win32wm::WM_EXITSIZEMOVE=> {
```

.changes/fix-webview-focus.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": patch
+---
+
+Focus webview when window starts moving or resizing on Windows to automatically close `<select>` dropdowns.
```
amrbashir:
```suggestion
Focus webview when window starts moving or resizing on Windows to automatically close `<select>` dropdowns. Also notify webview2 whenever the window position/size changes which fixes the `<select>` dropdown position
```

amrbashir (dismissed):

src/webview/webview2/mod.rs
```diff
@@ -616,11 +616,16 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
           });
         }
 
-        win32wm::WM_SETFOCUS => {
+        win32wm::WM_ENTERSIZEMOVE | win32wm::WM_SETFOCUS => {
           let controller = dwrefdata as *mut ICoreWebView2Controller;
           let _ = (*controller).MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC);
         }
 
+        win32wm::WM_EXITSIZEMOVE=> {
```
```suggestion
        win32wm::WM_EXITSIZEMOVE => {
```

amrbashir (dismissed):

src/webview/webview2/mod.rs
FabianLars: yeah i was a little scared from their docs https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2controller.notifyparentwindowpositionchanged?view=webview2-dotnet-1.0.1343.22
> Tells WebView that the main WebView parent (or any ancestor) HWND moved.
> This is needed for accessibility and certain dialogs in WebView to work correctly.

FabianLars: Found different docs where they show move and moving too: https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2host?view=webview2-0.9.430#notifyparentwindowpositionchanged

I mean, in this specific use case exitsizemove works perfect, but i wonder what else its used for. Imagine of OS dpcs would actually document stuff x)

amrbashir: hmm actually I think it is better to use `WM_MOVE` and `WM_SIZE` so it would account for maximized and fullscreen changes

FabianLars: My local copy still has the move + moving and resizing + maximizing seems to be handled by that already.

amrbashir: `WM_MOVE` is fine but `WM_MOVING` can be triggered multiple times and so we don't need to keep calling it every time we just need to call it once when moving is done.

amrbashir: `WM_WINDOWPOSCHANGED` also works

amrbashir: man win32 has lots of messages for the same damn thing

amrbashir (approved):

src/webview/webview2/mod.rs
FabianLars: > My local copy still has the move + moving and resizing + maximizing seems to be handled by that already.

Same for fullscreen i think. it's a bit tricky to test.

> WM_MOVE is fine but WM_MOVING can be triggered multiple times

Aren't they basically the same event? As-in they both get triggered continuously while the window is dragged?

> so we don't need to keep calling it every time we just need to call it once when moving is done.

Yeah well, that depends on what Microsoft's intention behind the notify api is x)

> WM_WINDOWPOSCHANGED also works

the docs of that sounded like it's only triggered by programmatic position changes to me x)

Anyway, i'm fine with sticking to WM_EXITSIZEMOVE (edit: Or WM_WINDOWPOSCHANGED) until we see something that needs some spammier events like the docs suggests. I can't think of anything 🤷

amrbashir: > Aren't they basically the same event? As-in they both get triggered continuously while the window is dragged?

I think `WM_MOVE` only triggers at the end but I haven't tested it though but should be according to the docs.

> the docs of that sounded like it's only triggered by programmatic position changes to me x)

yeah the docs is misleading because I tested it and it seems to work, maybe test this one more time before I merge?

amrbashir: @FabianLars could you test with the latest commit using `WM_WINDPOSCHANGED` before I merge?

FabianLars: works fine on my end.

amrbashir: cool

PR #708, opened 2022-09-29
Use Arc for Window in WebView

FredrikNoren:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This changes the Window held by WebView from `Rc<Window>` to `Arc<Window>`, so that the window can be
accessed from other threads. This also changes the `WebView.window` method to return `&Arc<Window>` instead
of `&Window`, which I think should be backwards compatible.

### DIFF
--- a/.changes/window-arc.md
+++ b/.changes/window-arc.md
@@ -1,0 +1,7 @@
+---
+"wry": "minor"
+---
+
+This changes the Window held by WebView from `Rc<Window>` to `Arc<Window>`, so that the window can be
+accessed from other threads. This also changes the `WebView.window` method to return `&Arc<Window>` instead
+of `&Window`, which I think should be backwards compatible.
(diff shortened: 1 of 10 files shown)


amrbashir:
I think `Window` on Linux is not Sync because it has some `Rc` inside it and also the getter methods are accessing the gtk window directly and will probably panic if it is accessed through another thread where gtk has not been initialized.

FredrikNoren:
@amrbashir Hm I think winit's Window is Sync Send always: https://docs.rs/winit/latest/winit/window/struct.Window.html (or at least it seems so from the docs), but maybe tao was forked after they fixed that.

amrbashir:
our Linux backend is different from winit's backend. Even though the type itself is Send+Sync I think using it on a different thread when calling a getter like `is_resizable` will cause a panic.

FredrikNoren:
Ah ok I see :/ Closing for now then

amrbashir:
I haven't tested if it will actually panic but nonetheless this is something that we will have to account for when migrating back to winit (i.e ensuring it is sync/send).

lucasfernog:
GTK does panic if you call a window API outside of the main thread.

Issue #666, opened 2022-08-21
[feat] access to `URISchemeRequest` body on Linux


JonasKruckenberg:
**Is your feature request related to a problem? Please describe.**

Both the Rust type [URISchemeRequest](https://docs.rs/webkit2gtk/0.18.0/webkit2gtk/struct.URISchemeRequest.html) and upstream [WebKitURISchemeRequest](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitURISchemeRequest.html) do not expose a way to access the requests body. For a variety of usecases including Tauri's IPC system it would be great if we could actually access the main content of the request.

**Describe the solution you'd like**
Implement this functionality in upstream projects mirroring the one from WKWebView [URLRequest](https://developer.apple.com/documentation/foundation/urlrequest) namely `httpBody` and `httpBodyStream` so we can consume the body both blocking and through a Rust `Stream`.

**Describe alternatives you've considered**
Not have access to the body on Linux and marking this as a platform difference, but sadly this is kind of a really important core feature. It being absent makes the custom protocol pretty useless.

**Would you assign yourself to implement this feature?**
- [x] Yes
- [ ] No

**Additional context**

wusyong:
Do we have an upstream issue to track it?

amrbashir:
I don't think we opened one for it yet.

TingPing:
This is tricky to implement in WebKitGTK.

There is no "request body" sitting around, it is a `GInputStream` that is read on-demand as network writes happen. In order to access a body a complete copy would have to be stored in-memory, which isn't reasonable in some cases.

JonasKruckenberg:
Let's expose that stream though, we actually care much more about that stream than a full body. Wry itself already has code to read a stream into a `Vec<u8>` on macOS too

TingPing:
"Just expose a stream" still has complications. It is possible for a stream to be read-once. In which case you have to copy the entire thing into memory. It is also possible for a stream to be `GSeekable` and read multiple times but you can't seek a stream that is being actively read somewhere else.

TingPing:
I am mostly thinking about this from a libsoup perspective, the http library webkitgtk uses, its possible things work out nicer inside of WebKit.

wusyong:
Upstream issue created: https://bugs.webkit.org/show_bug.cgi?id=252564

ModProg:
seems like the upstream issue was fixed: https://github.com/WebKit/WebKit/pull/10714

[...]

PR #721, opened 2022-10-14
feat(linux): add headers to URL scheme request

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/webview/webkitgtk/web_context.rs
+++ b/src/webview/webkitgtk/web_context.rs
@@ -20,9 +20,10 @@ use std::{
 };
 use url::Url;
 use webkit2gtk::{
-  traits::*, ApplicationInfo, CookiePersistentStorage, LoadEvent, UserContentManager, WebContext,
-  WebContextBuilder, WebView, WebsiteDataManagerBuilder,
+  traits::*, ApplicationInfo, CookiePersistentStorage, LoadEvent, URISchemeResponse,
+  UserContentManager, WebContext, WebContextBuilder, WebView, WebsiteDataManagerBuilder,
 };
+use webkit2gtk_sys::webkit_get_minor_version;
 
 #[derive(Debug)]
 pub struct WebContextImpl {
@@ -32,6 +33,7 @@ pub struct WebContextImpl {
   registered_protocols: HashSet<String>,
   automation: bool,
   app_info: Option<ApplicationInfo>,
+  is_2_36: bool,
 }
 
 impl WebContextImpl {
@@ -78,6 +80,8 @@ impl WebContextImpl {
         .expect("invalid wry version patch"),
     );
 
+    let is_above_2_36 = unsafe { webkit_get_minor_version() >= 36 };
+
     Self {
       context,
       automation,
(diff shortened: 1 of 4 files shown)


wusyong:
There's a race condition could cause page freeze right now.
Run `custom_protocol` example and hit the button quickly to constantly changing pages. You will get stuck at some point.
Sometimes it will even be blank page on start.
It will also stuck even if I just call `http_headers` and do nothing.

wusyong:
Okay turns out mutex will only reduce the chance. It can still stuck!
Guess I have to make a C example to reproduce it to upstream.

wusyong:
I finally found the cause. So  it seems the binding from `webkit2gtk-rs` will take header's ownership:
https://github.com/tauri-apps/webkit2gtk-rs/blob/crate/src/auto/uri_scheme_request.rs#L113
And the type in header field is a `unique_ptr`!
https://searchfox.org/wubkat/source/Source/WebKit/UIProcess/API/glib/WebKitURISchemeRequest.cpp#73
This cause a race condition that can result into use after free...

src/webview/webkitgtk/web_context.rs
```diff
@@ -221,14 +223,21 @@ where
     if let Some(uri) = request.uri() {
       let uri = uri.as_str();
 
-      //let headers = unsafe {
-      //  webkit_uri_request_get_http_headers(request.clone().to_glib_none().0)
-      //};
-
       // FIXME: Read the method
```
amrbashir: we should be able to get the method now too, no? I think it was added in 2.36 too

wusyong: Yes, I'll add those too.

wusyong:
@lucasfernog These methods require webkit2gtk 2.36. I think I probably need to add a feature flag for it. How do you think tauri should do? Also a flag to use it?

lucasfernog:
@wusyong can we detect the webkit2gtk version at runtime and not call the API if the version is below 2.36? I think that would be the safest approach.

wusyong:
Yes we can. I'll try this approach then.

lucasfernog (approved):

chairmanwow:
https://github.com/tauri-apps/wry/blob/16a7086ed7e38aa8c57bd14f067b0cc955b478d3/src/webview/webkitgtk/web_context.rs#L255

It looks like this doesn't support POST with bodies, say for example JSON in the POST request body. Is there a reason why the body is being set to the empty vector? Does WebKitURISchemeRequest have no way of getting the body?

I tried this in an app sending POST's with JSON and the custom scheme handler can't see those. I'm happy to dig into this and try to get to a pull request if there's interest from the tauri side.

wusyong:
@chairmanwow [WebKitURISchemeRequest](https://searchfox.org/wubkat/source/Source/WebKit/UIProcess/API/glib/WebKitURISchemeRequest.cpp) doesn't expose it yet. That's why there's still issue #666 for it. If you are interested in this, I suppose you can open a PR in Webkit directly to add it.

chairmanwow:
Yikes, understood. Thanks for the effort and for clarifying the why for me.

PR #731, opened 2022-10-22
Fix cross compilation.

dvc94ch:
this is one of the patches that let's dioxus apps be built for ios with xbuild.

### DIFF
--- a/.changes/ios-crosscompile.md
+++ b/.changes/ios-crosscompile.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Support cross compiling ios on a non macos host.
--- a/build.rs
+++ b/build.rs
@@ -3,10 +3,8 @@
 // SPDX-License-Identifier: MIT
 
 fn main() {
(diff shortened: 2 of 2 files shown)


wusyong (commented):
LGTM. Could you add a .change file?

dvc94ch:
Like this? I'm not sure what a .change file is. Added an entry to the change log. If this is ok I'll do the other PR too.

dvc94ch:
Ah ok think I figured it out

.changes/crosscompile.md
```diff
@@ -0,0 +1,8 @@
+```md
+---
+"wry": patch
+---
+
+Support cross compiling ios on a non macos host.
+
+```
```
amrbashir:
```suggestion
---
"wry": patch
---

Support cross compiling ios on a non macos host.

```

dvc94ch: :facepalm:

wusyong (approved):

PR #732, opened 2022-10-24
feat: add `.url()`to access the current url

JonasKruckenberg:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/src/webview/android/mod.rs
+++ b/src/webview/android/mod.rs
@@ -10,6 +10,7 @@ use http::{
   header::{HeaderValue, CONTENT_SECURITY_POLICY, CONTENT_TYPE},
   Request, Response,
 };
+use url::Url;
 use kuchiki::NodeRef;
 use once_cell::sync::OnceCell;
 use sha2::{Digest, Sha256};
@@ -227,6 +228,10 @@ impl InnerWebView {
   }
 
   pub fn print(&self) {}
+
+  pub fn url(&self) -> Url {
+    todo!();
+  }
 
   pub fn eval(&self, js: &str) -> Result<()> {
     MainPipe::send(WebViewMessage::Eval(js.into()));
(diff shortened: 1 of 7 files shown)


src/webview/webview2/mod.rs
```diff
@@ -767,6 +768,17 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
     let _ = self.eval("window.print()");
   }
 
+  pub fn url(&self) -> Url {
+    let mut pwstr = PWSTR::null();
+
+    let webview = unsafe { self.controller.CoreWebView2().unwrap() };
+    unsafe { webview.Source(&mut pwstr).unwrap() };
```
```suggestion
    unsafe { self.webview.Source(&mut pwstr).unwrap() };
```

src/webview/android/mod.rs
```diff
@@ -228,6 +229,10 @@ impl InnerWebView {
 
   pub fn print(&self) {}
 
+  pub fn url(&self) -> Url {
+    todo!();
```
amrbashir: this will panic, could it maybe replaced with a default value? or change the function signature to use `Result`

amrbashir (changes requested):

src/webview/android/mod.rs
JonasKruckenberg: My assumption was that we actually want this function on android too, I just don't know how to implement it that's why I marked it as `todo`.

Or is this not possible on android?

amrbashir: No it is possible, let me add it

amrbashir (approved):

amrbashir (changes requested):
I am once again asking for change file.

amrbashir (approved):

PR #742, opened 2022-10-28
fix(macos): do not trigger unsupported key feedback sound on keypress

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

Ref https://github.com/tauri-apps/tauri/issues/2626

### DIFF
--- a/.changes/macos-key-feedback.md
+++ b/.changes/macos-key-feedback.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Keypress on non-input element no longer triggers unsupported key feedback sound.
--- a/examples/custom_protocol.rs
+++ b/examples/custom_protocol.rs
@@ -31,7 +31,7 @@ fn main() -> wry::Result<()> {
       // Read the file content from file path
       let content = read(canonicalize(PathBuf::from("examples").join(
         if path == &"/" {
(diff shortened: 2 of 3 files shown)


amrbashir (approved):

wusyong:
I had a discussion with Lucas and the reason I didn't merge this is because I got a warning:
`TSM AdjustCapsLockLEDForKeyTransitionHandling - _ISSetPhysicalKeyboardCapsLockLED Inhibit`
This seems to be CJK IME only. The answer on SO isn't pretty helpful. They said just don't use Caplocks for input switching.

lucasfernog:
Maybe it's because we're not passing the event to the system input manager with `interpretKeyEvents`. The winit approach is much more complicated, see [the code](https://github.com/tauri-apps/tao/blob/5e85dbef325fd8050b5cc26079627a3b645285c7/src/platform_impl/macos/view.rs#L683).

wusyong:
I didn't see any significant error and the input seems intact. I'll pass this until we see any issue.

PR #745, opened 2022-11-06
On macOS, add WKWebview as subview of existing NSView directly

wusyong:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/webview/wkwebview/mod.rs
+++ b/src/webview/wkwebview/mod.rs
@@ -25,7 +25,7 @@ use std::{
   slice, str,
 };
 
-use core_graphics::geometry::{CGPoint, CGRect, CGSize};
+use core_graphics::geometry::CGRect;
 use objc::{
   declare::ClassDecl,
   runtime::{Class, Object, Sel, BOOL},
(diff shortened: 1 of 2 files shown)


src/webview/wkwebview/mod.rs
```diff
@@ -663,38 +663,13 @@ r#"Object.defineProperty(window, 'ipc', {
...
-        let parent_view_cls = match ClassDecl::new("WryWebViewParent", class!(NSView)) {
-          Some(mut decl) => {
-            decl.add_method(
-              sel!(keyDown:),
-              key_down as extern "C" fn(&mut Object, Sel, id),
-            );
-
-            extern "C" fn key_down(_this: &mut Object, _sel: Sel, _event: id) {
-              // do nothing; this prevents macOS from playing an unsupported key feedback sound
-            }
```
amrbashir, line 675: does this bug happen with winit and this PR?

src/webview/wkwebview/mod.rs
```diff
@@ -663,38 +663,13 @@ r#"Object.defineProperty(window, 'ipc', {
...
-          }
-          None => class!(NSView),
-        };
-
-        // Create a view to contain the webview, without it devtools will try to
-        // inject a subview into the frame causing an obnoxious warning.
-        // See https://github.com/tauri-apps/wry/issues/273
-        let parent_view: id = msg_send![parent_view_cls, alloc];
-        let _: () = msg_send![parent_view, init];
-        parent_view.setAutoresizingMask_(NSViewHeightSizable | NSViewWidthSizable);
```
amrbashir, line 687: will the removal of this disable auto resizing of the webview based on the parent? the current behavior is that the webview content will automatically fill the whole window when the window is resized.

src/webview/wkwebview/mod.rs
wusyong: No, TaoView/WinitView also delcared this method already.

src/webview/wkwebview/mod.rs
wusyong: Yes, it will be auto-resized when the window is resized.

amrbashir (approved):

PR #769, opened 2022-11-24
On macOS, fix arrow keys misprint texts

### LINKED ISSUE #5685: [bug] invalid characters appearing on the screen when pressing arrow right
### Describe the bug

Pressing arrow right in an autofocused `<textarea>` will result in adding invalid characters to its content (``)
This happens only initially. If I focus on a different element or window and come back, the `<textarea>` element works as intended. 

### Reproduction

1. on Mac (Ventura) clone the project from this repo: https://github.com/paprikka/tauri-keypress-bug
2. `npm run tauri dev`
3. start typing and press _right_
4. If the caret is at the end of the typed text `` gets appended to it (left, down, up don't cause this issue)
5. change focus to a different window or DOM element
6. change focus back to the texarea
7. type and press _right_
8. now the behaviour is correct

### Expected behavior

Pressing right arrow on the keyboard, when the caret is at the end of the typed text should result in no action.

### Platform and versions

Environment
  › OS: Mac OS 13.0.0 X64
  › Node.js: 18.9.0
  › npm: 8.19.1
  › pnpm: Not installed!
  › yarn: Not installed!
  › rustup: 1.25.1
  › rustc: 1.65.0
  › cargo: 1.65.0
  › Rust toolchain: stable-aarch64-apple-darwin

Packages
  › @tauri-apps/cli [NPM]: 1.2.0
  › @tauri-apps/api [NPM]: Not installed!
  › tauri [RUST]: 1.2.0,
  › tauri-build [RUST]: 1.2.0,
  › tao [RUST]: 0.15.6,
  › wry [RUST]: 0.22.4,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../src
  › devPath: ../src

App directory structure
  ├─ node_modules
  ├─ src-tauri
  ├─ .git
  ├─ .vscode
  └─ src

### Stack trace

_No response_

### Additional context

I've created a repo demonstrating the issue: https://github.com/paprikka/tauri-keypress-bug
And a video: https://drive.google.com/file/d/1ezFteEHw7TAZc-fpXiXOjh2VArkJJHyA/view?usp=sharing

(Thanks for great work! I've been using hybrid frameworks for +11 years, and I'm enjoying Tauri way more than anything else I've worked on 👌🏼)

wusyong:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
fix https://github.com/tauri-apps/tauri/issues/5685

### DIFF
--- a/src/webview/wkwebview/mod.rs
+++ b/src/webview/wkwebview/mod.rs
@@ -307,8 +307,12 @@ impl InnerWebView {
             // So we pass the event to menu if we have one
             extern "C" fn key_down(this: &mut Object, _: Sel, event: id) {
               unsafe {
-                let superclass: *const Class = msg_send![this, superclass];
-                let () = msg_send![super(this, &*superclass), keyDown: event];
+                let keycode: u16 = msg_send![event, keyCode];
+                // Arrow key should not send for text
+                if !(123..=126).contains(&keycode) {
+                  let superclass: *const Class = msg_send![this, superclass];
+                  let () = msg_send![super(this, &*superclass), keyDown: event];
+                }
 
                 let app = cocoa::appkit::NSApp();
                 let menu: id = msg_send![app, mainMenu];
--- a/src/webview/wkwebview/mod.rs
+++ b/src/webview/wkwebview/mod.rs
@@ -308,7 +308,7 @@ impl InnerWebView {
             extern "C" fn key_down(this: &mut Object, _: Sel, event: id) {
               unsafe {
                 let keycode: u16 = msg_send![event, keyCode];
-                // Arrow key should not send for text
+                // Arrow keys are handled by other event chains
                 if !(123..=126).contains(&keycode) {
                   let superclass: *const Class = msg_send![this, superclass];
                   let () = msg_send![super(this, &*superclass), keyDown: event];
--- a/.changes/mac-arrow.md
+++ b/.changes/mac-arrow.md
@@ -1,0 +1,6 @@
+---
+"wry": patch
+---
+
+On macOS, fix arrow keys misprint text on textarea.
+


wusyong:
Why are we still here? Just to suffer.

src/webview/wkwebview/mod.rs
```diff
@@ -307,8 +307,12 @@ impl InnerWebView {
             // So we pass the event to menu if we have one
             extern "C" fn key_down(this: &mut Object, _: Sel, event: id) {
               unsafe {
-                let superclass: *const Class = msg_send![this, superclass];
-                let () = msg_send![super(this, &*superclass), keyDown: event];
+                let keycode: u16 = msg_send![event, keyCode];
+                // Arrow key should not send for text
```
amrbashir: won't this disable `keydown` js events for Arrow keys?

wusyong: when I press left arrow, I found the cursor moved left twice. This means there's another key handling outside of this keydown chain. It could be performKeyEquivelant or doCommandBySelector, but I haven't verified which one and key event handling on macOS is a mess already.

But I know it's probably better to improve comment here. What do you think it should be?

amrbashir: Whatever you feel like, just make sure this event https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event is triggered for Arrow keys, otherwise this will be a regression.

wusyong: Yes, I can see keycode 37~40 being triggered.

amrbashir (dismissed):

paprikka:
Wow that was quick 🚀 thanks!

PR #777, opened 2022-11-29
feat: add `Webiew::navigate_to_url`, closes #776

### LINKED ISSUE #776: [feat] Add `window.navigate_to_url`
> This issue has been upstreamed from https://github.com/tauri-apps/tauri/issues/5716 

### Describe the problem

Now, We change the window url in a clumsy way.

The solution comes from  #5377
```rust

main_window.eval(&format!("window.location.replace('http://localhost:{}')", "1234"));
```

### Describe the solution you'd like

I think we can provide a new mod named `location` under window instance, The function of location can be partially similar to [Window.location](https://developer.mozilla.org/en-US/docs/Web/API/Window/location)

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -747,8 +747,8 @@ impl WebView {
     self.webview.set_background_color(background_color)
   }
 
-  pub fn navigate_to_url(&self, url: &str) {
-    self.webview.navigate_to_url(url)
+  pub fn load_url(&self, url: &str) {
+    self.webview.load_url(url)
   }
 }
 
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -746,6 +746,10 @@ impl WebView {
   pub fn set_background_color(&self, background_color: RGBA) -> Result<()> {
     self.webview.set_background_color(background_color)
   }
+
+  pub fn navigate_to_url(&self, url: &str) {
+    self.webview.navigate_to_url(url)
+  }
 }
 
 /// An event enumeration sent to [`FileDropHandler`].
--- a/.changes/navigate-to-url.md
+++ b/.changes/navigate-to-url.md
@@ -1,0 +1,5 @@
+---
+"wry": "patch"
+---
+
+Add `Webiew::navigate_to_url`.
--- a/.changes/navigate-to-url.md
+++ b/.changes/navigate-to-url.md
@@ -2,4 +2,4 @@
 "wry": "patch"
 ---
 
-Add `Webiew::navigate_to_url`.
+Add `Webiew::load_url`.
--- a/examples/custom_titlebar.rs
+++ b/examples/custom_titlebar.rs
@@ -150,10 +150,8 @@ fn main() -> wry::Result<()> {
         ..
       }
       | Event::UserEvent(UserEvents::CloseWindow) => {
-        webview
-          .as_ref()
-          .unwrap()
-          .navigate_to_url("https://google.com");
+        let _ = webview.take();
+        *control_flow = ControlFlow::Exit
       }
       _ => (),
     }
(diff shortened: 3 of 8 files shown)


src/webview/mod.rs
```diff
@@ -746,6 +746,10 @@ impl WebView {
   pub fn set_background_color(&self, background_color: RGBA) -> Result<()> {
     self.webview.set_background_color(background_color)
   }
+
+  pub fn navigate_to_url(&self, url: &str) {
```
wusyong: How about just `load_url`? This is more clear IMHO.

amrbashir: sure, seems most webviews have it named like that.

wusyong (approved):

patientplatypus6:
Some quick comments, because this looks like it would be important for the usefulness of the library. 
First, this feature is not yet in the main branch of the library and I would encourage you to push it as soon as possible. What I am doing is using an IPC_Channel (through Servo, not between Javascript code) to run this library in conjunction with Rust/Iced.

Here's is what I'm working on - 
https://github.com/patientplatypus6/ipciceexample
Specifically, I'd like to be able to programmatically push new URLs based on state changes that occur at top level - see line marked below. I can save local state in an enum, but I can't push a new URL. 
https://github.com/patientplatypus6/ipciceexample/blob/main/webview/src/main.rs#L84

This is a blocking issue for what I'm working on, so whenever this happens to be pushed into the main branch if you could be so kind as to just respond to this comment so I can repull the repo that would be helpful.

Thanks for all your hard work on this library. Merry Christmas and Happy Holidays.

amrbashir:
@patientplatypus6 this has been released https://docs.rs/wry/latest/wry/webview/struct.WebView.html#method.load_url

patientplatypus6:
Thanks for the reply. It took me a little bit of thinking in how this was supposed to work. Sorry I am unfamiliar with the library. Here is an example page on how you can pass a load_url. 

The important bit is that you can pass the webview that is returned from the builder

```
    let webview = WebViewBuilder::new(window)?
      .with_url(url)?
      .build()?;
```

into the `eventloop.run` which I didn't think was possible for some reason - 

```

    event_loop.run(move |event, x, control_flow| {
      *control_flow = ControlFlow::Wait;
      match event {
        Event::NewEvents(StartCause::Init) => {
            newstate = data_handler(prevstate.clone());
        },
        Event::MainEventsCleared => {
            if Some(newstate.data.clone()) != Some(prevstate.data.clone()) {  
                println!("newstate.data.clone() {:?}", newstate.data.clone());
                let newstatedata = newstate.data.clone().unwrap();
                let newstateparent = newstate.parent.clone().unwrap();
                println!("value of newstatedata: {:?}", newstatedata);
                state_updated(newstatedata, newstateparent);
                prevstate = newstate.clone();
                let newurl = "https://www.reddit.com".to_string();
                proxy.send_event(UserEvent::Navigation(newurl.clone()));
            }
        },
        Event::UserEvent(UserEvent::Navigation(uri)) => {
            println!("{}", uri);
            let url = uri.clone();
            println!("value of url {:?}", url); 
            webview.load_url(&uri);
        },
        Event::WindowEvent {
          ..
        } => *control_flow = ControlFlow::Exit,
        _ => (),
      }
    });
  }
```

`webview.load_url(&uri)` would be available under the arm ` Event::UserEvent(UserEvent::Navigation(uri)) => {` in the match statement. Here is a link to my repo.

https://github.com/patientplatypus6/ipciceexample/blob/main/webview/src/main.rs#L85

Thanks again for the reply and working on this library.

Issue #474, opened 2022-01-31
Feature: Implement Callback Functions for JavaScript Evaluation Result


Toromyx:
**Problem**
I use Tauri to open windows of external sites as well as the "main" application window of local HTML/JS/CSS. I still want to interact with the DOM of those external sites and know the result of that interaction. This doesn't have to be done via JavaScript.

**Solution**
All three webview APIs on the main operating systems support some kind of callback when executing JavaScript. For each os I added a link to the documentation and the place in the code where that callback possibility is just ignored.
Is there a (security) reason this  functionality is not provided?

**Linux**
- https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#webkit-web-view-run-javascript
- https://github.com/tauri-apps/wry/blob/next/src/webview/webkitgtk/mod.rs#L304

**Windows**
- https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.1072.54#executescript
- https://github.com/tauri-apps/wry/blob/next/src/webview/webview2/mod.rs#L561
- https://github.com/tauri-apps/wry/blob/next/src/webview/webview2/mod.rs#L550

**Apple**
- https://developer.apple.com/documentation/webkit/wkwebview/1415017-evaluatejavascript
- https://github.com/tauri-apps/wry/blob/next/src/webview/wkwebview/mod.rs#L414

**Alternatives**
An alternative I considered was to build a JavaScript wrapper which sends the result of the evaluated JavaScript over RPC. But this would be a security problem because now external sites have access to RPC.

**Would you assign yourself to implement this feature?**
I'm only just starting with Rust and would assign myself to do a proof of concept for WebView2 under Windows. I currently don't have the capability to develop under Linux and never will have the capability to develop under Apple.

wusyong:
I think this one is just because of history reason. When webview repo provide the API, it just want one that can run js script.
We do can support this but it won't be a trivial task IMHO. It's better wait for v1 launch and see what feature requests are wanted by most.

Toromyx:
I hacked together something which "works on my machine" (meaning Windows):

- https://github.com/tauri-apps/wry/compare/next...Toromyx:eval-callback
- https://github.com/tauri-apps/tauri/compare/next...Toromyx:eval-callback

I'm open for feedback, but I also understand if you say that now is not the right time to implement this feature.

franky-chanyau:
[...]

PR #778, opened 2022-11-29
feat: support callback function in eval

pewsheen:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This a draft for resolving issue #474. The API should expose to external after most of major platforms are implemented.

* Callback closure in `evaluate_script` should be like `| val: String | { ... }`.
* We will serialize the evaluated result into `JSON string` and pass it to the callback.
* Error from the execution result will not pass to the callback closure due to varying formats returned by each platform. I think this is a minor issue, we can deal with this later.

- [ ] TODO:
  - [x] macOS
  - [x] Unix
  - [x] Windows
  - [ ] Android


Cargo.toml
```diff
@@ -1,47 +1,47 @@
-workspace = { }
+workspace = {}
```
wusyong: You might want to clean Cargo.toml a bit

src/webview/webkitgtk/mod.rs
```diff
@@ -400,13 +404,36 @@ impl InnerWebView {
...
+
+      match callback {
+        Some(callback) => {
+          self.webview.run_javascript(js, cancellable, |result| {
+            let mut result_str = String::new();
+
+            if let Ok(js_result) = result {
+              if let Some(js_value) = js_result.js_value() {
+                unsafe {
+                  result_str = from_glib_full(jsc_value_to_json(js_value.as_ptr(), 0));
```
wusyong: `JSValue` has `to_jason` trait method from `ValueExt`

src/webview/webview2/mod.rs
```diff
@@ -804,17 +804,30 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
...
+    webview: &ICoreWebView2,
+    js: String,
+    callback: impl FnOnce(String) + Send + 'static,
+  ) -> windows::core::Result<()> {
     unsafe {
       webview.ExecuteScript(
         PCWSTR::from_raw(encode_wide(js).as_ptr()),
-        &ExecuteScriptCompletedHandler::create(Box::new(|_, _| (Ok(())))),
+        &ExecuteScriptCompletedHandler::create(Box::new(|_, return_str| {
+          let json_value: serde_json::Value = serde_json::from_str(&return_str).unwrap();
```
wusyong: It seems return_str is already json. Do we need to convert again?

wusyong:
I think we can expose another method called `evaluate_script_with_callback` or something.
We can leave Android unimplemented and open an issue for it later.

src/webview/webview2/mod.rs
pewsheen: No, I just tested it again, and it's unnecessary, I'll remove it.

pewsheen:
I was expected to pass  `Result<>` to the callback if an exception was thrown during the execution. However, it seems that Windows does not return an error but instead responds with the string "null" in the execution result.

It'd be best to wrap a try...catch... clause at the top level of the script if you anticipate that exceptions could occur and you want to handle it.

src/webview/mod.rs
```diff
@@ -813,7 +813,10 @@ impl WebView {
   ///
   /// [`EventLoopProxy`]: crate::application::event_loop::EventLoopProxy
   ///
-  /// - ** Android:** Callback function not implemented.
+  /// Please note that if an exception is thrown during execution, we will ignore the error
+  /// and call the callback function.
```
wusyong:
```suggestion
  /// Exception is ignored because of the limitation on windows. You can catch it your self and return as string as a workaround.
```

wusyong (dismissed):

wusyong (approved):

PR #783, opened 2022-12-03
perf: reduce extra allocation at `WebViewBuilderExtWindows::with_additional_browser_args` argument

rhysd:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #783
- [ ] No

### Checklist

### Other information

Currently `WebViewBuilderExtWindows::with_additional_browser_args` argument is typed with `AsRef<str>`. However the method converts the string slice into string buffer with heap allocation. This causes redundant allocation when `String` value is passed.

```rust
let args: String = ...; // We already have heap-allocated string buffer
builder = builder.with_additional_browser_args(args); // New buffer is allocated
```

This PR fixes the point.

Strictly speaking, changing from `AsRef<str>` to `Into<String>` is a breaking change, but it is very small because major string types like `&str`, `String`, and `Cow<'a, str>` can still be passed to the argument.

### DIFF
--- a/.changes/prefer-into-to-asref-to-reduce-alloc.md
+++ b/.changes/prefer-into-to-asref-to-reduce-alloc.md
@@ -1,0 +1,5 @@
+---
+"wry": "minor"
+---
+
+Change the type of `WebViewBuilderExtWindows::with_additional_browser_args` argument from `AsRef<str>` to `Into<String>` to reduce extra allocation.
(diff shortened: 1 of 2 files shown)


amrbashir (approved):
Very cool, thank you!

PR #790, opened 2022-12-07
fix(macos): suppress `unused_variables` warning reported only in release build

rhysd:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This PR fixes the warning when building this crate in release mode as follows:

```
$ cargo build --release
   Compiling wry v0.23.1 (/Users/rhysd/Develop/github.com/tauri-apps/wry)
warning: unused variable: `description`
   --> src/webview/wkwebview/download.rs:105:9
    |
105 |     let description = NSString(description).to_str().to_string();
    |         ^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_description`
    |

warning: `wry` (lib) generated 1 warning
    Finished release [optimized] target(s) in 2.31s
```

The `description` variable is used only when `debug_assertions` is enabled.

### DIFF
--- a/src/webview/wkwebview/download.rs
+++ b/src/webview/wkwebview/download.rs
@@ -99,17 +99,19 @@ pub extern "C" fn download_did_finish(this: &Object, _: Sel, download: id) {
   }
 }
 
-pub extern "C" fn download_did_fail(this: &Object, _: Sel, download: id, error: id, _: id) {
+pub extern "C" fn download_did_fail(this: &Object, _: Sel, download: id, _error: id, _: id) {
   unsafe {
-    let description: id = msg_send![error, localizedDescription];
-    let description = NSString(description).to_str().to_string();
+    #[cfg(debug_assertions)]
+    {
+      let description: id = msg_send![_error, localizedDescription];
+      let description = NSString(description).to_str().to_string();
+      eprintln!("Download failed with error: {}", description);
+    }
+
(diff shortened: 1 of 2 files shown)


src/webview/wkwebview/download.rs
```diff
@@ -102,6 +102,7 @@ pub extern "C" fn download_did_finish(this: &Object, _: Sel, download: id) {
 pub extern "C" fn download_did_fail(this: &Object, _: Sel, download: id, error: id, _: id) {
   unsafe {
     let description: id = msg_send![error, localizedDescription];
+    #[allow(unused_variables)]
```
wusyong: How about we move these to `debug_assertions` below? And rename `error` to `_error`.

```
#[cfg(debug_assertions)]
{
    let description: id = msg_send![_error, localizedDescription];
    let description = NSString(description).to_str().to_string();
    eprintln!("Download failed with error: {}", description);
}
```

rhysd: Looks better since allocation can be avoided. I'll have a try.

rhysd: @wusyong I revised my patch and force-pushed as e73c4e6

wusyong (approved):

PR #797, opened 2022-12-08
perf: Change return type of custom protocol handler from `Vec<u8>` to `Cow<'static, [u8]>`, closes #796

### LINKED ISSUE #796: Allow custom  protocol handlers to return static resources without heap allocation

The served files such as HTML/CSS/JavaScript are embedded in my application binary like below.

```rust
const SCRIPT: &[u8] = include_bytes!("bundled.js");
const LOGO: &[u8] = include_bytes!("logo.png");
```

And those resources are returned in [a custom protocol handler](https://docs.rs/wry/latest/wry/webview/struct.WebViewBuilder.html#method.with_custom_protocol) as if the response for HTTP request. Currently, custom protocol handlers return `Response<Vec<u8>>` so the body payload must be allocated on heap.

```rust
let webview = WebViewbuilder;;new(window)?
    .with_custom_protocol("MyApp".into(), move |request| {
      let path = request.uri().path();
      let body = match path {
          "/" => SCRIPT.to_vec(), // Unnecessary heap allocation
          "/logo.png" => LOGO.to_vec(), // Unnecessary heap allocation
          _ => fs::read(convert_to_filepath(path))?,
      };

      let mime = ...;

      Response::builder()
        .header(CONTENT_TYPE, mime)
        .body(body)
        .map_err(Into::into)
    });
```

However, static resources like `SCRIPT` or `LOGO` (both are a few MB) are not necessary to be allocated on heap.

In my environment, allocating a single 1~2MB file took about 3ms in release build. When the number of files grows, the overhead will increase.

I'd like to change the return type `Response<Vec<u8>>` to `Response<Cow<'static, [u8]>>` so that static resources are not necessary to be cloned on heap memory. It can return both `&'static [u8]` and `Vec<u8>`.

- Before: `Fn(&Request<Vec<u8>>) -> Result<Response<Vec<u8>>> + 'static`
- After: `Fn(&Request<Vec<u8>>) -> Result<Response<Cow<'static, [u8]>>> + 'static`

The above example can be rewritten as follows with this change:

```rust
let webview = WebViewbuilder;;new(window)?
    .with_custom_protocol("MyApp".into(), move |request| {
      let path = request.uri().path();
      let body = match path {
          "/" => SCRIPT.into(), // Same as Cow::Borrowed(SCRIPT)
          "/logo.png" => LOGO.into(), // Same as Cow::Borrowed(LOGO)
          _ => fs::read(convert_to_filepath(path))?.into(), // Same as Cow::Owned(...)
      };

      let mime = ...;

      Response::builder()
        .header(CONTENT_TYPE, mime)
        .body(body)
        .map_err(Into::into)
    });
```

The drawback of this solution is that this breaks current users' code base. `.into()` is necessary to convert `Vec<u8>` into `Cow<'static, [u8]>`.

Trivial alternative would be continuing to use the current signature not to break current users' code base. And current signature is simpler.

- [x] Yes
- [ ] No

**Additional context**
Nothing.

rhysd:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #796
- [ ] No

### Checklist

### Other information

I confirmed this change using [the custom protocol example](https://github.com/tauri-apps/wry/blob/dev/examples/custom_protocol.rs) on the following platforms:

- Windows 10
- macOS 11.7
- Ubuntu 22.04

I don't have Android and iOS development environment so I could not confirm this patch on it.

### DIFF
--- a/examples/custom_protocol.rs
+++ b/examples/custom_protocol.rs
@@ -34,7 +34,7 @@ fn main() -> wry::Result<()> {
       let content = if path == "/" {
         PAGE1_HTML.into()
       } else {
-        // 1.. for removing leading slash
+        // `1..` for removing leading slash
         read(canonicalize(PathBuf::from("examples").join(&path[1..]))?)?.into()
       };
 
--- a/examples/custom_protocol.rs
+++ b/examples/custom_protocol.rs
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
+const PAGE1_HTML: &[u8] = include_bytes!("custom_protocol_page1.html");
+
 fn main() -> wry::Result<()> {
   use std::{
     fs::{canonicalize, read},
@@ -27,21 +29,19 @@ fn main() -> wry::Result<()> {
   let _webview = WebViewBuilder::new(window)
     .unwrap()
     .with_custom_protocol("wry".into(), move |request| {
-      let path = &request.uri().path();
+      let path = request.uri().path();
       // Read the file content from file path
-      let content = read(canonicalize(PathBuf::from("examples").join(
-        if path == &"/" {
-          "custom_protocol_page1.html"
-        } else {
-          // remove leading slash
-          &path[1..]
-        },
-      ))?)?;
+      let content = if path == "/" {
+        PAGE1_HTML.into()
+      } else {
+        // 1.. for removing leading slash
+        read(canonicalize(PathBuf::from("examples").join(&path[1..]))?)?.into()
+      };
 
       // Return asset contents and mime types based on file extentions
       // If you don't want to do this manually, there are some crates for you.
       // Such as `infer` and `mime_guess`.
-      let mimetype = if path.ends_with(".html") || path == &"/" {
+      let mimetype = if path.ends_with(".html") || path == "/" {
         "text/html"
       } else if path.ends_with(".js") {
         "text/javascript"
--- a/examples/menu.rs
+++ b/examples/menu.rs
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
+const PAGE1_HTML: &[u8] = include_bytes!("custom_protocol_page1.html");
+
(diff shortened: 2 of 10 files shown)


examples/custom_protocol.rs
```diff
@@ -27,21 +29,19 @@ fn main() -> wry::Result<()> {
...
-          "custom_protocol_page1.html"
-        } else {
-          // remove leading slash
-          &path[1..]
-        },
-      ))?)?;
+      let content = if path == "/" {
+        PAGE1_HTML.into()
+      } else {
+        // 1.. for removing leading slash
```
amrbashir:
```suggestion
        // `1..` for removing leading slash
```

examples/menu.rs
```diff
@@ -43,21 +45,19 @@ fn main() -> wry::Result<()> {
...
-          "custom_protocol_page1.html"
-        } else {
-          // remove leading slash
-          &path[1..]
-        },
-      ))?)?;
+      let content = if path == "/" {
+        PAGE1_HTML.into()
+      } else {
+        // 1.. for removing leading slash
```
amrbashir:
```suggestion
        // `1..` for removing leading slash
```

amrbashir (dismissed):
Thank you! LGTM! regardless of the few nitpicks.

amrbashir:
for android, you can just install `rustup target add aarch64-linux-android` and then run `cargo check --target aarch64-linux-android` to make sure it compiles fine.

rhysd:
> for android, you can just install rustup target add aarch64-linux-android and then run cargo check --target aarch64-linux-android to make sure it compiles fine.

Ah, I didn't know that checking non-host target is so easy with cargo. Thank you. I'll check the compile errors and your review points tonight.

rhysd:
@amrbashir @wusyong I fixed Android build and resolved review comments. I apologize that I could not take time to work on this yesterday.

amrbashir (approved):
LGTM! Thank you.

PR #801, opened 2022-12-09
Fix beep sound, closes #799

### LINKED ISSUE #799: Best approch to disable key error sound
After #798, we remove all keyDown implementations. Because key events on macOS is too messy, it's not worth it to implement our own. This will bring back beep sound when pressing key.

The error sound will fire when the view doesn't handle it, eg. there's no menu accelerator for it, or users are not typing in the textarea. To remove it, we can offer a config to add `keyDown` back again with several workaround. But I think we should not handle any accelerator without `Ctrl` and `Cmd` this time. Because the key event chain is totally different without them, and I have struggled to satisfy every scenarios. See this [slides](https://docs.huihoo.com/apple/wwdc/2010/session_145__key_event_handling_in_cocoa_applications.pdf) for more information. Official Apple apps seem to follow this convention too. They don't have acceleartors without `Ctrl` and `Cmd`.

Do you think this is a good idea: add a option to disable error sound but menu accelerator won't work without `Cmd` and `Ctrl`. And should it be default?

probablykasper:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

All I did was have the `WryWebViewParent` call `performKeyEquivalent` on the `menu`. As far as I can tell, everything works:
- The frontend receives `Escape` keydown events
- There's no beep sound. Built-in menus like `Copy` can still beep, but that's correct behavior - Safari acts the same. You can prevent the menu (and by extension, the beep) from being called using `e.preventDefault()` in the frontend.
- `Q` and `Ctrl+Shift+Q` menus work
- Menus only fire once

Closes #799

@wusyong Is there anything I'm missing?

### DIFF
--- a/.changes/beepfix.md
+++ b/.changes/beepfix.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix the beep sound on macOS
--- a/examples/custom_protocol_page1.html
+++ b/examples/custom_protocol_page1.html
@@ -15,8 +15,14 @@
   <h1>Welcome to WRY!</h1>
   <p>Page 1</p>
   <textarea></textarea>
+  <p id="keypresses"></p>
   <a href="/custom_protocol_page2.html">Link</a>
   <script type="text/javascript" src="/custom_protocol_script.js"></script>
+  <script type="text/javascript">
+    document.body.addEventListener('keydown', (event) => {
+      document.querySelector('#keypresses').innerHTML += ' ' + event.key
+    })
+  </script>
 </body>
 
 </html>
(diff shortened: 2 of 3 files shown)


wusyong (approved):
Following bugs we encountered bugs before have tested:
- no beep sound
- keyboard input working including IME
- menu accelerator works with: CMD+, SHIFT+, and single keyinput in menu example.
- keyinput only reigger once using sample in https://github.com/tauri-apps/tauri/issues/5741
- works with release build with this config:
```
[profile.release]
strip = true 
lto = true
opt-level = "z"
```

wusyong:
@probablykasper Thanks again! This really fix everything around key events I believe.

probablykasper:
That's awesome! Pure luck that I tried that haha

PR #825, opened 2022-12-26
feat: add document title changed handler, closes #804

### LINKED ISSUE #804: callback for when document.title is changed?
If a document in a WebView sets `document.title`, it would be nice to get a some kind of notification. This could be used to update the window title. Electron has a `page-title-updated` event, Of course this is only needed when you don't control the JavaScript in the window, so the priority is probably low.

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist
- [X] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [X] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/wry/blob/dev/.changes/readme.md).

### DIFF
--- a/src/webview/wkwebview/mod.rs
+++ b/src/webview/wkwebview/mod.rs
@@ -387,8 +387,9 @@ impl InnerWebView {
         let cls = match cls {
           Some(mut cls) => {
             cls.add_ivar::<*mut c_void>("function");
+            cls.add_ivar::<id>("webview");
             cls.add_method(
-              sel!(observeValueForKeyPath),
+              sel!(observeValueForKeyPath:ofObject:change:context:),
               observe_value_for_key_path as extern "C" fn(&Object, Sel, id, id, id, id),
             );
             extern "C" fn observe_value_for_key_path(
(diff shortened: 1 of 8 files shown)


src/webview/wkwebview/mod.rs
```diff
@@ -406,8 +407,9 @@ impl InnerWebView {
                   if !function.is_null() {
                     let function = &mut *(*function
                       as *mut (Box<dyn for<'r> Fn(&'r Window, String)>, Rc<Window>));
-                    let title = NSString(change.valueForKey_(key_path)).to_str().to_string();
-                    (function.0)(&function.1, title);
+                    let webview = this.get_ivar::<id>("webview");
+                    let title: id = msg_send![*webview, title];
```
wusyong: `_of_object` is the source object. You can just use it directly:
```
let title: id = msg_send![_of_object, title];
```
But probably rename to object is better.

wusyong (changes requested):

src/webview/wkwebview/mod.rs
amrbashir: updated but kept the name as `of_object` as it seemed appropriate to keep close to the apple docs.

lucasfernog (approved):

PR #826, opened 2022-12-27
feat: add headers when loading URLs, closes #816

### LINKED ISSUE #816: Additional HTTP header with loading URL
I want to open a URL that needs additional custom http header. Is there any way to add custom http header?

It seems that android webview has a feature what I want.
https://developer.android.com/reference/android/webkit/WebView#loadUrl(java.lang.String,%20java.util.Map%3Cjava.lang.String,%20java.lang.String%3E)

It would be good to add some function for WebviewBuilder like below.

```

let _webview = WebViewBuilder::new(window)?
        .with_url(&some_url)?     
        .with_http_header("some", "header")
        .build()?;
```

I tried to add header with initialization_script but it failed.

- [ ] Yes
- [x] No

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/src/webview/webview2/mod.rs
+++ b/src/webview/webview2/mod.rs
@@ -852,13 +852,13 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
     set_background_color(&self.controller, background_color).map_err(Into::into)
   }
 
-  pub fn load_url(&self, url: &str) {
-    let url = encode_wide(url);
-    let _ = unsafe { self.webview.Navigate(PCWSTR::from_raw(url.as_ptr())) };
-  }
-
-  pub fn load_url_with_headers(&self, url: &str, headers: http::HeaderMap) {
-    load_url_with_headers(&self.webview, &self.env, url, headers);
+  pub fn load_url(&self, url: &str, headers: Option<http::HeaderMap>) {
+    if let Some(headers) = headers {
+      load_url_with_headers(&self.webview, &self.env, url, headers);
+    } else {
+      let url = encode_wide(url);
+      let _ = unsafe { self.webview.Navigate(PCWSTR::from_raw(url.as_ptr())) };
+    }
   }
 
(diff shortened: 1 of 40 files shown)


src/webview/webview2/mod.rs
```diff
@@ -828,6 +859,35 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
     let _ = unsafe { self.webview.Navigate(PCWSTR::from_raw(url.as_ptr())) };
   }
 
+  pub fn load_url_with_headers(&self, url: &str, headers: http::HeaderMap) {
```
lucasfernog, line 860: we should try to reuse this function in the Webview::new code too, it's too big to just copy paste it.

amrbashir: fixed now! I forgot to remove the duplicate code and the one in `Webview::new` was missing the `GET` method.

lucasfernog (dismissed):

lucasfernog (dismissed):

lucasfernog:
Initially I thought `with_headers` would apply to all URLs which is not the case, so maybe we chould change the signature to `with_url_and_headers`.

lucasfernog:
I've changed the signatures to use the existing methods instead of adding new ones. So now `with_url`, `with_html` and `load_url` takes an `Option<http::HeaderMap>`. Let me know what you think @amrbashir I think this is less confusing to users.

amrbashir:
There is no point in using headers with `with_html` since it loads local content.
Also, most users won't use headers, so I think it should be a separate function. I like the idea of `with_url_and_headers` instead.

lucasfernog:
Cool @amrbashir

lucasfernog (approved):

PR #829, opened 2022-12-29
refactor: improve android env vars naming

dvc94ch:
Renames the wry environment variables to `WRY_ANDROID_PACKAGE` and `WRY_ANDROID_LIBRARY`. These make the intent clearer. Also avoids wry adding an unwanted postfix to the package name (app-snake-case-name).


amrbashir (approved):
Thanks

PR #836, opened 2023-01-08
fix empty string bug (fix: #833)

### LINKED ISSUE #833: Initialisation Script of `""` crashes the Webview
**Describe the bug**
Forgive me for using bad terminology: I'm a JS guy...

I'm building a wrapper around Wry which uses child processes to spawn Webviews (because technically they're considered enough of a "main thread" to let it work - idk why). I'm making this into as much of a flexible API as I can, and so when it came to implementing the WebView's options, one caused my WebViews to crash: `.with_initialization_script`. Passing it a parameter of `""` kept causing the crashes, even though it works as expected when you send `"()=>{}"` which leads me to believe that sending an empty string causes incorrect behaviour.

After doing a very very very shallow dive of the code on [docs.rs](https://docs.rs/wry/latest/src/wry/webview/mod.rs.html#366-369), it looks like (on the surface at least) that many init scripts can be included, and they'll all be added to a big chunk prior to passing to the platform specific API. If that's the case then I don't know why an empty script would cause issues, because an empty script is valid JS?

**Steps To Reproduce**
1. Create a Wry WebView with `.with_initialization_script("()=>{}")`
2. Successful run
3. Modify the init script to be `""` (an empty string)
4. Run and expect crash

**Expected behavior**
`.with_initialization_script("")` should not cause a crash.

OS: Windows 11 Pro, `10.0.22621`
Rustc: `rustc 1.66.0 (69f9c33d7 2022-12-12)`

- [ ] Yes
- [x] No (I don't Rust - this shit too hard. 😞)

**Additional context**
If there's any insight as to why this bug happens, or if it's platform specific, I'd love to know!

TheBrenny:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #833
- [x] I have added a convincing reason for adding this feature, if necessary (not necessary - otherwise look at #833)
- [x] It can be built on all targets and pass CI/CD. (I imagine so. I changed one line, and forgot to enable CI/CD before pushing to my branch.)

### Other information
The bug appears to only affect Windows (macOS untested, but highly unlikely to be affected). The change catches empty strings higher in the call-chain - specifically in the "userland" function - as a just in case.

### DIFF
--- a/.changes/check-for-empty-init-scripts.md
+++ b/.changes/check-for-empty-init-scripts.md
@@ -1,0 +1,5 @@
+---
+"wry": "patch"
+---
+
+Ensures that the script passed to `.with_initialization_script("here")` is not empty.
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -374,7 +374,9 @@ impl<'a> WebViewBuilder<'a> {
   /// - **Android:** The Android WebView does not provide an API for initialization scripts,
   /// so we prepend them to each HTML head. They are only implemented on custom protocol URLs.
   pub fn with_initialization_script(mut self, js: &str) -> Self {
-    self.webview.initialization_scripts.push(js.to_string());
+    if !js.is_empty() {
+      self.webview.initialization_scripts.push(js.to_string());
+    }
     self
   }
 


wusyong (approved):

wusyong:
Thanks for the PR!

PR #880, opened 2023-02-13
fix(deps): Use less specific version for once_cell

FabianLars:
ref: https://discord.com/channels/616186924390023171/1074435079964725329

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ linux-headers = [ "webkit2gtk/v2_36" ]
 [dependencies]
 libc = "0.2"
 log = "0.4"
-once_cell = "1.15"
+once_cell = "1"
 serde = { version = "1.0", features = [ "derive" ] }
 serde_json = "1.0"
 thiserror = "1.0"


amrbashir:
we need to be careful with this one, once_cell author has bumped MSRV before in a patch version? don't remember exactly and it is unlikely that once_cell will bump MSRV again anytime soon but we should still be careful

FabianLars:
this is actually why i opened this pr, they increased the MSRV in 1.15 and there are crates locking it to 1.14 because of that. So us setting a _minimum_ version this high is not good for compatibility. ("this high" was meant for the dev branch where we just always merge the latest renovate pr. it's on 1.17 right now)

amrbashir:
So setting it to `1` would allow cargo to choose a compatible version?

FabianLars:
yeah, then cargo would be allowed to just use 1.14 for all deps for example.

`1` is basically the same as `1.0` (or `^1.0` in npm land) just without renovate bothering us on each new minor version.

iirc we started adding less specific versions back when someone reported that they had to pin tokio to a slightly lower version but we always merged the latest renovate pr into tauri.

But yes, this will not prevent msrv bumps. cargo does not factor that in unfortunately.

PR #892, opened 2023-02-24
fix: Added thread sleep to resize event to hack around bad resize performance (fix #tauri/issues/4012)

metkm:

- [x] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve https://github.com/tauri-apps/tauri/issues/4012

### Other information
Also kind of related to another issue https://github.com/tauri-apps/tauri/issues/6322

### DIFF
--- a/.changes/resize-delay.md
+++ b/.changes/resize-delay.md
@@ -1,0 +1,5 @@
+---
+"wry": minor
+---
+
+Added resize-delay feature for better resize performance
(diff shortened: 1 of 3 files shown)


amrbashir:
Thanks for the PR but looks like we don't need to include it in tauri or wry at all, you can add this directly in your tauri app code and it will still work:
```rs
 tauri::Builder::default()
        .on_window_event(|e| {
            if let WindowEvent::Resized(_) = e.event() {
                std::thread::sleep(std::time::Duration::from_nanos(1));
            }
        })
        .run(tauri::generate_context!())
        .expect("error while running tauri application");
```

PR #900, opened 2023-03-08
feat: synthesize forward/backward mouse button on Linux and macOS

amrbashir:

- [X] Bugfix
- [X] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

### Checklist
- [X] This PR will resolve https://github.com/tauri-apps/tauri/issues/4019
- [X] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/wry/blob/dev/.changes/readme.md).
- [X] I have added a convincing reason for adding this feature, if necessary


wusyong (changes requested):
I feel like this should move into another file as it starts to grow larger.
And could you add some doc comments about what they do? Especially some numbers about buttons.

amrbashir:
Should be good to go now! Although I want to wait for confirmation if this synthetic event is needed on macOS too

wusyong (dismissed):

wusyong:
I don't have a mouse for mac. And I don't have a hub that can let me plug my PC mouse either. Is there anyone can report this?

goenning:
hey @wusyong and @amrbashir I've been following this closely as I'm interested on this feature. Anything I can do to help test this on a mac?

amrbashir:
I still haven't implemented this on macOS yet, still researching. I will ping you when there is some progress. Thanks @goenning

pewsheen:
Hmm... I can't receive forward / back mouse events on the Mac either.

No event emitted from `mousedown`, `mouseup`, `click`
```
window.addEventListener('mousedown', (e) => {
  console.log(e.button);
});
```

The mouse event for forward / back seems not standard in W3C?
https://github.com/w3c/pointerevents/issues/191

And it looks like Chromium did a particular patch for it
[Feature: Send mouse back/forward buttons to javascript](https://chromestatus.com/feature/5088301178421248)


However, the page forward and back are still working on my Mac.
There might be some system mouse events to trigger the action. Maybe we can catch the mouse event from cocoa window.

pewsheen:
Previously I was using my logi mx anywhere and these's no event emitted, so I tried another mouse on Mac, the events are emitted with button '1'.

wusyong:
I'm not sure if we want to include this after reading w3c issue above. Looks like it's not standard at all. Only chrome did a edge case patch for it.

amrbashir:
it is listed in MDN as supported plus almost all browsers implement back/forward navigation

src/webview/wkwebview/mod.rs
```diff
@@ -297,6 +301,8 @@ impl InnerWebView {
         _ => class!(WryWebView),
       };
       let webview: id = msg_send![cls, alloc];
+      (*webview).set_ivar("webview", webview);
```
wusyong, line 308: I don't think this is required. Can we just call `this`?

amrbashir: I don't know if the PR even works on macOS at all, so feel free to adjust as you see fit

wusyong: alright, I'll test tmr. Or maybe @pewsheen can also give it a try.

src/webview/wkwebview/synthetic_mouse_events.rs
```diff
@@ -0,0 +1,123 @@
...
+        match button_number {
+          4 => {
+            let js = create_js_mouse_event(*webview, event, true, true);
+            let _: id = msg_send![*webview, evaluateJavaScript:NSString::new(&js) completionHandler:null::<*const c_void>()];
+          }
+          5 => {
+            let js = create_js_mouse_event(*webview, event, true, false);
+            let _: id = msg_send![*webview, evaluateJavaScript:NSString::new(&js) completionHandler:null::<*const c_void>()];
+          }
+
```
amrbashir: I have no idea if this is correct, I couldn't what the values of `even.buttonNumber` means so I speculated they would be 4 & 5 but please do test this and maybe throw a `dbg!` here to see what the values are actually. 

Also I am not sure if `x` and `y` positions are correct so please test that too and try to compare it with the same event in chrome or something.

cc @pewsheen @goenning

pewsheen: The button numbers are 3 & 4.

`x` is correct, but `y` is upside down.

src/webview/wkwebview/synthetic_mouse_events.rs
```diff
@@ -0,0 +1,123 @@
...
+unsafe fn create_js_mouse_event(view: id, event: id, down: bool, back_button: bool) -> String {
+  let event_name = if down { "mousedown" } else { "mouseup" };
+  let (button, buttons) = if back_button { (3, 8) } else { (4, 16) };
+  let mods_flags = event.modifierFlags();
+  let window_point = event.locationInWindow();
+  let view_point = view.convertPoint_fromView_(window_point, nil);
+  let view_rect = NSView::frame(view);
+  let x = view_point.x as f64;
+  let y = view_rect.size.height as f64 - view_point.y as f64;
```
pewsheen: Looks like we should leave it to `view_point.y` only

```rust
let y = view_point.y as f64;
```

src/webview/wkwebview/synthetic_mouse_events.rs
amrbashir: thanks, please give it another test and feel free to modify the PR directly with any fixes you do

pewsheen: Tested, and it looks good!

amrbashir: thanks @pewsheen

amrbashir:
I guess all that is left now is:
1. [x] [macOS] test that normal mouse events are patched through to the next responder ( this commit https://github.com/tauri-apps/wry/pull/900/commits/189f9600314d35385ac9aa44ef793ab73609d43d)
2. [x] [macOS] test that there isn't any duplicate mouse events for back/forward buttons in javascript
3. [ ] [macOS & Linux] properly populate the js event `buttons` field which should contain a combination of all pressed mouse buttons when that event is fired (currently it only have the back/forward button)

amrbashir:
@pewsheen could you check the first 2 items on the above list and I will try to implement 3 later today

pewsheen:
I guess the events you meant on the first item are OtherMouse except back and forward; if so, both 1 & 2 look fine on my side.

amrbashir:
@pewsheen could you give another test for `event.buttons`, you may use this site https://www.onlinemictest.com/mouse-test/ to see if clicking the two buttons at the same time light up?

cc @wusyong @goenning @Narcha if you guys have a linux to test on too, that would be appreciated (I don't currently have access to a mouse with back/forward buttons).

rmburg:
I've tested my mouse using the link you provided on Linux (X11).
On Firefox, neither side button is shown when held down.
On Chromium, both side buttons are shown.
All other buttons (left, right, mouse wheel) work in both browsers as expected.

amrbashir:
thanks @Narcha I was hoping you'd test using this PR

rmburg:
Alright, sorry :sweat_smile: Do you have a demo tauri app I can test this with?

goenning:
Can you share some instructions on how to use this PR on my app?

I can test on Linux and Mac

amrbashir:
```
git clone https://github.com/tauri-apps/wry
cd wry
git checkout feat/mouse-backward-forward-synthesized
```
then change the url in `examples/hello_world.rs` to this url `https://www.onlinemictest.com/mouse-test/`, then:
```
cargo run --example hello_world
```

rmburg:
On my machine, using the setup from your last comment, both side buttons do not light up when clicked. The other buttons work as expected.

amrbashir:
what about manually listening to `mousedown` in the developer console?

rmburg:
Curiously, the manual event listener works.
image

amrbashir:
what about `e.buttons`?

rmburg:
e.buttons is 1/2/4 for the first three buttons (as expected), but zero for both side buttons.
image

amrbashir:
thanks @Narcha for testing, my new mouse will arrive soon, and I will try to investigate it further.

pewsheen:
Hmm. I just found out that although otherMouse events were passed to nextResponder, I can't receive the event from Javascript.

For the `buttons`, `NSEvent::pressedMouseButtons` already gives the correct code, including back and forward, so we don't have to calculate it ourselves.

By the way, the js event fix only applies when the final click is a backward or forward button.  So the `buttons` might get wrong if we pressed a back / forward button first and then a regular button.

amrbashir:
> For the `buttons`, `NSEvent::pressedMouseButtons` already gives the correct code, including back and forward, so we don't have to calculate it ourselves.

cool, didn't really think that it matches the js behavior by default.

> Hmm. I just found out that although otherMouse events were passed to nextResponder, I can't receive the event from Javascript.

What buttons exactly did you not receive?

> By the way, the js event fix only applies when the final click is a backward or forward button. So the `buttons` might get wrong if we pressed a back / forward button first and then a regular button.

Not sure I understand, please explain more

pewsheen:
> What buttons exactly did you not receive?
> 

middle button

Also, I found a doc [Event Handling Basics](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/EventHandlingBasics/EventHandlingBasics.html#//apple_ref/doc/uid/10000060i-CH5-SW11) says it's better to pass the event to superclass than call the next responder.
Therefore I have tried its superclass and itself. Only `this` and instance of `NSView` could work, I can receive the event from javascript event listener. I'm not sure if using `this` here is good enough.
```rust
let _: () = msg_send![this, mouseDown: event];
```

> Not sure I understand, please explain more
>
We calculate the value of `buttons` in the function `create_js_mouse_event` and `create_js_mouse_event` will be called only when we press or release the back or forward button.

So we have some scenarios:
LEFT + BACK -> BACK button pressed at the end and triggered `create_js_mouse_event`
BACK + LEFT -> LEFT button pressed at the end and triggered default handler instead of `create_js_mouse_event`

If the value of `buttons` is correct with those BACK / FORWARD buttons on default behavior, then it's totally fine.

amrbashir:
> e.buttons is 1/2/4 for the first three buttons (as expected), but zero for both side buttons.

@Narcha I could only fix this if the last pressed button is the forward/backward button, other mouse buttons are not handled by us so I can't inject it, see below.

> We calculate the value of `buttons` in the function `create_js_mouse_event` and `create_js_mouse_event` will be called only when we press or release the back or forward button.
> 
> So we have some scenarios: LEFT + BACK -> BACK button pressed at the end and triggered `create_js_mouse_event` BACK + LEFT -> LEFT button pressed at the end and triggered default handler instead of `create_js_mouse_event`
> 
> If the value of `buttons` is correct with those BACK / FORWARD buttons on default behavior, then it's totally fine.

Yeah, I think that is fine for now, I don't want to replicate all mouse button events tbh but will discuss it later if someone asks for it.

amrbashir:
> Also, I found a doc [Event Handling Basics](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/EventHandlingBasics/EventHandlingBasics.html#//apple_ref/doc/uid/10000060i-CH5-SW11) says it's better to pass the event to superclass than call the next responder.
> Therefore I have tried its superclass and itself. Only `this` and instance of `NSView` could work, I can receive the event from javascript event listener. I'm not sure if using `this` here is good enough.

@pewsheen could you test this and see if it passes the event correctly? thanks

pewsheen:
> > Also, I found a doc [Event Handling Basics](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/EventHandlingBasics/EventHandlingBasics.html#//apple_ref/doc/uid/10000060i-CH5-SW11) says it's better to pass the event to superclass than call the next responder.
> > Therefore I have tried its superclass and itself. Only `this` and instance of `NSView` could work, I can receive the event from javascript event listener. I'm not sure if using `this` here is good enough.
> 
> @pewsheen could you test this and see if it passes the event correctly? thanks

I just pushed a commit and tested it on my Mac, and it seems to be working properly.

amrbashir:
The linux implementation now should make `https://www.onlinemictest.com/mouse-test/` light up the back/forward button correctly, @pewsheen do they lightup on macOS too?

pewsheen:
> The linux implementation now should make `https://www.onlinemictest.com/mouse-test/` light up the back/forward button correctly, @pewsheen do they lightup on macOS too?

Yes.

amrbashir:
Cool, thanks @pewsheen for your awesome help and @Narcha @goenning for testing.

@wusyong I think this PR is ready to be merged.

.changes/linux-mouse-backward-forward.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": "minor"
+---
+
+On Linux, add synthesized event for mouse backward and forward buttons.
```
amrbashir:
```suggestion
On Linux and macOS, add synthesized event for mouse backward and forward buttons.
```

wusyong:
I just remember webkitgtk and wkwebview's webview instance can call back and forward. Could we just call these methods?
https://webkitgtk.org/reference/webkit2gtk/stable/method.WebView.go_forward.html
https://developer.apple.com/documentation/webkit/wkwebview?language=objc

amrbashir:
We can't use these APIs because we still want to let users disable the navigation by listening to `mouseup` event and calling `event.preventDefault()` so we have to do the navigation on the JS side

wusyong (approved):

PR #907, opened 2023-03-20
fix(ios): set scroll bounce default to NO

pewsheen:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### Other information

Relates to #557, this PR will set scroll bounce to default as NO on iOS.

Prior to this PR, scroll bounce was set to default as YES. With this patch, the behavior will be changed. Given the usage scenario and the fact that the mobile platform is still in the alpha stage, I believe it's acceptable to implement this change at this time.

### DIFF
--- a/.changes/ios-default-scroll-bounce.md
+++ b/.changes/ios-default-scroll-bounce.md
@@ -1,0 +1,5 @@
+---
+"wry": "patch"
+---
+
+On iOS, set webview scroll bounce default to NO.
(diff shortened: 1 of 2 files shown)


wusyong (approved):

daveycodez:
How do we re-enable scroll bounce ? This needs to be in Tauri config not Rust

PR #913, opened 2023-03-27
feat: Add setting to enable autoplay

FabianLars:
3 Open tasks/questions i need help with:
- I have no idea how to make the android one configurable
- The macos implementation didn't work the last time we tested the autoplay-0.19 branch. I can't test it myself because either the webview or my whole vm crashses every time media is involved.
- I'm not sure what the default should be. For actual apps enabling autoplay makes sense, but for displaying external URLs not so much, i went with enabling it by default since this is what was used in android already.

Tauri request: https://github.com/tauri-apps/tauri/issues/3478

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [ ] It can be built on all targets and pass CI/CD. -> Not yet

### DIFF
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -229,6 +229,10 @@ pub struct WebViewAttributes {
   pub document_title_changed_handler: Option<Box<dyn Fn(&Window, String)>>,
 
   /// Whether all media can be played without user interaction.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Android:** Autoplay will always be enabled on window creation.
   pub autoplay: bool,
 }
 
@@ -653,6 +657,7 @@ pub trait WebViewBuilderExtWindows {
   ///
   /// By default wry passes `--disable-features=msWebOOUI,msPdfOOUI,msSmartScreenProtection`
   /// so if you use this method, you also need to disable these components by yourself if you want.
+  /// Furthermore you need to add `--autoplay-policy=no-user-gesture-required` if autoplay is enabled.
   fn with_additional_browser_args<S: Into<String>>(self, additional_args: S) -> Self;
 
   /// Determines whether browser-specific accelerator keys are enabled. When this setting is set to
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -232,7 +232,7 @@ pub struct WebViewAttributes {
   ///
   /// ## Platform-specific:
   ///
-  /// - **Android:** Autoplay will always be enabled on window creation.
+  /// - **Android:** Unsupported. Autoplay will always be enabled in new windows.
   pub autoplay: bool,
 }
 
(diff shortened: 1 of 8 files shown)


amrbashir (commented):
For Android, you need to add a method like `loadUrlMainThread` and generate a rust binding for it and invoke it whenever you need to change this setting

FabianLars:
> whenever you need to change this setting

I'd match it with the desktop platforms where this can only be set when creating the webview. I _hoped_ one can modify the class constructor or something but i guess this will work too, thanks 👍 

i'm gonna see if i can implement it at some point, i currently don't have enough disk space to set up the android dev env but i guess i have enough time until someone fixes macos anyway :D

wusyong:
@FabianLars It's okay to leave Android unsupported and add document to mention this. I can test others if you want to promote to PR.

FabianLars:
Well, the old autoplay-0.19 was once used in semi-production ready apps so windows and linux should work, so that part would be ready for review.

I just added a note about the android specific behavior so that we could potentially fix that in a follow-up PR.

That leaves us with the macOS part which i can't implement/test because my VM can't handle media playback. I just reverted the macos change to something a bit more sane so if someone could try that real quick? I _think_ i had someone test that variant already but worth to double check.
If it doesn't work, i can remove it from this PR like the android implementation if you want 🤷

src/webview/wkwebview/mod.rs
```diff
@@ -315,6 +315,12 @@ impl InnerWebView {
 
       let _: id = msg_send![_preference, setValue:_yes forKey:NSString::new("allowsPictureInPictureMediaPlayback")];
 
+      if attributes.autoplay {
+        let none: NSUInteger = 0;
+        // FIXME: No idea how to make this not crash the webview, we need to set it to `WKAudiovisualMediaTypeNone` which _should_ be `0` https://developer.apple.com/documentation/webkit/wkaudiovisualmediatypes/wkaudiovisualmediatypenone
+        let _: id = msg_send![config, setValue:none forKey:NSString::new("mediaTypesRequiringUserActionForPlayback")];
```
wusyong:
```suggestion
        let _: id = msg_send![config, setMediaTypesRequiringUserActionForPlayback:0];
```

src/webview/mod.rs
```diff
@@ -643,6 +657,7 @@ pub trait WebViewBuilderExtWindows {
   ///
   /// By default wry passes `--disable-features=msWebOOUI,msPdfOOUI,msSmartScreenProtection`
   /// so if you use this method, you also need to disable these components by yourself if you want.
+  /// Furthermore you need to add `--autoplay-policy=no-user-gesture-required` if autoplay is enabled.
```
wusyong: Does this still apply? I saw there's a if statement in webview2's code.

src/webview/mod.rs
```diff
@@ -227,6 +227,13 @@ pub struct WebViewAttributes {
 
   /// Set a handler closure to process the change of the webview's document title.
   pub document_title_changed_handler: Option<Box<dyn Fn(&Window, String)>>,
+
+  /// Whether all media can be played without user interaction.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Android:** Autoplay will always be enabled on window creation.
```
wusyong: Maybe we can say it's not supported yet and point to the line to get started.

wusyong (changes requested):

src/webview/mod.rs
FabianLars: Yes, the additional_browser_args settings will overwrite the autoplay flag similar to how it overwrites wry's default --disable-features stuff.

But i think i messed up the wording a bit, i'll try to make it more clear.

wusyong (dismissed):
I'm okay with Android not supported yet, so this looks good to me.

amrbashir:
I will implement Android

amrbashir (dismissed):

wusyong:
@FabianLars Can you rebase/merge the branch? It's good to go after this.

wusyong (approved):

PR #915, opened 2023-03-28
feat: add `Webview::clear_browsing_data`

### LINKED ISSUE #914: [feat] API to clear webview-managed data
> This issue has been upstreamed from https://github.com/tauri-apps/tauri/issues/6567 

### Describe the problem

My application uses cookies/localStorage on the javascript side for various purposes.

I don't see a clear way to clear all the stuff stored by the Webview engine. I could have the javascript side clear these, but there are things that this side can't clear like disk cache, etc

### Describe the solution you'd like

I want to put a native button like 'Logout' that clears all this data. 

An API on app/app_handle that clears this data might make sense. Maybe it needs to be part of exit or relaunch?

### Alternatives considered

I considered manually deleting some of the path_resolver() directories, but directories like `/Users/pttamkin/Library/HTTPStorage/{pkg}` and `/Users/pttamkin/Library/WebKit/{pkg}` aren't captured here. Also not sure I can delete these while my app is running?

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #914 


amrbashir:
@wusyong I am not sure if macOS implementation works so do test it before merging.

src/webview/android/kotlin/RustWebView.kt
```diff
@@ -33,5 +33,16 @@ class RustWebView(context: Context): WebView(context) {
...
     }
 
+    fun clearAllBrowsingData() {
+        try {
+            super.context.deleteDatabase("webviewCache.db");
+            super.context.deleteDatabase("webview.db");
+            super.clearCache(true);
+            super.clearHistory();
+            super.clearFormData();
+        } catch (ignore: Exception) { }
```
wusyong: Could you add a log about the exception?

amrbashir: done

src/webview/wkwebview/mod.rs
```diff
@@ -907,6 +907,15 @@ r#"Object.defineProperty(window, 'ipc', {
     self.navigate_to_url(url, Some(headers))
   }
 
+  pub fn clear_all_browsing_data(&self) -> Result<()> {
+    unsafe {
+      let store: id = msg_send![class!(WKWebsiteDataStore), defaultDataStore];
+      let all_data_types: id = msg_send![class!(WKWebsiteDataStore), allWebsiteDataTypes];
+      let _: () = msg_send![store, removeDataOfTypes:all_data_types modifiedSince:null::<*const c_void>() completionHandler:null::<*const c_void>()];
```
amrbashir: I am not so sure about this implementation at all and someone with macOS needs to test it

wusyong: I tested it by setting `document.cookie` manually and call this method.
But I agree more people to test is always better.

amrbashir: I haven't seen that you fixed the implementation in [36f3d5f](https://github.com/tauri-apps/wry/pull/915/commits/36f3d5f9310eab78117a7afe61365133a736ea5b), I was just concerned with my own implementation. This should be fine to merge now I guess

wusyong (approved):

PR #916, opened 2023-03-29
feat: Add a way to launch WebViews as incognito `WebView::as_incognito`, closes #908

### LINKED ISSUE #908: Launch Webviews as incognito
For browser based OAuth where you cannot guarantee the existence of a web browser but can guarantee the existence of a WebView, having an option to launch a webview as incognito would be useful.

- [x] Yes
- [ ] No

Other (I did a PoC for WebView2)

**Additional context**
On Windows, it could be done by modifying ```InnerWebView::create_controller``` to something like this:
```
  fn create_controller(
    hwnd: HWND,
    env: &ICoreWebView2Environment,
    incognito: bool
  ) -> webview2_com::Result<ICoreWebView2Controller> {
    let (tx, rx) = mpsc::channel();
    let env = env.clone().cast::<ICoreWebView2Environment10>()?;
    let controller_opts = unsafe {env.CreateCoreWebView2ControllerOptions()?};

    unsafe {controller_opts.SetIsInPrivateModeEnabled(incognito)?}

    CreateCoreWebView2ControllerCompletedHandler::wait_for_async_operation(
      Box::new(move |handler| unsafe {
        env
          .CreateCoreWebView2ControllerWithOptions(hwnd, &controller_opts, &handler)
          .map_err(webview2_com::Error::WindowsError)
      }),
      Box::new(move |error_code, controller| {
        error_code?;
        tx.send(controller.ok_or_else(|| windows::core::Error::from(E_POINTER)))
          .expect("send over mpsc channel");
        Ok(())
      }),
    )?;

    rx.recv()
      .map_err(|_| webview2_com::Error::SendError)?
      .map_err(webview2_com::Error::WindowsError)
  }
```
It would be needed to cast ```env``` to ICoreWebView2Environment10 first as ICoreWebView2Environment wouldn't have CreateCoreWebView2ControllerWithOptions.

Hyphrio:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #908 

### Other information
This adds an option for `WebViewBuilder` to allow launching a WebView on incognito mode, to allow for things that would need epheremal instances, such as webview sign-ins.


Hyphrio:
Currently I only implemented this on Windows and WebKitGTK, I don't have a Mac or an iPhone so I can't do WKWebView that would be guaranteed to work and my computer is too slow to implement Android.

Hyphrio:
Android would be unsupported as incognito mode would have to be called on the destructors and the non-destructor API is depreciated.

Only thing now is I hope MacOS functionality could be verified as working as I can't test it.

wusyong:
@Hyphrio It's okay to leave Android unsupported. Thank you for the work! I'll test them all once I get back.

Btw I can't figure out how your example works. Can you elaborate?

Hyphrio:
Uh the example is unfortunately still broken (on Windows) because of read-only access on the HTML file, I'll probably just spin up a temp HTTP server to work around that but that example just sets a cookie then when you close the window and reopen it it should attempt to load that cookie, but since it's incognito it will appear as nothing

~EDIT: Example works on Linux~ Nevermind it doesn't

Hyphrio:
WebKitGTK impl is currently broken so I converted this back to draft

Hyphrio:
Hopefully Linux is fixed now (is_ephemeral was read-only, had to modify both WebContext and WebContextImpl to add a new function named new_ephemeral)

wusyong:
The example still doesn't work for me, even in custom protocol. I prefer to use this since this is closer to our real usage instead of a temp HTTP server.
Does anyone have any idea if I can just test it in devtools?

wusyong:
@Hyphrio Hey the example doesn't work on Linux because of webkit2gtk's problem. I think macOS's wkwebview also have some problem about persisting cookies. Do you have any other way to test this?

pewsheen:
Hello @Hyphrio,  I can help test if the Mac fix works.

So I just ran the `incognito` example, the solution with data store configuration works pretty well, but a few things need fixing.

The data store can be retrieved directly from the `WKWebsiteDataStore` class:
```rust
let data_store: id = msg_send![class!(WKWebsiteDataStore), defaultDataStore];
let data_store: id = msg_send![class!(WKWebsiteDataStore), nonPersistentDataStore];
```

And the data store can be set to `WKWebViewConfiguration` by calling `setWebsiteDataStore` from the configuration instance:
```rust
let () = msg_send![config, setWebsiteDataStore: data_store];
```

Hyphrio:
Alright @pewsheen thank you for the insight about WKWebView!

src/webview/wkwebview/mod.rs
```diff
@@ -239,6 +239,15 @@ impl InnerWebView {
       // Config and custom protocol
       let config: id = msg_send![class!(WKWebViewConfiguration), new];
       let mut protocol_ptrs = Vec::new();
+
+      // Incognito mode
+      
+      let data_store = if (attributes.as_incognito == false) {
+        msg_send![class!(WKWebsiteDataStore), defaultDataStore];
+      } else {
+        msg_send![class!(WKWebsiteDataStore), nonPersistentDataStore];
```
pewsheen: Hmmm, it looks like the semicolon should be removed to assign `defaultDataStore` and `nonPersistentDataStore` to `data_store`.

Hyphrio: Well that's definitely an oversight on my side. Fixed in new commit (hopefully).

wusyong:
I want to make a note on Linux side. So the cookie is persist if we specify a web context even with incognito is enabled.
So I think it will have to skip creating webcontext if we want to use incognito mode.

I'll test macOS and windows later.

Edit: Windows seems great. So it's only macOS left.

wusyong (approved):
I'm happy with how this turns out. Thank you @Hyphrio!

PR #943, opened 2023-04-26
fix(macos): remove `webview` ivar in `WryWebView`

### LINKED ISSUE #942: Remove webview ivar in WryWebview class
In #900, we define a ivar `webview` in Wrywebview.
This is unnecessary since we could just call `this` when handling its callback.
This also break ios build during runtime.

pewsheen:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #942 

### Other information

`webview` ivar on set on macos, thus ios will panic on runtime when building innerwebview.

Synthesize mouse event is good after replacing `webview` with `this`, also fixing ios runtime panic.

### DIFF
--- a/.changes/remove-webview-ivar.md
+++ b/.changes/remove-webview-ivar.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Remove redundant webview ivar in WryWebView.
--- a/.changes/remove-webview-ivar.md
+++ b/.changes/remove-webview-ivar.md
@@ -2,4 +2,4 @@
 "wry": patch
 ---
 
-Remove redundant webview ivar in WryWebView.
+On iOS, fix panic at runtime due to setting webview ivar.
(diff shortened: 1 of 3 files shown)


.changes/remove-webview-ivar.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": patch
+---
+
+Remove redundant webview ivar in WryWebView.
```
amrbashir: I think this should instead explain the issue to solve rather than what the actual fix is.
```suggestion
On iOS, fix panic at runtime due to setting webview ivar.
```

amrbashir (changes requested):

amrbashir (approved):

PR #953, opened 2023-05-08
fix: fix all clippy warnings and enable clippy on CI

rhysd:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### Other information

This PR fixes all warnings reported by clippy on my local machine. I confirmed all examples fixed by this PR on my local machine manually.

### DIFF
--- a/.github/workflows/fmt.yml
+++ b/.github/workflows/fmt.yml
@@ -14,10 +14,11 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-            profile: minimal
-            toolchain: stable
-            override: true
-            components: rustfmt,clippy
+          profile: minimal
+          toolchain: stable
+          override: true
+          components: rustfmt,clippy
+      - uses: Swatinem/rust-cache@v2
       - name: fmt
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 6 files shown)


rhysd:
Clippy passed on CI. https://github.com/tauri-apps/wry/actions/runs/4915685028/jobs/8778465188?pr=953

FabianLars:
Thanks for contributing!

i think we should add (back) a caching step to that workflow now (basically just a `- uses: Swatinem/rust-cache@v2` step)

rhysd:
Thanks for the point. I added the action to the workflow at 6a90182.

FabianLars (approved):
i hope clippy is sure about the `drop()` change 😅

rhysd:
[`drop_in_place`](https://doc.rust-lang.org/std/ptr/fn.drop_in_place.html) returns `()`. So

```rust
drop(self.download_delegate.drop_in_place());
```

is exactly the same as

```rust
self.download_delegate.drop_in_place();
drop(());
```

And `drop(())` does nothing. We can safely remove the `drop()` call.

PR #966, opened 2023-06-05
refactor: consistent `with_html` across platforms

amrbashir:

- [x] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/.changes/android-html.md
+++ b/.changes/android-html.md
@@ -1,0 +1,5 @@
+---
+"wry": "patch"
+---
+
+Add support for `WebViewBuilder::with_html` and `WebViewAttributes.html` on Android.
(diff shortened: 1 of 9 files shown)


wusyong:
Why do we choose null instead of `http://localhost`? I feel like null origin is more restricted.

amrbashir:
1. consistency across platforms, Windows already behaves this way.
2. because it doesn't make sense to have an origin when serving a simple HTML string like using `data:text/html,`, instead, they should be using a custom protocol or a web server

wusyong (approved):

PR #969, opened 2023-06-09
feat: add request body on Linux

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### Other information
You can test with `form` example.
Don't merge this yet! There are a lot of new deprecated methods in 2.40. We should fix them before merging.

### DIFF
--- a/src/webview/webkitgtk/web_context.rs
+++ b/src/webview/webkitgtk/web_context.rs
@@ -5,7 +5,9 @@
 //! Unix platform extensions for [`WebContext`](super::WebContext).
 
 use crate::{webview::web_context::WebContextData, Error};
-use glib::FileError;
+use gdk::prelude::{InputStreamExt, InputStreamExtManual};
+use gio::Cancellable;
+use glib::{Bytes, FileError};
 use http::{header::CONTENT_TYPE, Request, Response};
 use std::{
   borrow::Cow,
@@ -301,11 +303,13 @@ where
       // FIXME: Read the body (forms post)
       #[allow(unused_mut)]
       let mut http_request = Request::builder().uri(uri).method("GET");
+      let body;
       #[cfg(feature = "linux-headers")]
       {
         use http::{header::HeaderName, HeaderValue};
 
-        if let Some(mut headers) = request.http_headers() {
+        // Set request http headers
+        if let Some(headers) = request.http_headers() {
           if let Some(map) = http_request.headers_mut() {
             headers.foreach(move |k, v| {
               if let Ok(name) = HeaderName::from_bytes(k.as_bytes()) {
@@ -317,11 +321,39 @@ where
           }
         }
 
+        // Set request http method
         if let Some(method) = request.http_method() {
           http_request = http_request.method(method.as_str());
         }
+
+        // Set request http body
+        let cancellable: Option<&Cancellable> = None;
+        body = request
+          .http_body()
+          .map(|s| {
+            const BUFFER_LEN: usize = 1024;
+            let mut result = Vec::new();
+            let mut buffer = vec![0; BUFFER_LEN];
+            while let Ok(count) = s.read(&mut buffer[..], cancellable) {
+              if count == BUFFER_LEN {
+                result.append(&mut buffer);
+                buffer.resize(BUFFER_LEN, 0);
+              } else {
+                buffer.truncate(count);
+                result.append(&mut buffer);
+                break;
+              }
+            }
+            result
+          })
+          .unwrap_or(Vec::new());
       }
-      let http_request = match http_request.body(Vec::new()) {
+      #[cfg(not(feature = "linux-headers"))]
+      {
+        body = Vec::new();
+      }
+
+      let http_request = match http_request.body(body) {
         Ok(req) => req,
         Err(_) => {
           request.finish_error(&mut glib::Error::new(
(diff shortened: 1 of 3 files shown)


src/webview/webkitgtk/web_context.rs
```diff
@@ -317,11 +321,39 @@ where
           }
         }
 
+        // Set request http method
         if let Some(method) = request.http_method() {
           http_request = http_request.method(method.as_str());
         }
+
+        // Set request http body
```
amrbashir, line 331: should probably be behind another feature flag `linux-body` since headers support and body support require two different webkit2gtk versions

amrbashir (dismissed):

src/webview/webkitgtk/web_context.rs
wusyong: I'm considering removing these flags and checking the version at runtime instead.
We encountered so many webkitgtk version difficulties. I think we should simplify the feature flag for users.
wdyt? cc @lucasfernog

amrbashir: sounds good

lucasfernog: would be amazing

FabianLars: I agree it would be super useful, as long as we can still compile the app against older webkitgtk version. I think we will need it for the IPC anyway if we really end up implementing version specific implementations like Jonas talked about.

lucasfernog: Right, last time we had to add feature flags for the headers because we couldn't compile against older webkit2gtk due to linking issues I think. @wusyong

lucasfernog: Like this is problematic https://github.com/tauri-apps/wry/actions/runs/5222821122/jobs/9428823012?pr=969#step:10:235

wusyong: damn. I'll add linux-body flag then.

wusyong: But I think we could remove linux-headers flag right? Since it's in 2.36. And we default to 2.38 now.

FabianLars: In this specific case we may not need a feature flag. With the switch to webkitgtk 4.1 we raised our minium version to 2.38 basically measured by ubuntu 22.04, the first ubuntu release to include webkitgtk 4.1 - it received the update to 2.40 too so imho we _could_ require that version if it makes our lives easier.

For what it's worth, the first debian release to include 4.1 is bookworm, released 2(?) days ago which came with 2.40 out of the box.

fedora 37 also got the 2.40 update and idk what other non-rolling distros may matter.

lucasfernog: We can enable everything right now and if we see a problem, mark it as feature flags enabled by default that can be disabled by users. We can still use this final alpha / beta time to check it. @wusyong

lucasfernog: If you prefer we can keep the feature flags here and leave this to the tauri side.

lucasfernog:
Btw I tested this and it works well, nice one!

src/webview/webkitgtk/web_context.rs
lucasfernog: @FabianLars actually only ubuntu 23.04 has the 2.40 update it seems. https://packages.ubuntu.com/search?keywords=webkit2gtk

wusyong: I'll keep the decision of replacing header flag with body flag. This should be compatible with tauri v2 version.

FabianLars: oh damn. pro tip: don't try to read these small letters when you just woke up after just 3 hours of sleep.

But yeah, that's _really_ bad and means you can't build tauri apps with the new ipc using github's default runners 😫 
Edit: And potentially worse, .deb bundles won't work for endusers on LTS either

lucasfernog: I can do something crazy on the Tauri side. Run the current IPC on Linux, and if the response is a blob, send a message to JS telling "hey, use a fetch for custom protocol url `ipc://localhost/$requestId` to get the blob response". At least reading a binary file wouldn't take a minute.

FabianLars: yeah, that'd be really cool! Or tbh i'd explore it for non-blobs too, maybe check if it's above 1kb or something (if not blob) because postMessage isn't exactly fast for large strings either.

lucasfernog (approved):

src/webview/webkitgtk/web_context.rs
lucasfernog: > yeah, that'd be really cool! Or tbh i'd explore it for non-blobs too, maybe check if it's above 1kb or something (if not blob) because postMessage isn't exactly fast for large strings either.

@FabianLars
https://github.com/tauri-apps/tauri/pull/7170/files#diff-cfab6e0634dd785c7949c059698552e4e8f6139ccbcdb29aeb6b4c29d57805b7R1738 
https://github.com/tauri-apps/tauri/pull/7170/files#diff-cfab6e0634dd785c7949c059698552e4e8f6139ccbcdb29aeb6b4c29d57805b7R1748
https://github.com/tauri-apps/tauri/pull/7170/files#diff-f5b23dca1c1951b8661b16b558998059ed8f2029ec6806392dcd430ddec82f02R74

for now i'm just using the protocol for any object/array/raw data. I could check the array and object length, but it's hard to determine the proper threshold.

PR #970, opened 2023-06-10
feat(android): read request body on Android

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

### Checklist

### Other information

The Android's [WebResourceRequest](https://developer.android.com/reference/android/webkit/WebResourceRequest) class does not have an option to read the request body. I had to research and find [request_data_webviewclient](https://github.com/KonstantinSchubert/request_data_webviewclient) and [Android-Request-Inspector-WebView](https://github.com/acsbendi/Android-Request-Inspector-WebView) to find a way to implement it.

I've decided to only support string because byte array performance is bad, see https://stackoverflow.com/a/45506857

### DIFF
--- a/.changes/android-custom-protocol-req-body.md
+++ b/.changes/android-custom-protocol-req-body.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Load the custom protocol request body when it is a string on Android.
--- a/src/webview/android/binding.rs
+++ b/src/webview/android/binding.rs
@@ -19,7 +19,7 @@ use super::{
   WITH_ASSET_LOADER,
 };
 
-fn handle_request(env: JNIEnv, request: JObject) -> Result<jobject, JniError> {
+fn handle_request(env: JNIEnv, request: JObject, body: JString) -> Result<jobject, JniError> {
   let mut request_builder = Request::builder();
 
   let uri = env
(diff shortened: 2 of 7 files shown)


lucasfernog:
This uses a hack where we change the `window.fetch` signature to send the body to the Android layer before the request is performed. If it's too hacky to be merged, I can use the existing Tauri IPC on Android (I'm already doing that for Linux so it wouldn't add extra code to maintain).

wusyong:
I had a discussion with Lucas yesterday and I have to say it's a bit difficult for me to accept this PR.
However, the response body is still flexible instead of string only. So it's still a bit improvement in some ways.
I would like to see some more feedbacks and see if we can find even better solution for this.
cc @amrbashir @JonasKruckenberg

lucasfernog:
For now I'll just forget this PR exists and use the Linux IPC on Android. It basically uses the existing `Ipc.kt` JavaScript interface, but uses the custom protocol to send big responses. It's not bad at all, just a little extra overhead.

But I have to agree this workaround here is ugly.

lucasfernog:
If we close this PR, we should document we cannot read the request body on Android. And maybe see if there's a way to request this feature.

amrbashir:
yeah it feels hacky and only supports strings, not worth IMO.

lucasfernog:
I'm actually happy with the postMessage request -> customProtocol response IPC performance. Let's close this one.

wusyong:
So I asked Lucas if we could utilize `WebResourceRequest`. Basically, we move the body to headers on JS side, and put it back to request body during `handleRequest`.
While it's still `Map<String, String>`, it could be more straightforward than interceptor workaround.
Lucas tested it can work on small headers size like 90 bytes, but not on large size.
With this consideration, we'll stick to postMessage on Android for now.

PR #973, opened 2023-06-11
feat(android): support navigation handler (#972)

### LINKED ISSUE #972: [feat] support navigation handler on android
I would like to be able to apply a navigation handler to android clients.

Add support for setting the navigation handler on android.

None, this is a feature parity request.

- [x] Yes
- [ ] No

**Additional context**
This should be pretty straight forward.

jhutchins:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #972 

### DIFF
--- a/.cargo/android-navigation-handler.md
+++ b/.cargo/android-navigation-handler.md
@@ -1,0 +1,5 @@
+---
+"wry": "patch"
+---
+
+Add support for `WebViewBuilder::with_navigation_handler` on Android.
--- a/src/webview/android/binding.rs
+++ b/src/webview/android/binding.rs
@@ -16,7 +16,7 @@ use tao::platform::android::ndk_glue::jni::{
 
 use super::{
   create_headers_map, ASSET_LOADER_DOMAIN, IPC, REQUEST_HANDLER, TITLE_CHANGE_HANDLER,
-  WITH_ASSET_LOADER,
+  URL_LOADING_OVERRIDE, WITH_ASSET_LOADER,
 };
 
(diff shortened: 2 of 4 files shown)


lucasfernog:
I can't test this right now but the code looks amazing, nice work!

lucasfernog (approved):
Works well. The only problem I see is navigations triggered by [loadUrl](https://developer.android.com/reference/android/webkit/WebView#loadUrl(java.lang.String)) does not trigger the [shouldOverrideUrlLoading](https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest)) callback. Seems like a limitation on Android itself, only JavaScript navigations trigger it. Of course we could change `loadUrl(url)` calls with `evaluateJavascript("window.location.href = `$url`", null)`, but that does not work for the `loadUrl(url, additionalHttpHeaders)` overload.

jhutchins:
@lucasfernog is there currently a mechanism to call `loadUrl`? Regardless, if you want that I can absolutely make that work, wouldn't even be difficult.

lucasfernog:
Right we can implement that manually. I think it's valuable for cross platform compatibility. What do you think @wusyong ?

lucasfernog:
And yeah we're overriding the loadUrl functions on the RustWebview Kotlin class.

PR #974, opened 2023-06-17
feat(android): apply navigation handler to loadUrl calls

jhutchins:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #___

### DIFF
--- a/.cargo/android-navigation-handler-improvements.md
+++ b/.cargo/android-navigation-handler-improvements.md
@@ -1,0 +1,5 @@
+---
+"wry": "patch"
+---
+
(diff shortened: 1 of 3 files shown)


jhutchins:
I waffled on whether to make external function public on the WebViewClient and have the WebView call it, but ultimately it just seemed more direct to add an external function to the WebView that binds to the same rust function.

lucasfernog:
That's how I would do it too :)

amrbashir (approved):

PR #978, opened 2023-06-21
fix(windows): ignore resize events when minimized

ahodesuka:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### Other information

On Windows when a window is minimized it is set to a minimum size, this size change event and subsequent resize of the webview causes events within the webview contents such as ResizeObervers to fire.

The resizing of the window when minimizing dates back to when Windows did not have a taskbar, and is still used today for cases when the taskbar is unavailable or for MDI (windows within windows).  A quick explanation can be found [here](https://devblogs.microsoft.com/oldnewthing/20050210-00/?p=36483).

Neither Chrome or Firefox fire off ResizeObserver or DOM resize events when they are minimized on Windows.

### DIFF
--- a/src/webview/webview2/mod.rs
+++ b/src/webview/webview2/mod.rs
@@ -743,15 +743,17 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
     ) -> LRESULT {
       match msg {
         win32wm::WM_SIZE => {
-          let controller = dwrefdata as *mut ICoreWebView2Controller;
-          let mut client_rect = RECT::default();
-          win32wm::GetClientRect(hwnd, &mut client_rect);
-          let _ = (*controller).SetBounds(RECT {
-            left: 0,
-            top: 0,
-            right: client_rect.right - client_rect.left,
-            bottom: client_rect.bottom - client_rect.top,
-          });
+          if wparam.0 != win32wm::SIZE_MINIMIZED as usize {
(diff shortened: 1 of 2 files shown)


src/webview/webview2/mod.rs
```diff
@@ -743,15 +743,17 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
...
-          let controller = dwrefdata as *mut ICoreWebView2Controller;
-          let mut client_rect = RECT::default();
-          win32wm::GetClientRect(hwnd, &mut client_rect);
-          let _ = (*controller).SetBounds(RECT {
-            left: 0,
-            top: 0,
-            right: client_rect.right - client_rect.left,
-            bottom: client_rect.bottom - client_rect.top,
-          });
+          if wparam != WPARAM(win32wm::SIZE_MINIMIZED as _) {
```
amrbashir: Avoids creating new instance of `WPARAM`
```suggestion
          if wparam.0 != win32wm::SIZE_MINIMIZED as _ {
```

amrbashir (changes requested):
Thanks for your contribution

amrbashir (approved):

PR #980, opened 2023-06-25
feat: on page load handler

jhutchins:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/src/webview/android/binding.rs
+++ b/src/webview/android/binding.rs
@@ -208,7 +208,7 @@ pub unsafe fn assetLoaderDomain(env: JNIEnv, _: JClass) -> jstring {
 }
 
 #[allow(non_snake_case)]
-pub unsafe fn onPageNavigating(env: JNIEnv, _: JClass, url: JString) {
+pub unsafe fn onNavigationStarted(env: JNIEnv, _: JClass, url: JString) {
   match env.get_string(url) {
     Ok(url) => {
       let url = url.to_string_lossy().to_string();
--- a/src/webview/android/binding.rs
+++ b/src/webview/android/binding.rs
@@ -15,8 +15,8 @@ use tao::platform::android::ndk_glue::jni::{
 };
 
 use super::{
-  create_headers_map, ASSET_LOADER_DOMAIN, IPC, REQUEST_HANDLER, TITLE_CHANGE_HANDLER,
-  URL_LOADING_OVERRIDE, WITH_ASSET_LOADER,
+  create_headers_map, ASSET_LOADER_DOMAIN, IPC, ON_LOAD_HANDLER, REQUEST_HANDLER,
+  TITLE_CHANGE_HANDLER, URL_LOADING_OVERRIDE, WITH_ASSET_LOADER,
 };
 
 fn handle_request(env: JNIEnv, request: JObject) -> Result<jobject, JniError> {
@@ -203,5 +203,18 @@ pub unsafe fn assetLoaderDomain(env: JNIEnv, _: JClass) -> jstring {
     env.new_string(domain).unwrap().into_raw()
   } else {
     env.new_string("wry.assets").unwrap().into_raw()
+  }
+}
+
+#[allow(non_snake_case)]
+pub unsafe fn onPageLoad(env: JNIEnv, _: JClass, url: JString) {
+  match env.get_string(url) {
+    Ok(url) => {
+      let url = url.to_string_lossy().to_string();
+      if let Some(h) = ON_LOAD_HANDLER.get() {
+        (h.0)(&h.1, url)
+      }
+    }
+    Err(e) => log::warn!("Failed to parse JString: {}", e),
   }
 }
--- a/src/webview/android/binding.rs
+++ b/src/webview/android/binding.rs
@@ -212,7 +212,7 @@ pub unsafe fn onPageLoad(env: JNIEnv, _: JClass, url: JString) {
     Ok(url) => {
       let url = url.to_string_lossy().to_string();
       if let Some(h) = ON_LOAD_HANDLER.get() {
-        (h.0)(&h.1, url)
+        (h.0)(&h.1, Url::parse(url).unwrap())
       }
     }
     Err(e) => log::warn!("Failed to parse JString: {}", e),
--- a/src/webview/android/binding.rs
+++ b/src/webview/android/binding.rs
@@ -13,11 +13,11 @@ use tao::platform::android::ndk_glue::jni::{
   sys::jobject,
   JNIEnv,
 };
-use url::Url;
 
 use super::{
-  create_headers_map, ASSET_LOADER_DOMAIN, IPC, ON_LOAD_HANDLER, REQUEST_HANDLER,
-  TITLE_CHANGE_HANDLER, URL_LOADING_OVERRIDE, WITH_ASSET_LOADER,
+  create_headers_map, ASSET_LOADER_DOMAIN, IPC, ON_LOADED_HANDLER, ON_LOADING_HANDLER,
+  ON_NAVIGATING_HANDLER, REQUEST_HANDLER, TITLE_CHANGE_HANDLER, URL_LOADING_OVERRIDE,
+  WITH_ASSET_LOADER,
 };
 
 fn handle_request(env: JNIEnv, request: JObject) -> Result<jobject, JniError> {
@@ -208,12 +208,38 @@ pub unsafe fn assetLoaderDomain(env: JNIEnv, _: JClass) -> jstring {
 }
 
 #[allow(non_snake_case)]
-pub unsafe fn onPageLoad(env: JNIEnv, _: JClass, url: JString) {
+pub unsafe fn onPageNavigating(env: JNIEnv, _: JClass, url: JString) {
   match env.get_string(url) {
     Ok(url) => {
       let url = url.to_string_lossy().to_string();
-      if let Some(h) = ON_LOAD_HANDLER.get() {
-        (h.0)(&h.1, Url::parse(&url).unwrap())
+      if let Some(h) = ON_NAVIGATING_HANDLER.get() {
+        (h.0)(url)
+      }
+    }
+    Err(e) => log::warn!("Failed to parse JString: {}", e),
+  }
+}
+
+#[allow(non_snake_case)]
+pub unsafe fn onPageLoading(env: JNIEnv, _: JClass, url: JString) {
+  match env.get_string(url) {
+    Ok(url) => {
+      let url = url.to_string_lossy().to_string();
+      if let Some(h) = ON_LOADING_HANDLER.get() {
+        (h.0)(url)
+      }
+    }
+    Err(e) => log::warn!("Failed to parse JString: {}", e),
+  }
+}
+
+#[allow(non_snake_case)]
+pub unsafe fn onPageLoaded(env: JNIEnv, _: JClass, url: JString) {
+  match env.get_string(url) {
+    Ok(url) => {
+      let url = url.to_string_lossy().to_string();
+      if let Some(h) = ON_LOADED_HANDLER.get() {
+        (h.0)(url)
       }
     }
     Err(e) => log::warn!("Failed to parse JString: {}", e),
--- a/src/webview/android/binding.rs
+++ b/src/webview/android/binding.rs
@@ -15,9 +15,11 @@ use tao::platform::android::ndk_glue::jni::{
 };
 
 use super::{
-  create_headers_map, ASSET_LOADER_DOMAIN, IPC, ON_LOADED_HANDLER, ON_LOADING_HANDLER,
-  REQUEST_HANDLER, TITLE_CHANGE_HANDLER, URL_LOADING_OVERRIDE, WITH_ASSET_LOADER,
+  create_headers_map, ASSET_LOADER_DOMAIN, IPC, ON_LOAD_HANDLER, REQUEST_HANDLER,
+  TITLE_CHANGE_HANDLER, URL_LOADING_OVERRIDE, WITH_ASSET_LOADER,
 };
+
+use crate::webview::PageLoadEvent;
 
 fn handle_request(env: JNIEnv, request: JObject) -> Result<jobject, JniError> {
   let mut request_builder = Request::builder();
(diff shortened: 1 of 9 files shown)


jhutchins:
I haven't added the change files yet and I need to test the Mac implementation, which I'm the most uncertain about, but thought I'd put this out there for early feedback.

src/webview/wkwebview/mod.rs
```diff
@@ -1036,6 +1043,22 @@ r#"Object.defineProperty(window, 'ipc', {
   }
 }
 
+pub fn url(webview: *mut Object) -> Url {
```
amrbashir:
```suggestion
pub fn url_from_id(webview: id) -> Url {
```

src/webview/mod.rs
```diff
@@ -232,6 +232,8 @@ pub struct WebViewAttributes {
 
   /// Whether all media can be played without user interaction.
   pub autoplay: bool,
+
+  pub on_load_handler: Option<Box<dyn Fn(&Window, Url)>>,
```
amrbashir: Avoid using `Url` for consistency with other handlers
```suggestion
  pub on_page_load_handler: Option<Box<dyn Fn(&Window, String)>>,
```

src/webview/mod.rs
```diff
@@ -629,6 +632,14 @@ impl<'a> WebViewBuilder<'a> {
     self
   }
 
+  /// Set a handler to process changes in page location.
+  ///
+  /// The handler will be called when the webview begins loading the content for the new page.
+  pub fn with_on_load_handler(mut self, handler: impl Fn(&Window, Url) + 'static) -> Self {
```
amrbashir: ditto
```suggestion
  pub fn with_on_page_load_handler(mut self, handler: impl Fn(&Window, String) + 'static) -> Self {
```

src/webview/webview2/mod.rs
```diff
@@ -311,6 +311,23 @@ impl InnerWebView {
       }
     }
 
+    if let Some(on_load_handler) = attributes.on_load_handler {
+      let window_c = window.clone();
+      unsafe {
+        webview
+          .add_ContentLoading(
```
amrbashir, line 320: shouldn't we use [`DOMContentLoaded`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.domcontentloaded?view=webview2-dotnet-1.0.1823.32) instead? or [`NavigationCompleted`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.navigationcompleted?view=webview2-dotnet-1.0.1823.32)

amrbashir (changes requested):

src/webview/mod.rs
jhutchins: I'm not sure how this would make it more consistent. The APIs that take a url as a parameter do so as a `&str`, but everywhere that a url is returned or stored it is done so as a `url::Url`, given that this is data been returned from wry to the client of the API it seems like `url::Url` would be the most consistent choice. None of the public interactions with urls use `String` that I can find.

src/webview/webview2/mod.rs
jhutchins: It depends on how we want to define on_load, [ContentLoading](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.contentloading?view=webview2-dotnet-1.0.1823.32) seemed to most closely match the APIs provided by the other platforms. I would lean towards being as consistent as possible across the platforms.

 This probably does slightly change the definition of on_load as it was being used in tauri triggering an event via loading javascript into the page and that probably does most closely track with DOMContentLoaded. This new definition would be slightly earlier in the load process. on_load in tauri now with it being triggered by injected javascript is something like server data has been retrieved, parsed and started loaded, but might not yet be visible. It is currently possible to sometimes navigate to a different page in an on_load callback before any of the page is displayed, if can always be done before it's fully loaded. I'm not sure how closely you want to track to the previous behavior.
 
All of the target platform have APIs that allow callbacks on starting to navigate, starting load (which is what this targets) and load complete. If wanted I can additionally add handlers for all three events.

src/webview/mod.rs
amrbashir: all of these handlers take the url as a String https://github.com/tauri-apps/wry/blob/4ab2a45a6a7cc1233c00c008c8808606edd4cbf4/src/webview/mod.rs#L159-L188

src/webview/wkwebview/mod.rs
amrbashir: or actually call it `url_from_webview` because not all `id`s are webviews

src/webview/webview2/mod.rs
amrbashir: Yeah for me personally, I think `on_page_load` would be used to interact with the HTML in the app so I think it should be firedd when one can evaluate a script and access the DOM. cc @lucasfernog, what do you think this API should do?

src/webview/mod.rs
jhutchins: I guess all callbacks do. I'll make that change.

src/webview/webview2/mod.rs
FabianLars: imo handling all 3 events makes the most sense, to me navigation start and load complete are even more important than load start i think. And with that we could name the 3 apis accordingly (or have one api via enums or whatever)

Also fwiw on Windows, NavigationCompleted fires even later than DOMContentLoaded which is what many people expected from tauri's on_page_load

jhutchins: I can add all three. Something like on_page_navigation, on_page_loading, on_page_loaded (suggestions on those name would be appreciated). Then should we change on_page_load from tauri to line up with final state?

src/webview/mod.rs
jhutchins: Should I take out the &Window parameter also? None of the handlers you pointed out take the Window as a parameter and I ended up not being able to use it from the tauri side like I had anticipated.

amrbashir: Sure, we can add it later if needed

src/webview/webview2/mod.rs
amrbashir: sounds fine me to me but lets keep tauri decisions for later

jhutchins:
I've added all three handlers and I believe addressed all concerns and comments.

src/webview/mod.rs
```diff
@@ -232,6 +232,15 @@ pub struct WebViewAttributes {
...
   pub autoplay: bool,
+
+  /// Set a handler closer to process chages to navigation.
+  pub on_page_navigating_handler: Option<Box<dyn Fn(String)>>,
+
+  /// Set a handler closer to process to start of loading a new page.
+  pub on_page_loading_handler: Option<Box<dyn Fn(String)>>,
+
+  /// Set a handler closer to process when a new page has been loaded.
+  pub on_page_loaded_handler: Option<Box<dyn Fn(String)>>,
```
```suggestion
  /// Set a handler closure to process changes to navigation.
  pub on_page_navigating_handler: Option<Box<dyn Fn(String)>>,

  /// Set a handler closure to process to start of loading a new page.
  pub on_page_loading_handler: Option<Box<dyn Fn(String)>>,

  /// Set a handler closure to process when a new page has been loaded.
  pub on_page_loaded_handler: Option<Box<dyn Fn(String)>>,
```

src/webview/mod.rs
```diff
@@ -232,6 +232,15 @@ pub struct WebViewAttributes {
 
   /// Whether all media can be played without user interaction.
   pub autoplay: bool,
+
+  /// Set a handler closer to process chages to navigation.
+  pub on_page_navigating_handler: Option<Box<dyn Fn(String)>>,
```
amrbashir: I feel like this clashes with `navigation_handler`, maybe we should change it to `on_page_load_start`

amrbashir (changes requested):

src/webview/mod.rs
jhutchins: I actually started there, but then backed it out because load start seemed misleading because this triggers before even a response has been returned. I agree that the name as is has high potential for confusion and should be changed.

Would on_navigation_start, on_location_change or on_page_change be better names?

amrbashir: Maybe `on_navigation_started_handler`

amrbashir: And then explain that it is fired after `navigation_handler`

src/webview/android/binding.rs
```diff
@@ -205,3 +206,42 @@ pub unsafe fn assetLoaderDomain(env: JNIEnv, _: JClass) -> jstring {
     env.new_string("wry.assets").unwrap().into_raw()
   }
 }
+
+#[allow(non_snake_case)]
+pub unsafe fn onPageNavigating(env: JNIEnv, _: JClass, url: JString) {
```
amrbashir:
```suggestion
pub unsafe fn onNavigationgStarted(env: JNIEnv, _: JClass, url: JString) {
```

src/webview/android/kotlin/RustWebViewClient.kt
```diff
@@ -32,6 +33,18 @@ class RustWebViewClient(context: Context): WebViewClient() {
         return shouldOverride(request.url.toString())
     }
 
+    override fun onLoadResource(view: WebView, url: String): Unit {
+        return onPageNavigating(url)
```
amrbashir:
```suggestion
        return onNavigationgStarted(url)
```

src/webview/android/kotlin/RustWebViewClient.kt
```diff
@@ -43,6 +56,9 @@ class RustWebViewClient(context: Context): WebViewClient() {
     private external fun withAssetLoader(): Boolean
     private external fun handleRequest(request: WebResourceRequest): WebResourceResponse?
     private external fun shouldOverride(url: String): Boolean
+    private external fun onPageNavigating(url: String)
```
amrbashir:
```suggestion
    private external fun onNavigationgStarted(url: String)
```

src/webview/android/mod.rs
```diff
@@ -90,6 +90,27 @@ macro_rules! android_binding {
       [JString],
       jboolean
     );
+    android_fn!(
+      $domain,
+      $package,
+      RustWebViewClient,
+      onPageNavigating,
```
amrbashir:
```suggestion
      onNavigationgStarted,
```

src/webview/mod.rs
```diff
@@ -629,6 +642,32 @@ impl<'a> WebViewBuilder<'a> {
     self
   }
 
+  /// Set a handler to process when navigation to new page begins.
+  ///
+  /// The handler will be called when the webview begins requesting context for a new URL.
+  /// This handler is called after the navigation_handler has been called to determine whether
+  /// nagivdation should be allow.
```
```suggestion
  /// This handler is called after [`Self::with_navigation_handler`].
```

src/webview/webview2/mod.rs
```diff
@@ -311,6 +311,54 @@ impl InnerWebView {
       }
     }
 
+    if let Some(on_navigation_started_handler) = attributes.on_navigation_started_handler {
+      unsafe {
+        webview
+          .add_NavigationStarting(
```
amrbashir: there is another call to `add_NavigationStarting` below, move this logic there and only run `on_navigation_started_handler` if `navigation_handler` allowed the navigation

src/webview/webview2/mod.rs
```diff
@@ -1047,3 +1089,11 @@ fn get_windows_ver() -> Option<(u32, u32, u32)> {
     None
   }
 }
+
+fn url_from_webview(webview: &ICoreWebView2) -> String {
+  let mut pwstr = PWSTR::null();
+
+  unsafe { webview.Source(&mut pwstr).unwrap() };
+
+  take_pwstr(pwstr)
```
```suggestion
  let mut pwstr = PWSTR::null();
  unsafe { webview.Source(&mut pwstr).unwrap() };
  take_pwstr(pwstr)
```

src/webview/wkwebview/mod.rs
```diff
@@ -603,6 +647,36 @@ impl InnerWebView {
...
+
+        if let Some(on_page_loaded_handler) = attributes.on_page_loaded_handler {
+          let on_page_loaded_handler = Box::into_raw(Box::new(Box::new(move || {
+            on_page_loaded_handler(url_from_webview(webview));
+          }) as Box<dyn Fn()>));
+          (*navigation_policy_handler).set_ivar(
+            "on_page_loaded_function",
+            on_page_loaded_handler as *mut _ as *mut c_void,
+          );
+        }
```
amrbashir: this is leaking resources, you need to keep a pointer to these handlers and drop them in the Drop impl of `InnerWebView`

amrbashir: This piece of code needs refactor, so I'd also suggest decoupling this logic from the `navigation_handler` logic, preferably in a module on its own, like `download.rs` module

src/webview/android/kotlin/RustWebViewClient.kt
```diff
@@ -32,6 +33,18 @@ class RustWebViewClient(context: Context): WebViewClient() {
         return shouldOverride(request.url.toString())
     }
 
+    override fun onLoadResource(view: WebView, url: String): Unit {
```
amrbashir: From the [docs](https://developer.android.com/reference/android/webkit/WebViewClient#onLoadResource(android.webkit.WebView,%20java.lang.String)): 
> Notify the host application that the WebView will load the resource specified by the given url.

Looks like this function will be called for any resource and not only page navigation

amrbashir (changes requested):

jhutchins:
I'm just going to remove the handler for navigation starting.

src/webview/wkwebview/mod.rs
jhutchins: Okay, sorry this took so long. I've done the refactor and I believe resolved the leak.

amrbashir (dismissed):

amrbashir:
LGTM but I will keep the PR open a while for @wusyong and @lucasfernog to give their approvals too

wusyong:
LGTM but there's still a clippy error.

lucasfernog:
I would consider using a single event function `with_on_page_load_handler` with an enum payload `PageLoadEvent { Started, Finished }` instead of two separate functions.

jhutchins:
@wusyong There is a clippy error, but I'm not sure how you all would like it resolved. The error is due to new lints applied to old code, not anything that I added. The issue is a structure that is a unit class for some environment and not for others so default is used to create it, but using default to build a unit struct is now a clippy violation. I think the most straight forward is to add a lint exclusion, but it's possible you want a different solution.

@lucasfernog I could make that change.

jhutchins:
@lucasfernog I made those changes to the public API.

@wusyong I opted to just just allow the clippy violation, but could do something else.

wusyong (dismissed):

lucasfernog (approved):

PR #994, opened 2023-07-27
refactor(windows): use `http` scheme by default for custom protocols and add option to change it

FabianLars:
We get pretty much daily requests/complains about https://github.com/tauri-apps/tauri/issues/3007 and since http://localhost is still a secure context and using http will now match the behavior when using custom schemes (on Linux&macOS) i think this is honestly a no-brainer at this point.

I do plan to backport this to the version used in tauri 1.x after we polished&merged this PR :)

- [x] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve https://github.com/tauri-apps/tauri/issues/3007

### DIFF
--- a/.changes/windows-http.md
+++ b/.changes/windows-http.md
@@ -2,4 +2,4 @@
 "wry": "minor"
 ---
 
-Add `WebViewBuilderExtWindows::with_http_scheme` to be able to choose between `http` and `https` for custom protocols on Windows.
+Add `WebViewBuilderExtWindows::with_http_scheme` to be able to choose between `http` and `https` for custom protocols on Windows. **Wry now uses `http` by default!**
--- a/.changes/windows-http.md
+++ b/.changes/windows-http.md
@@ -1,0 +1,5 @@
+---
+"wry": "minor"
+---
+
+Add `WebViewBuilderExtWindows::with_http_scheme` to be able to choose between `http` and `https` for custom protocols on Windows.
--- a/.changes/windows-http.md
+++ b/.changes/windows-http.md
@@ -2,4 +2,4 @@
 "wry": "minor"
 ---
 
-Add `WebViewBuilderExtWindows::with_http_scheme` to be able to choose between `http` and `https` for custom protocols on Windows. **Wry now uses `http` by default!**
+**Breaking change** Wry now defaults to `http://<scheme>.localhost/` for custom protocols on Windows.
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -700,7 +700,7 @@ pub trait WebViewBuilderExtWindows {
 
   /// Determines whether the custom protocols should use `http://<scheme>.localhost` instead of the default `https://<scheme>.localhost`.
   ///
-  /// Using a `http` scheme will allow mixed content warnings when trying to fetch `http` endpoints
+  /// Using a `http` scheme will allow mixed content when trying to fetch `http` endpoints
   /// and is therefore less secure but will match the behavior of the `<scheme>://localhost` protocols on macOS and Linux.
   ///
   /// The default value is `true`.
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -277,6 +277,7 @@ pub(crate) struct PlatformSpecificWebViewAttributes {
   additional_browser_args: Option<String>,
   browser_accelerator_keys: bool,
   theme: Option<Theme>,
+  http_scheme: bool,
 }
 #[cfg(windows)]
 impl Default for PlatformSpecificWebViewAttributes {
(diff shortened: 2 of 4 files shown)


src/webview/mod.rs
```diff
@@ -695,6 +697,14 @@ pub trait WebViewBuilderExtWindows {
   ///
   /// Defaults to [`Theme::Auto`] which will follow the OS defaults.
   fn with_theme(self, theme: Theme) -> Self;
+
+  /// Determines whether the custom protocols should use `http://<scheme>.localhost` instead of `http://<scheme>.localhost`.
```
```suggestion
  /// Determines whether the custom protocols should use `http://<scheme>.localhost` instead of the default `https://<scheme>.localhost`.
```

src/webview/mod.rs
```diff
@@ -695,6 +697,14 @@ pub trait WebViewBuilderExtWindows {
...
+  ///
+  /// Using a `http` scheme will prevent mixed content warnings when trying to fetch `http` endpoints
+  /// and is therefore less secure but will match the behavior of the `<scheme>://localhost` protocols on macOS and Linux.
```
```suggestion
  /// Using a `http` scheme will allow mixed content when trying to fetch
  /// and is therefore less secure but will match the behavior of the `<scheme>://localhost` protocols on macOS and Linux.
```

amrbashir (changes requested):
Maybe we could make this the default (and flipped around if ported to tauri 1.x ofc)? what about android too?

src/webview/mod.rs
FabianLars: Maybe "Using a `http` scheme will allow mixed content http requests" or ""Using a `http` scheme will allow mixed content when trying to send http requests" ? It reads a bit awkward cutting off at fetch

FabianLars:
> Maybe we could make this the default (and flipped around if ported to tauri 1.x ofc)?

Sounds good to me.

> what about android too?

1) idk its limitations == didn't know if it's even possible
2) idk how ios behaves -> whether it allows http and https requests like linux/macos

.changes/windows-http.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": "minor"
+---
+
+Add `WebViewBuilderExtWindows::with_http_scheme` to be able to choose between `http` and `https` for custom protocols on Windows. **Wry now uses `http` by default!**
```
amrbashir: lets split this breaking change into its own change file, just so it is more clear.

src/webview/mod.rs
```diff
@@ -695,6 +697,14 @@ pub trait WebViewBuilderExtWindows {
...
   /// Defaults to [`Theme::Auto`] which will follow the OS defaults.
   fn with_theme(self, theme: Theme) -> Self;
+
+  /// Determines whether the custom protocols should use `http://<scheme>.localhost` or `https://<scheme>.localhost`.
+  ///
+  /// Using a `http` scheme will allow mixed content when trying to fetch `http` endpoints
+  /// and is therefore less secure but will match the behavior of the `<scheme>://localhost` protocols used on macOS and Linux.
+  ///
+  /// The default value is `true`, meaning it will use `http`.
+  fn with_http_scheme(self, enabled: bool) -> Self;
```
```suggestion
  /// Determines whether the custom protocols should use `https://<scheme>.localhost` or the default `http://<scheme>.localhost`.
  ///
  /// Using a `http` scheme will allow mixed content when trying to fetch `http` endpoints
  /// and is therefore less secure but will match the behavior of the `<scheme>://localhost` protocols used on macOS and Linux.
  ///
  /// The default value is `false`.
  fn with_https_scheme(self, enabled: bool) -> Self;
```

amrbashir (changes requested):

amrbashir:
> > what about android too?
> 
> 1. idk its limitations == didn't know if it's even possible
> 2. idk how ios behaves -> whether it allows http and https requests like linux/macos

Alright, lets worry about it later since mobile is still in alpha.

src/webview/webview2/mod.rs
```diff
@@ -537,15 +537,20 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
       }
     }
 
+    let scheme = if pl_attrs.https_scheme {
+      "http"
+    } else {
+      "https"
+    };
```
```suggestion
    let scheme = if pl_attrs.https_scheme {
      "https"
    } else {
      "http"
    };
```

amrbashir (approved):

Dirreke:
Hello! I'm not famaliar about it. I want to consult that if  we should change https://github.com/tauri-apps/tauri/blob/af3268a4be69e2dee72ba867ee54bc36d12712e6/core/tauri/src/manager.rs#L515 to http too after this?

Except it, What others should I do to use tauri with a http backend? And if tauri-2.X can use it? Thanks!

FabianLars:
@Dirreke For v1 i'm working on it here: https://github.com/tauri-apps/tauri/pull/7645 (mainly waiting for a new wry release). With the release of 1.5 we'll merge v1 into v2 one last time so this change will also be available in a future v2 release.

PR #1006, opened 2023-08-15
feat: support webview proxy

### LINKED ISSUE #991: Expose WebkitGTK Proxy Feature on Linux
Currently wry does not expose or support upstream proxies for the webview on Linux.
WebkitGTK supports this out of the box since version `2.16` but it is not exposed in a uniform interface in wry.

I would like to have an exposed wry proxy configuration, where it is possible to set the webview proxy independent
of the underlying WebkitGTK method. Possibly this interface could be used for `Webview2` by adding custom parameters
to the window creation by default.
I would like to have a patch for older wry versions, so we could use this feature in Tauri `1.x` instead of just in the `2.x` releases.

The most recent supported interface is exposed in the [`WebsiteDataManager.set_network_proxy_settings`](https://webkitgtk.org/reference/webkit2gtk/stable/method.WebsiteDataManager.set_network_proxy_settings.html)
and the deprecated version is exposed via [`WebContext.set_network_proxy_settings`](https://webkitgtk.org/reference/webkit2gtk/stable/method.WebContext.set_network_proxy_settings.html
).

There are no alternatives present and using this for auditing or compliance would require a custom built version of Tauri and Wry.
The functionality is needed in corporate environments where there are upstream proxies to be honored and when auditing Tauri applications it is helpful to use common security tooling in the process.

The proxy feature is exposed on windows via the custom webview arguments (`--proxy-server 127.0.0.1:8080`) and on MacOs in a new [beta version](https://developer.apple.com/documentation/network/nwparameters/privacycontext/4156642-proxyconfigurations).

Discussed this shortly with @wusyong before opening the issue here.

pewsheen:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #991 

#### Todo:
- [x] macOS (requires 14.0+)
- [x] Linux
- [x] Windows

#### Support proxy types
- HTTP CONNECT
- SOCKSv5

We're not ready to support authentications  and custom TLS configuration for proxy at this stage

Also, the API provided by wkwebview is still in Beta, which means it requires macOS 14.0+


wusyong:
Are we able to check macos version during runtime?
So if it's beneath the version we just perform nop.

src/webview/wkwebview/proxy.rs
```diff
@@ -0,0 +1,67 @@
...
+  fn nw_proxy_config_create_relay(
+    first_hop: nw_relay_hop_t,
+    second_hop: nw_relay_hop_t,
+  ) -> nw_proxy_config_t;
+  pub fn nw_proxy_config_create_socksv5(proxy_endpoint: nw_endpoint_t) -> nw_proxy_config_t;
+  pub fn nw_proxy_config_create_http_connect(
+    proxy_endpoint: nw_endpoint_t,
+    proxy_tls_options: nw_protocol_options_t,
+  ) -> nw_proxy_config_t;
+}
```
pewsheen, line 50: @wusyong I guess even if we can detect the macOS version on runtime, it would still fail in the building phase since we are linking to the system library here.

wusyong: Let's add a `mac_proxy` feature flag and add this module behind it.
We mention this in the documentation to help users understand if they want this feature, they need to use mac14+

pewsheen:
> Are we able to check macos version during runtime? So if it's beneath the version we just perform nop.

For runtime os version check, we can use [isOperatingSystemAtLeastVersion](https://developer.apple.com/documentation/foundation/nsprocessinfo/1414876-isoperatingsystematleastversion)

src/webview/wkwebview/proxy.rs
pewsheen: ok

src/webview/mod.rs
```diff
@@ -235,6 +237,13 @@ pub struct WebViewAttributes {
 
   /// Set a handler closure to process page load events.
   pub on_page_load_handler: Option<Box<dyn Fn(PageLoadEvent, String)>>,
+
+  /// Set a proxy configuration for the webview.
+  ///
+  /// - **macOS**: Requires macOS 14.0+ and the `mac-proxy` feature flag to be enabled. Supports HTTP CONNECT and SOCKSv5 proxies.
+  /// - **Windows / Linux**: Supports HTTP CONNECT and SOCKSv5 proxies.
```
```suggestion
  /// Set a proxy configuration for the webview. Supports HTTP CONNECT and SOCKSv5 proxies
  ///
  /// - **macOS**: Requires macOS 14.0+ and the `mac-proxy` feature flag to be enabled.
```

src/webview/proxy.rs
```diff
@@ -0,0 +1,26 @@
...
+  /// Connect to proxy server via HTTP CONNECT
+  Http(ProxyEndpoint),
+  /// Connect to proxy server via SOCKSv5
+  Socks5(ProxyEndpoint),
+}
+#[derive(Debug, Clone)]
+pub struct ProxyConfig {
+  pub proxy_type: ProxyType,
+  pub proxy_connection: ProxyConnection,
+}
```
amrbashir: The `ProxyType` and `ProxyConnection` are redundant, this should be enough:
```suggestion
#[derive(Debug, Clone)]
pub struct ProxyEndpoint {
  /// Proxy server host (e.g. 192.168.0.100, localhost, example.com, etc.)
  pub host: String,
  /// Proxy server port (e.g. 1080, 3128, etc.)
  pub port: String,
}
#[derive(Debug, Clone)]
pub enum ProxyConfig {
  /// Connect to proxy server via HTTP CONNECT
  Http(ProxyEndpoint),
  /// Connect to proxy server via SOCKSv5
  Socks5(ProxyEndpoint),
}

```

src/webview/webview2/mod.rs
```diff
@@ -74,7 +81,12 @@ impl InnerWebView {
     let file_drop_handler = attributes.file_drop_handler.take();
     let file_drop_window = window.clone();
 
-    let env = Self::create_environment(&web_context, pl_attrs.clone(), attributes.autoplay)?;
+    let env = Self::create_environment(
+      &web_context,
+      pl_attrs.clone(),
+      attributes.autoplay,
+      attributes.proxy_config.clone(),
```
amrbashir: Could probably just pass a reference instead of cloning here

src/webview/webview2/mod.rs
```diff
@@ -131,11 +144,24 @@ impl InnerWebView {
...
               } else {
                 ""
+              },
+              if let Some(proxy_setting) = proxy_config {
+                match proxy_setting.proxy_connection {
+                  ProxyConnection::Http(endpoint) => {
+                    format!(" --proxy-server=http://{}:{}", endpoint.host, endpoint.port)
+                  }
+                  ProxyConnection::Socks5(endpoint) => format!(
+                    " --proxy-server=socks5://{}:{}",
```
amrbashir: we should update the documentation of `WebViewBuilderExtWindows::with_additional_browser_args` to mention this.

amrbashir (changes requested):

src/webview/webview2/mod.rs
dklassic: Actually I think we might still need a clone here, same as the reason why `pl_attrs` is cloned here originally.

Inside `create_environment`, `CreateCoreWebView2EnvironmentCompletedHandler::wait_for_async_operation` will cause a lifetime problem.

amrbashir: right, maybe we should then construct the browser args string before calling `CreateCoreWebView2EnvironmentCompletedHandler::wait_for_async_operation`, that way we don't have lifetime issues and do one string allocation instead of two. Same could probably be done for `pl_atts` if possible.

src/webview/proxy.rs
pewsheen: This should be fixed now.

src/webview/webview2/mod.rs
```diff
@@ -105,6 +111,35 @@ impl InnerWebView {
...
+        // and "smart screen" - See https://github.com/tauri-apps/tauri/issues/1345
+        format!(
+          "--disable-features=msWebOOUI,msPdfOOUI,msSmartScreenProtection{}{}",
+          if attributes.autoplay {
+            " --autoplay-policy=no-user-gesture-required"
+          } else {
+            ""
+          },
+          if let Some(proxy_setting) = &attributes.proxy_config {
+            match &proxy_setting {
```
pewsheen: Do we need another reference here?

dklassic: Right, we don't, let me remove that real quick

src/webview/mod.rs
```diff
@@ -4,6 +4,7 @@
 
 //! [`WebView`] struct and associated types.
 
+pub mod proxy;
```
amrbashir: lets hide the module and 
```suggestion
mod proxy;
```
and then export the needed types only 
```
pub use proxy::ProxyConfig;
```

amrbashir (dismissed):

amrbashir (approved):

PR #1017, opened 2023-09-05
refactor(protocol): respond with a function call instead of return value

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

### Checklist

### Other information

This enhances the custom protocol handler to allow working asynchronously instead of freezing the main thread for longer tasks. I wanted to do this for a long time but now with Tauri using custom protocols for IPC, we actually need this (my biggest issue right now is using the camera on mobile, it was freezing the app).

### DIFF
--- a/examples/custom_protocol.rs
+++ b/examples/custom_protocol.rs
@@ -28,7 +28,7 @@ fn main() -> wry::Result<()> {
 
   let _webview = WebViewBuilder::new(window)
     .unwrap()
-    .with_custom_protocol("wry".into(), move |request| {
+    .with_custom_protocol("wry".into(), move |request, api| {
       let path = request.uri().path();
       // Read the file content from file path
       let content = if path == "/" {
@@ -53,10 +53,13 @@ fn main() -> wry::Result<()> {
         unimplemented!();
       };
 
-      Response::builder()
+      let response = Response::builder()
         .header(CONTENT_TYPE, mimetype)
-        .body(content)
-        .map_err(Into::into)
+        .body(content)?;
+
+      api.respond(response);
+
+      Ok(())
     })
     // tell the webview to load the custom protocol
     .with_url("wry://localhost")?
--- a/examples/custom_protocol.rs
+++ b/examples/custom_protocol.rs
@@ -29,18 +29,17 @@ fn main() -> wry::Result<()> {
 
   let _webview = WebViewBuilder::new(window)
     .unwrap()
-    .with_asynchronous_custom_protocol(
-      "wry".into(),
-      move |request, response| match get_wry_response(request) {
-        Ok(http_response) => response.respond(http_response),
-        Err(e) => response.respond(
+    .with_asynchronous_custom_protocol("wry".into(), move |request, responder| {
+      match get_wry_response(request) {
+        Ok(http_response) => responder.respond(http_response),
+        Err(e) => responder.respond(
           http::Response::builder()
             .header(CONTENT_TYPE, "text/plain")
             .body(e.to_string().as_bytes().to_vec())
             .unwrap(),
         ),
-      },
-    )
+      }
+    })
     // tell the webview to load the custom protocol
     .with_url("wry://localhost")?
     .build()?;
--- a/examples/custom_protocol.rs
+++ b/examples/custom_protocol.rs
@@ -2,24 +2,25 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
+use std::{
+  fs::{canonicalize, read},
+  path::PathBuf,
+};
+
+use http::Request;
+use wry::{
+  application::{
+    event::{Event, StartCause, WindowEvent},
+    event_loop::{ControlFlow, EventLoop},
+    window::WindowBuilder,
+  },
+  http::{header::CONTENT_TYPE, Response},
+  webview::WebViewBuilder,
+};
+
 const PAGE1_HTML: &[u8] = include_bytes!("custom_protocol_page1.html");
 
 fn main() -> wry::Result<()> {
-  use std::{
-    fs::{canonicalize, read},
-    path::PathBuf,
-  };
-
-  use wry::{
-    application::{
-      event::{Event, StartCause, WindowEvent},
-      event_loop::{ControlFlow, EventLoop},
-      window::WindowBuilder,
-    },
-    http::{header::CONTENT_TYPE, Response},
-    webview::WebViewBuilder,
-  };
-
   let event_loop = EventLoop::new();
   let window = WindowBuilder::new()
     .with_title("Custom Protocol")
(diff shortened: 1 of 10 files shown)


amrbashir (dismissed):
LGTM! but I am wondering if we can keep the same DX as before, e.g. allow either responding early or just return the response from the closure, for example:
```rs
.with_custom_protocol("wry".into(), move |request, api| {
  // do work
  api.respond(response);
  // do morework
})
```
and 
```rs
.with_custom_protocol("wry".into(), move |request, api| {
  // do work
  Ok(response)
})
```

lucasfernog:
That would be more confusing.. you can only respond once, so if we wanna keep the return value, we'd need to make it an Option where None means you're using the api parameter (yikes).

amrbashir:
Right, that would be confusing, how about two separate APIs then? maybe `with_custom_protocol_responder`?

lucasfernog:
Good idea! Let me know what you think.

src/webview/mod.rs
```diff
@@ -60,6 +60,20 @@ use crate::application::{dpi::PhysicalSize, window::Window};
 
 use http::{Request, Response};
 
+pub struct RequestApi {
```
wusyong: Let's also add documentation about what's this and where will use it.
```suggestion
pub struct Response{
```

src/webview/mod.rs
```diff
@@ -60,6 +60,20 @@ use crate::application::{dpi::PhysicalSize, window::Window};
...
+  pub(crate) responder: Box<dyn FnOnce(Response<Cow<'static, [u8]>>)>,
+}
+
+unsafe impl Send for RequestApi {}
+unsafe impl Sync for RequestApi {}
```
wusyong: I don't think these should be implemented in wry.
If it's really required for tokio context. tauri-runtime-wry should be better place.
I believe they will have to write the wrapper again anyway.

wusyong (changes requested):

src/webview/mod.rs
lucasfernog: it's required for the use case this PR covers (spawning another thread to do the operation). If we don't do this, it's useless.

And it's indeed safe to do so (at least it's working).

wusyong: How about just impl Send? And please add documentation to justify why it's safe.

src/webview/mod.rs
```diff
@@ -135,7 +153,7 @@ pub struct WebViewAttributes {
   /// [bug]: https://bugs.webkit.org/show_bug.cgi?id=229034
   pub custom_protocols: Vec<(
     String,
-    Box<dyn Fn(&Request<Vec<u8>>) -> Result<Response<Cow<'static, [u8]>>>>,
+    Box<dyn Fn(Request<Vec<u8>>, Response) -> Result<()>>,
```
wusyong: It doesn't make sense to return `Result` here if you want to be async. `Response` should be the one to determine the `Result`.

wusyong (changes requested):

src/webview/mod.rs
lucasfernog: pushed @wusyong

wusyong: I mean the implementation should also handle this.
If it's responder dealing with it, it should decide which finish method should be called.

src/webview/mod.rs
```diff
@@ -58,7 +58,25 @@ pub use url::Url;
 use crate::application::platform::windows::WindowExtWindows;
 use crate::application::{dpi::PhysicalSize, window::Window};
 
-use http::{Request, Response};
+use http::{Request, Response as HttpResponse};
+
+/// Resolves a custom protocol [`Request`] asynchronously.
+///
+/// See [`WebViewBuilder::with_asynchronous_custom_protocol`] for more information.
+pub struct Response {
```
amrbashir: I don't like this name, it conflicts with the re-exported `http::Response`, this should be renamed to `RequestAsyncResponder`

examples/custom_protocol.rs
```diff
@@ -28,36 +29,18 @@ fn main() -> wry::Result<()> {
...
-      };
-
-      Response::builder()
-        .header(CONTENT_TYPE, mimetype)
-        .body(content)
-        .map_err(Into::into)
-    })
+    .with_asynchronous_custom_protocol(
+      "wry".into(),
+      move |request, response| match get_wry_response(request) {
```
amrbashir: Could we also change the examples and documentation to show that the second parameter is a responder and not an actual response ?
```suggestion
      move |request, responder| match get_wry_response(request) {
```

amrbashir (changes requested):

src/webview/mod.rs
lucasfernog: I like this name, what do you think @wusyong

src/webview/mod.rs
amrbashir: We don't necessarily need to add `.respond_with_err()` variant, since you can just create a response with status code between 400-599 and you should be fine. However, I do agree that this function shouldn't return a result to allow the responder to respond even if this function errored and the responder were sent somewhere else.

lucasfernog: yeah at least it's less confusing for the user where you can call respond() and return an error later and it would crash the app.

lucasfernog: about manually creating error responses: that's what I'm trying to convince @wusyong  :D

src/webview/mod.rs
wusyong: Ok I can agree on `Result` can be removed.
But I think it becomes more and more complicated lol.
How about custom protocol attribute to be: 

```
custom_protocols: Vec<String, Box<dyn Fn(CustomURLHandler)>
```

And `CustomURLHandler` has two methods:
```
fn request(&self) -> &Request<Vec<u8>
fn response(self, Response<Cow<'static, [u8]>>>)
```

This also keep `use http::{request, response}`.

lucasfernog: The downside with this approach is I'd like to call `request.into_parts()` so I don't have to clone the request body or other data, and it needs to own the Request reference.

If we expose the request field itself it would be fine, but I still don't see much difference to the existing API.

wusyong: Original `with_custom_protocol` can still keep the signature (with Result removed.)
`F: Fn(&Request<Vec<u8>>) -> Response<Cow<'static, [u8]>> + 'static`

lucasfernog: For the original one i think it's fine to keep the Result (just to not have a breaking change for wry devs).

wusyong: @lucasfernog Well, just rename it to `AsyncResponder` then.
I still think this type will need a better name or more detailed docs.
Otherwise, it's going to be confusing for users.

amrbashir: If we really don't want to confuse users, then rename it to `CustomProtocolAsyncResponder` :joy: and the docs seems sufficient enough for its use-case especially since it links back to the `with_async_custom_protocol`.

lucasfernog: whatever you guys decide

examples/custom_protocol.rs
```diff
@@ -28,35 +29,16 @@ fn main() -> wry::Result<()> {
...
-        .header(CONTENT_TYPE, mimetype)
-        .body(content)
-        .map_err(Into::into)
+    .with_asynchronous_custom_protocol("wry".into(), move |request, responder| {
+      match get_wry_response(request) {
+        Ok(http_response) => responder.respond(http_response),
+        Err(e) => responder.respond(
+          http::Response::builder()
+            .header(CONTENT_TYPE, "text/plain")
+            .body(e.to_string().as_bytes().to_vec())
```
amrbashir, line 39:
```suggestion
			.status(500)
            .body(e.to_string().as_bytes().to_vec())
```

src/webview/mod.rs
wusyong: Alright keep it this way.

src/webview/mod.rs
```diff
@@ -135,7 +153,7 @@ pub struct WebViewAttributes {
   /// [bug]: https://bugs.webkit.org/show_bug.cgi?id=229034
   pub custom_protocols: Vec<(
     String,
-    Box<dyn Fn(&Request<Vec<u8>>) -> Result<Response<Cow<'static, [u8]>>>>,
+    Box<dyn Fn(Request<Vec<u8>>, RequestAsyncResponder) -> Result<()>>,
```
wusyong:
```suggestion
    Box<dyn Fn(Request<Vec<u8>>, RequestAsyncResponder)>,
```

Since the decision is removing Result, let's change it and remove the underlying function it calls.

amrbashir: we need to add `custom_protocols_async` field here and keep the existing one as is to avoid a breaking change.

lucasfernog: let me know if it looks good now @wusyong

amrbashir (approved):

wusyong (approved):

Issue #677, opened 2022-08-31
Integrate a WebView into a raw window


FredrikNoren:
I'd like to integrate a webview into a [wgpu](https://github.com/gfx-rs/wgpu) application, but I'm not sure what the best way to do that is

**Describe the solution you'd like**
I'd like to render a webview as a sidebar in my wgpu 3d application.

**Describe alternatives you've considered**
1. Run the webview in a hidden window, snapshot it and use as a texture, and feed user input back by simulating events through eval. Would hinge on #266 or equivalent. Not sure how performant this would be, in case of animations etc.
2. Run the webview inside of the wgpu application somehow, and assign it to a part of the screen. Not sure if this is at all possible
3. Swap it around; make everything a webview and then somehow get a handle to a canvas or something which I can use as a draw target for my wgpu application. Not sure if this is possible either.

**Would you assign yourself to implement this feature?**
Possibly, but not sure even what the right solution is here yet

**Additional context**
I'm not sure if those are the only options, or if there are more. If anyone has any input or experience on this I'd love to hear.

FrankenApps:
I would say you also have to other options which seem to be more practical imho:
1. Only use tauri and compile your wgpu application to WASM, then render inside the browser using WebGL2 (maybe WebGPU if it ever becomes widely supported).
2. Run wgpu in headless mode and render to a bitmap (see wgpus' "capture" example). Then pass the bitmap from your rust backend to the frontend via RPC where you can display it (for example in a canvas). Of course if your application depends on user input you would have to always feed any input back to the renderer in which case the first option might be superior.

amrbashir:
This is something I am personally interested in (and actually related to multiple webviews per one window) but there is other important priorities so it is not going to be implemented soon. Ofc PRs are always welcome.

FredrikNoren:
@FrankenApps My comments on those two; 

1. We're unfortunately using API's that aren't available in WASM, so we can't do that currently.
2. With WGPU there's usually 1-3 frames of delay for reading a texture from the gpu to the cpu, so I think doing it that way would lead to a bit too much lag.

@amrbashir Ok cool, so you'd say that the second path (assign a webview to a section of the window) would be the way to go here? Do you have a rough idea what needs to be done to make that happen?

amrbashir:
@FredrikNoren Yup, I meant the second option.

[...]

PR #1018, opened 2023-09-06
Add winit feature flag

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, tao and winit feature flags are added.
- [ ] No

### Checklist

### Other information
I want to add winit support for a very long time. Make wry to extend it and support Windows and macOS isn't that difficult.
I'm also working on [winit gtk](https://github.com/wusyong/winit-gtk) port which should make Linux support easier in the future.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,6 +93,6 @@ jobs:
       - name: build wry with winit
         if: (
           contains(matrix.platform.target, 'windows') ||
-          contains(matrix.platform.target, 'apple') ||
+          contains(matrix.platform.target, 'apple'))
         run: cargo build --no-default-features --features winit --target ${{ matrix.platform.target }}
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,3 +89,10 @@ jobs:
           !contains(matrix.platform.target, 'android') &&
           !contains(matrix.platform.target, 'ios'))
         run: cargo test --verbose --target ${{ matrix.platform.target }}
+
+      - name: build wry with winit
+        if: (
+          contains(matrix.platform.target, 'windows') ||
+          contains(matrix.platform.target, 'apple') ||
+        run: cargo build --no-default-features --features winit --target ${{ matrix.platform.target }}
+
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -89,3 +89,10 @@ jobs:
           !contains(matrix.platform.target, 'android') &&
           !contains(matrix.platform.target, 'ios'))
         run: cargo test --verbose --target ${{ matrix.platform.target }}
+
+      - name: build wry with winit
+        if: (
+          contains(matrix.platform.target, 'windows') ||
+          contains(matrix.platform.target, 'apple') ||
```
amrbashir:
```suggestion
        if: (
          contains(matrix.platform.target, 'windows') ||
          contains(matrix.platform.target, 'apple'))
```

src/webview/webview2/mod.rs
```diff
@@ -392,6 +390,8 @@ window.addEventListener('mousemove', (e) => window.chrome.webview.postMessage('_
             let mut js = PWSTR::null();
             args.TryGetWebMessageAsString(&mut js)?;
             let js = take_pwstr(js);
+            // Winit 0.28 doesn't have manual hittest yet.
+            #[cfg(not(feature = "winit"))]
```
amrbashir: we can just port the code of hittest here, it shouldn't be a problem

amrbashir (changes requested):
This is a fine way to add winit, but if there is time, we could maybe look into https://github.com/tauri-apps/wry/issues/677

src/webview/webview2/mod.rs
wusyong: There's also `window.begin_resize_drag`. And the hit_test function will need to call several DPI[ foreign functions](https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/windows/util.rs#L318) which aren't included in windows crate.
Can you open an issue instead for future improvement?

amrbashir (dismissed):

amrbashir (dismissed):

lucasfernog (dismissed):

lucasfernog (approved):

PR #1036, opened 2023-10-16
chore(deps): upgrade gtk to 0.18

dklassic:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### Other information

This draft PR currently uses my repository of `winit-gtk` for `winit` so it is possible that CI will fail for MacOS and Windows

### DIFF
--- a/src/webview/webkitgtk/undecorated_resizing.rs
+++ b/src/webview/webkitgtk/undecorated_resizing.rs
@@ -37,10 +37,9 @@ pub fn setup(webview: &WebView) {
         }
       }
     }
-    Inhibit(false)
+    glib::Propagation::Proceed
   });
   webview.connect_button_press_event(move |webview, event| {
-    let inhibit = false;
     if event.button() == 1 {
       let (cx, cy) = event.root();
       // This one should be GtkBox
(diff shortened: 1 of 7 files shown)


src/webview/webkitgtk/undecorated_resizing.rs
```diff
@@ -64,7 +64,11 @@ pub fn setup(webview: &WebView) {
         }
       }
     }
-    Inhibit(inhibit)
+    if inhibit {
```
pewsheen: Just curious about the `inhibit` variable, it seems not been used and modified in the scope

dklassic: You're right, it is indeed just `false` all the time.

After some digging I think it is probably just a copy and paste from other `connect_button_press_event` callback with modification. I'll patch it out.

amrbashir:
`tao@0.23` is now released @dklassic you can continue with this PR

pewsheen (approved):

PR #1041, opened 2023-10-22
Refactor new method entry point to rwh

wusyong:
RFC: https://github.com/tauri-apps/rfcs/pull/12

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

### Checklist


Cargo.toml
```diff
@@ -90,11 +81,15 @@ cocoa = "0.25"
...
 [target."cfg(target_os = \"android\")".dependencies]
 crossbeam-channel = "0.5"
 html5ever = "0.26"
 kuchiki = { package = "kuchikiki", version = "0.8" }
 sha2 = "0.10"
 base64 = "0.21"
+
+[dev-dependencies]
+http-range = "0.1.5"
+tao = { git = "https://github.com/tauri-apps/tao", branch = "rwh" }
```
amrbashir:
```suggestion
tao = { git = "https://github.com/tauri-apps/tao" }
```

amrbashir:
Here is what is left to do:
- [x] Fix Documentation and add missing ones
- [x] Cleanup examples and add new examples for winit, wgpu and some other projects
- [x] Figure out how to resize the child gtk in x11
- [X] Webview created directly in the passed X11, should resize when the X11 is resized? No

wusyong:
>     * [ ]  Cleanup examples and add new examples for winit, wgpu and some other projects

I think we should reduce the amount of example to only some basic ones and move others to integration test.
Let's do this in next isses/PRs.
 
>     * [ ]  Figure out how to resize the child gtk in x11

Can Windows auto resize? On macOS, it can auto resize. But I feel like we should just tell users to handle child position and size manually themselves.

amrbashir:
> > ```
> > * [ ]  Figure out how to resize the child gtk in x11
> > ```
> 
> Can Windows auto resize? On macOS, it can auto resize. But I feel like we should just tell users to handle child position and size manually themselves.

Windows can auto resize correctly yes, but on x11 it doesn't, the problem is if we take a raw x11 handle and we initialize the webview as a child, you can call `webview.set_size` which changes the x11 child correctly but the gtk render surface stays the same

amrbashir:
Alright, I fixed the `webview.set_size` for child webviews, now remains only one-case which I am not sure if even possible to fix. When you create a webview using raw-window-handle `Webview::new(&window)`, on macOS and Linux when the passed window is resized, the webview also gets resized, but on X11, it stays the same size when was created.

amrbashir:
Alright, I have opted into leaving that behavior as is and documenting it and also changed `webview.set_size` to work only on `X11` if the webview was created that way, also modified the `winit` example to showcase this.

amrbashir:
I should note that any example that uses winit right now is broken due to a crash in winit and I am reaching out to the winit team to find a solution to this.

niklaskorz:
Transparency does not seem to work anymore with this branch (on macOS):

grafik

Using the following code:

```rs
    let event_loop = EventLoop::new()?;
    let window = WindowBuilder::new().build(&event_loop)?;

    let parent = window.window_handle()?;
    let mut builder = WindowBuilder::new()
        .with_title("child window")
        .with_decorations(false)
        .with_has_shadow(false)
        .with_transparent(true)
        .with_inner_size(LogicalSize::new(400.0f32, 400.0f32))
        //.with_position(Position::Logical(LogicalPosition::new(0.0, 0.0)))
        .with_visible(true);
    builder = unsafe { builder.with_parent_window(Some(parent.into())) };
    let child_window = builder.build(&event_loop)?;

    let builder = WebViewBuilder::new(&child_window);
    let webview = builder
        .with_transparent(true)
        .with_devtools(false)
        .with_html(
            r#"
                  <!doctype html>
                  <html style="background-color: rgba(0, 0, 0, 0);">
                    <body style="background-color: rgba(0, 0, 0, 0);">hello</body>
                    <script>
                      window.onload = function() {
                        document.body.innerText = `hello, ${navigator.userAgent}`;
                      };
                    </script>
                  </html>"#,
        )?
        .build()?;
```

The parent window here is rendered by a wgpu surface filled with a static color.

niklaskorz:
Also, the application crashes every time the webview loses focus:

```
thread 'main' panicked at 'ivar _state not found on class WryWebViewParent', /Users/niklaskorz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.4.1/src/runtime/mod.rs:1015:17
```

amrbashir:
@niklaskorz thanks for testing, this is really nice feedback.

About the transparent thing, are you using `winit` or `tao` as the windowing library? can you test with both?

Also cc @wusyong

niklaskorz:
> @niklaskorz thanks for testing, this is really nice feedback.
> 
> About the transparent thing, are you using `winit` or `tao` as the windowing library? can you test with both?
> 
> Also cc @wusyong

This was using winit, but I also tried with tao (dev-branch) right now. This code:

```rs
    use tao::{
        event::*,
        event_loop::{EventLoop, ControlFlow},
        platform::macos::WindowBuilderExtMacOS,
        window::WindowBuilder, keyboard::Key,
    };
    use wry::WebViewBuilder;

        .with_transparent(true)
        .with_decorations(false)
        .with_has_shadow(false)
        .build(&event_loop)?;
    
    let builder = WebViewBuilder::new(&window);
    let webview = builder
        .with_transparent(true)
        .with_devtools(false)
        .with_html(
            r#"
                  <!doctype html>
                  <html style="background-color: rgba(0, 0, 0, 0);">
                    <body style="background-color: rgba(0, 0, 0, 0);">hello</body>
                    <script>
                      window.onload = function() {
                        document.body.innerText = `hello, ${navigator.userAgent}`;
                      };
                    </script>
                  </html>"#,
        )?
        .build()?;
```

gives me a web view with a white background:

Bildschirmfoto 2023-10-26 um 17 56 24

On the upside, when using tao instead of winit, the app does not crash when the webview loses focus.

amrbashir:
Thanks for the feedback, @wusyong will take a look as soon as he can.

niklaskorz:
@amrbashir Sorry, I actually didn't know transparent is a feature flag, maybe this should be added to the comment of to the `with_transparent` method. Or that method should be behind the transparent flag as well. Anyhow, with `features = ["transparent"]`, it works now... Both in tao and winit:

grafik

But the crash on focus-lost (`ivar _state not found on class WryWebViewParent`) in winit still occurs.

Edit: this only happens when creating a real child window through winit. Using wry's own `WebViewBuilder::new_as_child`, which adds the webview as a child view instead of child window on macOS, does not have this issue.

Edit 2: when creating the wgpu surface and the webview on the same window (directly, neither of them as a child), the crash on focus loss also occurs, but despite that this approach works fine, I'm amazed:

grafik

amrbashir:
Thank you very much, I believe we should be able to fix that ivar crash

amrbashir:
And I will add a compile error on the transparent example to make it harder to miss the feature flag.

wusyong:
@niklaskorz You don't need to create a child window. Our `new_as_child` will handle that for you already.
All you need to do is provide position and size when creating webview.

``` 
    let event_loop = EventLoop::new()?;
    let window = WindowBuilder::new().build(&event_loop)?;

    let builder = WebViewBuilder::new_as_child(&window);
    let webview = builder
        .with_transparent(true)
        .with_size((100, 100))
        .with_devtools(false)
        .with_html(
            r#"
                  <!doctype html>
                  <html style="background-color: rgba(0, 0, 0, 0);">
                    <body style="background-color: rgba(0, 0, 0, 0);">hello</body>
                    <script>
                      window.onload = function() {
                        document.body.innerText = `hello, ${navigator.userAgent}`;
                      };
                    </script>
                  </html>"#,
        )?
        .build()?;
```

As for the crash, I think winit will try to set some state to NSView and our Wry's default behavior will try to replace the view with our own. I guess we should update the doc to suggest winit users should use `new_as_child` instead of `new` for now. @amrbashir Does this look good to you?

niklaskorz:
@wusyong On Windows, using `new_as_child` instead of creating a winit child window has the problem that transparency is only set for the webview, not the child window wry creates:

grafik


Repository: https://github.com/niklaskorz/wgpu-wry-example (webview is created here: https://github.com/niklaskorz/wgpu-wry-example/blob/82ce2923100a1cfbf8e742835c25eb44a1e476d3/src/lib.rs#L669)

Edit: if you don't mind, I can try implementing setting transparency on the win32 child window and provide you a patch

amrbashir:
Tbh I feel like we shouldn't care what crate the user is using and we should avoid replacing the view, keep the original and then add ours as a new subview.

wusyong:
@amrbashir There are a lot of trouble if we just add as subview. For example like, the cursor won't change if unfocused, menu item accelerators won't work, global accelerator won't work, and IME troubles.
We come a long way for this, and I spent tremendous of time fixing all these issues. And we endup replace the content view in order to work best with tauri. So for tauri usage, it still must replace the view.

wusyong:
> @wusyong On Windows, using `new_as_child` instead of creating a winit child window has the problem that transparency is only set for the webview, not the child window wry creates:

@amrbashir I think we should make the child window be also transparent if webview has transparent attribute on Windows and Linux then.

amrbashir:
Can we at least reuse the view? without removing its state

amrbashir:
> > @wusyong On Windows, using `new_as_child` instead of creating a winit child window has the problem that transparency is only set for the webview, not the child window wry creates:
> 
> @amrbashir I think we should make the child window be also transparent if webview has transparent attribute on Windows and Linux then.

I will take care of that

wusyong:
> Can we at least reuse the view? without removing its state

I have tried that but menu item accelerator won't work. And we need them for fundamental items like cut/copy/paste.
winit's NSView will stop the input chain to itself and won't pass to menu.
I really tried every single combination I could think of, they would just bring other side affects such as webview won't get keyinput instead.

amrbashir:
Maybe we should contact winit maintainers for a possible solution?

wusyong:
Whole winit's window event heavily relies on their custom NSView.
Many events were initially implemented in NSWindow, but they moved to NSView.
We have made a lot of effort to avoid this in tao.
I was on the side to keep original view as well. But every time I did this way, it always backfired with horrifying regressions.
I don't see any possible way winit can get away with this.
Best we could do is making `new` and `new_as_child` literally the same (add as subview).
And offer another platform specific method just like `new_gtk` to do the replacing approach.

amrbashir:
I'll keep the winit and x11 implementation like this for now, if you feel you can refactor `new` to add a subview and it will fill the window and resize with it, I am fine with that.

wusyong:
@amrbashir I added the new method for macOS

src/lib.rs
```diff
@@ -64,109 +163,1406 @@
...
+        target_os = "openbsd",
+      ))]
+      gtk_widget: None,
+    }
+  }
+
+  /// Create the webview as the content view instead of subview on macOS to avoid original content
+  /// view blocks several event delegate methods.
+  #[cfg(target_os = "macos")]
+  pub fn new_as_content_view(window: &'a impl HasWindowHandle) -> Self {
```
amrbashir: We need to add a similar method on `WebView` struct

src/lib.rs
```diff
@@ -64,109 +163,1406 @@
...
+  ///
+  /// - **Linux**: Only X11 is supported, if you want to support Wayland too, use [`WebViewBuilder::new_gtk`].
+  ///
+  ///   Although this methods only needs an X11 window handle, you use webkit2gtk, so you still need to initialize gtk
+  ///   by callling [`gtk::init`] and advance its loop alongside your event loop using [`gtk::main_iteration_do`].
+  ///   Checkout the [Platform Considerations](https://docs.rs/wry/latest/wry/#platform-considerations) section in the crate root documentation.
+  /// - **macOS**: This method is as same as `new_as_child` which will create the webview as a `NSView` subview of the `parent` window's
+  ///   content view. The webview will auto-resize when the passed handle is resized. If you wan
+  ///   to prevent several bugs menu items, accelerators, IME, cursor icons not working because of
+  ///   existing content view blocks the event chain (like how `winit` does), use `new_as_content_view` instead.
```
amrbashir: I would remove explicit winit mention here.

lucasfernog:
@wusyong WebViewBuilder::new was completely broken on macOS (both for winit and tao) so I changed that to replace the contentView by default and changed `new_as_content_view` to `with_as_subview` as an ext method instead of new constructor.

amrbashir:
That's sounds the same as new_as_child

lucasfernog:
> That's sounds the same as new_as_child

Semantics matter 😅 just like the previous new was the same as new_as_child, but you can't reuse the same function for that purpose.

lucasfernog (approved):

rhysd:
@wusyong I think the overview description in README would need update as well :)

> WRY connects the web engine on each platform and provides easy to use and unified interface to render WebView. It also re-exports [TAO](https://crates.io/crates/tao) as a module for event loop and window creation.

https://github.com/tauri-apps/wry/blob/dev/README.md#overview

wusyong:
@rhysd Thanks for notifying! I've opened a PR to update it.
https://github.com/tauri-apps/wry/pull/1065

PR #1053, opened 2023-10-27
feat(windows): allow setting memory usage target level to reduce memory consumption of inactive application

rhysd:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist

### Other information

This PR adds an API to control [the memory usage target level](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2memoryusagetargetlevel) for WebView2.

I noticed my application consumed much memory (more than 100MB) and looked for how to reduce the memory usage, and I found the following guide.

https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/MemoryUsageTargetLevel.md

To try this best practice, I implemented the API and had some experiment. I modified `hello_world` example to switch to 'Low' level when the window lost its focus. Then I ran it with monitoring its memory usage with Task Manager. As the result, I could see 94.8% improvement in memory usage on inactive state.

| | In 'Normal' | Just after switch to 'Low' | 30 secs after switch to 'Low' |
|-|-|-|-|
| Mem | 149.8MB | 82.3MB | 7.8MB |
| Rate | 100% | 54.9% | 5.2% |

```diff
diff --git a/examples/hello_world.rs b/examples/hello_world.rs
index f74d276..8509023 100644
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -12,11 +12,14 @@ fn main() -> wry::Result<()> {
     webview::WebViewBuilder,
   };

+  #[cfg(target_os = "windows")]
+  use wry::webview::WebviewExtWindows as _;
+
   let event_loop = EventLoop::new();
   let window = WindowBuilder::new()
     .with_title("Hello World")
     .build(&event_loop)?;
-  let _webview = WebViewBuilder::new(window)?
+  let webview = WebViewBuilder::new(window)?
     .with_url("https://www.netflix.com/browse")?
     // .with_incognito(true)
     .build()?;
@@ -30,6 +33,14 @@ fn main() -> wry::Result<()> {
         event: WindowEvent::CloseRequested,
         ..
       } => *control_flow = ControlFlow::Exit,
+      #[cfg(target_os = "windows")]
+      Event::WindowEvent {
+        event: WindowEvent::Focused(gained),
+        ..
+      } => {
+        println!("Focus event: {}", gained);
+        webview.set_low_memory_usage(!gained);
+      }
       _ => (),
     }
   });
```

### In 'Normal' level

normal

### Just after going to 'Low'

just after

### 30 seconds passed after going to 'Low'

tmp3

### DIFF
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -1140,8 +1140,24 @@ pub trait WebviewExtWindows {
   /// Returns WebView2 Controller
   fn controller(&self) -> ICoreWebView2Controller;
 
-  // Changes the webview2 theme.
+  /// Changes the webview2 theme.
   fn set_theme(&self, theme: Theme);
+
+  /// Sets [the memory usage target level][1] to 'Low'. There are two levels 'Low' and 'Normal' and
+  /// the default level is 'Normal'. When the application is going inactive, setting the level to
+  /// 'Low' can significantly save the memory used by the application.
+  ///
+  /// Passing `true` to `enabled` argument means the WebView will switch the level to 'Low' and
+  /// passing `false` will switch to 'Normal'.
+  ///
+  /// The logic when application goes to inactive/active is specific to the application.
+  /// [`tao::event::WindowEvent::Focused`] event may be useful.
+  ///
+  /// Please read [the guide for WebView2][2] for more details.
+  ///
+  /// [1]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2memoryusagetargetlevel
+  /// [2]: https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/MemoryUsageTargetLevel.md
+  fn set_low_memory_usage(&self, enabled: bool);
 }
 
 #[cfg(target_os = "windows")]
@@ -1152,6 +1168,10 @@ impl WebviewExtWindows for WebView {
 
   fn set_theme(&self, theme: Theme) {
     self.webview.set_theme(theme)
+  }
+
+  fn set_low_memory_usage(&self, enabled: bool) {
+    self.webview.set_low_memory_usage(enabled);
   }
 }
 
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -1143,23 +1143,20 @@ pub trait WebviewExtWindows {
   /// Changes the webview2 theme.
   fn set_theme(&self, theme: Theme);
 
-  /// Sets [the memory usage target level][1] to 'Low'. There are two levels 'Low' and 'Normal' and
+  /// Sets the [memory usage target level][1] to 'Low'. There are two levels 'Low' and 'Normal' and
   /// the default level is 'Normal'. When the application is going inactive, setting the level to
-  /// 'Low' can significantly save the memory used by the application.
+  /// 'Low' can significantly reduce the application's memory consumption.
   ///
-  /// Passing `true` to `enabled` argument means the WebView will switch the level to 'Low' and
-  /// passing `false` will switch to 'Normal'.
+  /// When to best use this mode depends on the app in question. Most commonly it's called when
+  /// the app's visiblity state changes.
   ///
-  /// The logic when application goes to inactive/active is specific to the application.
-  /// [`tao::event::WindowEvent::Focused`] event may be useful.
+  /// Please read the [guide for WebView2][2] for more details.
   ///
-  /// Please read [the guide for WebView2][2] for more details.
-  ///
-  /// This method internally uses an API added at WebView2 v1.0.1020.30. When it is called with
-  /// WebView2 older than the version, it does nothing.
+  /// This method uses a WebView2 API added in Runtime version 114.0.1823.32. When it is used in
+  /// an older Runtime version, it does nothing.
   ///
   /// [1]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2memoryusagetargetlevel
-  /// [2]: https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/MemoryUsageTargetLevel.md
+  /// [2]: https://learn.microsoft.com/de-de/dotnet/api/microsoft.web.webview2.core.corewebview2.memoryusagetargetlevel?view=webview2-dotnet-1.0.2088.41&preserve-view=true#remarks
   fn set_low_memory_usage(&self, enabled: bool);
 }
 
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -1171,7 +1171,7 @@ pub trait WebviewExtWindows {
   /// an older Runtime version, it does nothing.
   ///
   /// [1]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2memoryusagetargetlevel
-  /// [2]: https://learn.microsoft.com/de-de/dotnet/api/microsoft.web.webview2.core.corewebview2.memoryusagetargetlevel?view=webview2-dotnet-1.0.2088.41&preserve-view=true#remarks
+  /// [2]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.memoryusagetargetlevel?view=webview2-dotnet-1.0.2088.41#remarks
   fn set_memory_usage_level(&self, level: MemoryUsageLevel);
 }
 
--- a/src/webview/mod.rs
+++ b/src/webview/mod.rs
@@ -1134,6 +1134,23 @@ pub fn webview_version() -> Result<String> {
   platform_webview_version()
 }
 
+/// The [memory usage target level][1]. There are two levels 'Low' and 'Normal' and the default
+/// level is 'Normal'. When the application is going inactive, setting the level to 'Low' can
+/// significantly reduce the application's memory consumption.
+///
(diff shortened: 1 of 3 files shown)


FabianLars:
First of all, thanks for the awesome contribution!

Secondly, this now introduces a problem we haven't dealt with in wry yet afaik. This api is fairly new and we can't assume that the user has a webview2 version this new. in general, in case the wv2 auto updater is disabled or something, but more importantly, windows 7 and 8 are stuck with an older webview2 version than required.
I am not really sure how we would implement a failguard here, just raising awareness for the team :)

One thing i'm not sure about is if we're even alowed to update the wv2 sdk (via the webview2-com crate) or if that alone can already cause issues if the sdk is newer than the runtime 🤔

rhysd:
@FabianLars Thanks for the point.

> This api is fairly new and we can't assume that the user has a webview2 version this new. 

If this method is not available, I believe the program reaches [this line](https://github.com/tauri-apps/wry/pull/1053/files#diff-7134ef3789fee9def2afb1d6196ed5eff5f37b7bbad5eceaa5d85cbf3e3513deR973) (please correct me if I'm wrong). So `set_low_memory_usage` is no-op in this case and it is okay because this is kind of optimization. When this method is called with older WebView2, it does nothing. It means the memory usage doesn't change and there would be no problem. That's why I implemented this API not to return `Error` value and instead to ignore any errors.

It would be better to describe this method is no-op with older WebView2 instance in the API document.

FabianLars:
I'm not sure how aware of the runtime the sdk is, but i still have a Window 7 VM so i'm just gonna try it out :)

rhysd:
Thank you. I have only a Windows 10 machine...

FabianLars:
Alrighty, indeed doesn't seem to crash on Win7 (and also builds fine on it directly too). great!

src/webview/mod.rs
```diff
@@ -1140,8 +1140,27 @@ pub trait WebviewExtWindows {
...
+  /// Changes the webview2 theme.
   fn set_theme(&self, theme: Theme);
+
+  /// Sets [the memory usage target level][1] to 'Low'. There are two levels 'Low' and 'Normal' and
+  /// the default level is 'Normal'. When the application is going inactive, setting the level to
+  /// 'Low' can significantly save the memory used by the application.
+  ///
+  /// Passing `true` to `enabled` argument means the WebView will switch the level to 'Low' and
+  /// passing `false` will switch to 'Normal'.
+  ///
```
FabianLars:
```suggestion
```

I feel like this is clear from the function declaration tbh.

src/webview/mod.rs
```diff
@@ -1140,8 +1140,27 @@ pub trait WebviewExtWindows {
   /// Returns WebView2 Controller
   fn controller(&self) -> ICoreWebView2Controller;
 
-  // Changes the webview2 theme.
+  /// Changes the webview2 theme.
   fn set_theme(&self, theme: Theme);
+
+  /// Sets [the memory usage target level][1] to 'Low'. There are two levels 'Low' and 'Normal' and
```
FabianLars:
```suggestion
  /// Sets the [memory usage target level][1] to 'Low'. There are two levels 'Low' and 'Normal' and
```

src/webview/mod.rs
```diff
@@ -1140,8 +1140,27 @@ pub trait WebviewExtWindows {
...
+  /// the default level is 'Normal'. When the application is going inactive, setting the level to
+  /// 'Low' can significantly save the memory used by the application.
+  ///
+  /// Passing `true` to `enabled` argument means the WebView will switch the level to 'Low' and
+  /// passing `false` will switch to 'Normal'.
+  ///
+  /// The logic when application goes to inactive/active is specific to the application.
+  /// [`tao::event::WindowEvent::Focused`] event may be useful.
+  ///
+  /// Please read [the guide for WebView2][2] for more details.
```
FabianLars:
```suggestion
  /// Please read the [guide for WebView2][2] for more details.
```

src/webview/mod.rs
```diff
@@ -1140,8 +1140,27 @@ pub trait WebviewExtWindows {
...
+  ///
+  /// This method internally uses an API added at WebView2 v1.0.1020.30. When it is called with
+  /// WebView2 older than the version, it does nothing.
```
FabianLars: Since Wry dictates the SDK version we can focus on the runtime version here 
```suggestion
  /// This method uses a WebView2 API added in Runtime version 114.0.1823.32. When it is used in
  /// an older Runtime version, it does nothing.
```

src/webview/mod.rs
```diff
@@ -1140,8 +1140,27 @@ pub trait WebviewExtWindows {
...
+  ///
+  /// [1]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2memoryusagetargetlevel
+  /// [2]: https://github.com/MicrosoftEdge/WebView2Feedback/blob/main/specs/MemoryUsageTargetLevel.md
```
FabianLars: I _think_ we should link to the rendered version here. I'd recommend the C# version over C++/WinRT even though that would be closer to the Rust version but the latter is really unreadable lol
```suggestion
  /// [2]: https://learn.microsoft.com/de-de/dotnet/api/microsoft.web.webview2.core.corewebview2.memoryusagetargetlevel?view=webview2-dotnet-1.0.2088.41&preserve-view=true#remarks
```

src/webview/mod.rs
```diff
@@ -1140,8 +1140,27 @@ pub trait WebviewExtWindows {
...
+
+  /// Sets [the memory usage target level][1] to 'Low'. There are two levels 'Low' and 'Normal' and
+  /// the default level is 'Normal'. When the application is going inactive, setting the level to
+  /// 'Low' can significantly save the memory used by the application.
+  ///
+  /// Passing `true` to `enabled` argument means the WebView will switch the level to 'Low' and
+  /// passing `false` will switch to 'Normal'.
+  ///
+  /// The logic when application goes to inactive/active is specific to the application.
+  /// [`tao::event::WindowEvent::Focused`] event may be useful.
```
FabianLars:
```suggestion
  /// When to best use this mode depends on the app in question. Most commonly it's called when
  /// the app's visiblity state changes.
```
I'd rather recommend using it when the app goes invisible, but if you prefer we can include both, visibility and focus changes.

src/webview/mod.rs
FabianLars: If you want to keep this we should rework the phrasing a bit imho, so please scream at me if you disagree :)

src/webview/mod.rs
```diff
@@ -1140,8 +1140,27 @@ pub trait WebviewExtWindows {
   /// Returns WebView2 Controller
   fn controller(&self) -> ICoreWebView2Controller;
 
-  // Changes the webview2 theme.
+  /// Changes the webview2 theme.
   fn set_theme(&self, theme: Theme);
+
+  /// Sets [the memory usage target level][1] to 'Low'. There are two levels 'Low' and 'Normal' and
+  /// the default level is 'Normal'. When the application is going inactive, setting the level to
+  /// 'Low' can significantly save the memory used by the application.
```
FabianLars:
```suggestion
  /// 'Low' can significantly reduce the application's memory consumption.
```

FabianLars (changes requested):
Just some few wording changes because i feel picky today, sorry 😅

src/webview/mod.rs
```diff
@@ -1153,6 +1172,10 @@ impl WebviewExtWindows for WebView {
   fn set_theme(&self, theme: Theme) {
     self.webview.set_theme(theme)
   }
+
+  fn set_low_memory_usage(&self, enabled: bool) {
```
amrbashir: I would suggest changing this method to take an Enum instead, to allow the possibility of other memory modes.

```suggestion
  fn set_memory_usage_level(&self, memory: MemoryLevel) {
```

rhysd: Yeah, more levels may be added in the future. And this API may be promoted to cross-platform and other WebView may have different variations. I agree with making this an enum.

rhysd:
`set_memory_usage` and `MemoryUsage` may be more generic and appropriate (I have no strong opinion).

amrbashir (dismissed):

src/webview/mod.rs
```diff
@@ -1134,14 +1134,45 @@ pub fn webview_version() -> Result<String> {
...
+  /// When to best use this mode depends on the app in question. Most commonly it's called when
+  /// the app's visiblity state changes.
+  ///
+  /// Please read the [guide for WebView2][2] for more details.
+  ///
+  /// This method uses a WebView2 API added in Runtime version 114.0.1823.32. When it is used in
+  /// an older Runtime version, it does nothing.
+  ///
+  /// [1]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2memoryusagetargetlevel
+  /// [2]: https://learn.microsoft.com/de-de/dotnet/api/microsoft.web.webview2.core.corewebview2.memoryusagetargetlevel?view=webview2-dotnet-1.0.2088.41&preserve-view=true#remarks
```
FabianLars: accidentally used the german link, sry
```suggestion
  /// [2]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.memoryusagetargetlevel?view=webview2-dotnet-1.0.2088.41#remarks
```

FabianLars (approved):

PR #1067, opened 2023-11-09
feat: re-export raw-window-handle crate

### LINKED ISSUE #1066: Make `HasRawWindowHandle` public

As [README.md](https://github.com/tauri-apps/wry#readme) states, `HasRawWindowHandle` is an important part of public API of this crate.

> The webview requires a running event loop and a window type that implements HasWindowHandle, or a gtk container widget if you need to support X11 and Wayland

However, currently wry doesn't re-export the trait. This means that, if users want to use the trait, they need to manage `raw-window-handle` crate as their dependency keeping the version compatible with wry's `raw-window-handle` dependency.

The latest version of [raw-window-handle](https://crates.io/crates/raw-window-handle) crate is v0.6.0. And wry depends on v0.5. So users can easily install wrong version of the crate by `cargo add` and will face a weird error. Though they are the same name, traits from different versions of same crate are handled as completely different traits.

To avoid managing the dependency separately in two places, it is a good idea to re-export wry's `HasRawWindowHandle` trait as wry re-exported `tao` as `wry::application` previously.

Keep hiding the trait would be an option as strategy of maintenance though I can't come up with any benefits.

- [x] Yes
- [ ] No

**Additional context**

N/A

rhysd:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve #1066

### Other information

I didn't add a file under `.changes` directory because this change can be amended to changes of #1041.

### DIFF
--- a/.changes/rwh.md
+++ b/.changes/rwh.md
@@ -10,7 +10,7 @@ Refactor new method to take raw window handle instead. Following are APIs got af
   - Position field in `FileDrop` event is now `Position` instead of `PhysicalPosition`. Users need to handle scale factor
   depend on the situation they have.
   - `Webview::inner_size` is removed.
-  - `HasRawWindowHandle` trait is re-exported as `wry::HasRawWindowHandle`.
+  - [raw-window-handle](https://docs.rs/raw-window-handle/latest/raw_window_handle/) crate is re-exported as `wry::raw_window_handle`.
 
 This also means that we removed `tao` as a dependency completely which required some changes to the Android backend:
   - We exposed the `android_setup` function that needs to be called once to setup necessary logic.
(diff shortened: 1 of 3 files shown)


amrbashir:
I would prefer to re-export the whole `raw-window-handle` crate.

rhysd:
That also makes sense to me. Do you have any suggestion about module name?

- `wry::raw_window_handle`
- `wry::rwh`

amrbashir:
`raw_window_handle` is fine

rhysd:
Sure. I will do the change tonight.

rhysd:
@amrbashir I updated this branch based on your comments.

wusyong (approved):

PR #1077, opened 2023-11-13
Add cfg_aliases for easier feature configuration

wusyong:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Checklist


pewsheen (dismissed):

src/lib.rs
```diff
@@ -145,6 +145,8 @@
 //!
 //! Wry uses a set of feature flags to toggle several advanced features.
 //!
+//! - `native` (default): Enables native WebView framework on the platform. This must be enabled
+//! for the crate to work. This feature is in preparasion of other ports like cef and servo.
```
FabianLars:
```suggestion
//! for the crate to work. This feature was added in preparation for other ports like cef and servo.
```
Only wanted to fix the preparation typo, so if you're not happy with the new wording we can change it back again ;)

wusyong: Thanks! I'll fix it

Cargo.toml
```diff
@@ -24,15 +24,19 @@ rustc-args = ["--cfg", "docsrs"]
...
 objc-exception = ["objc/exception"]
 file-drop = []
 protocol = []
 devtools = []
 transparent = []
 fullscreen = []
-linux-body = ["webkit2gtk/v2_40"]
+linux-body = ["webkit2gtk/v2_40", "native"]
 mac-proxy = []
+native = ["javascriptcore-rs", "webkit2gtk", "webkit2gtk-sys", "dep:gtk", "soup3", "x11-dl", "gdkx11"]
```
lucasfernog: what does `native` mean here? specially since it's enabling gtk

wusyong: I was thinking about macOS and Windows support in the future. They may also have servo and cef ports.
With `native` feature, it indicates existing frameworks from platforms.
And `gtk` is cfg alias only will be used in private to indicate it's native and on Linux.
This could be easier for us to configure internally.
And for public they just need to know, there's a native feature.

amrbashir: Maybe `os-webview` is a better name?

Cargo.toml
```diff
@@ -24,15 +24,19 @@ rustc-args = ["--cfg", "docsrs"]
...
-default = ["file-drop", "objc-exception", "protocol"]
+default = ["file-drop", "objc-exception", "protocol", "os-webview"]
 objc-exception = ["objc/exception"]
 file-drop = []
 protocol = []
 devtools = []
 transparent = []
 fullscreen = []
-linux-body = ["webkit2gtk/v2_40"]
+linux-body = ["webkit2gtk/v2_40", "os-webview"]
```
lucasfernog, line 34: maybe rename this one to webkit2gtk-body now that we are preparing for other implementations?

if we rename it here, we need to rename in code too

lucasfernog (approved):

PR #1091, opened 2023-11-21
feat: add WebView::bounds getter

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

### Checklist

### DIFF
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -600,6 +600,68 @@ impl InnerWebView {
     Ok(())
   }
 
+  pub fn bounds(&self) -> Rect {
+    let mut bounds = Rect {
+      x: 0,
+      y: 0,
+      width: 0,
+      height: 0,
+    };
+
+    if let (Some(xlib), Some(display), Some(window_handle)) =
+      (&self.xlib, self.x11_display, self.x11_window)
+    {
+      unsafe {
+        let mut attributes = std::mem::MaybeUninit::new(x11_dl::xlib::XWindowAttributes {
+          x: 0,
+          y: 0,
+          height: 0,
+          width: 0,
+          border_width: 0,
+          depth: 0,
+          visual: std::ptr::null_mut(),
+          root: 0,
+          class: 0,
+          bit_gravity: 0,
+          win_gravity: 0,
+          backing_pixel: 0,
+          backing_store: 0,
+          backing_planes: 0,
+          save_under: 0,
+          colormap: 0,
+          map_installed: 0,
+          map_state: 0,
+          all_event_masks: 0,
+          your_event_mask: 0,
+          do_not_propagate_mask: 0,
+          override_redirect: 0,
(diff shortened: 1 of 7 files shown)


src/webview2/mod.rs
```diff
@@ -1009,6 +1010,35 @@ impl InnerWebView {
...
+        width: (rect.right - rect.left) as u32,
+        height: (rect.bottom - rect.top) as u32,
+      }
+    } else {
+      Rect {
+        x: 0,
+        y: 0,
+        width: 0,
+        height: 0,
+      }
```
amrbashir: we can return `self.controller.bounds` here.

src/wkwebview/mod.rs
```diff
@@ -1090,6 +1090,30 @@ r#"Object.defineProperty(window, 'ipc', {
     Ok(())
   }
 
+  pub fn bounds(&self) -> Rect {
+    if self.is_child {
```
amrbashir: I believe we can run the same logic whether it is child or not or with a small tweak.

src/webkitgtk/mod.rs
```diff
@@ -600,6 +600,68 @@ impl InnerWebView {
...
+    let mut bounds = Rect {
+      x: 0,
+      y: 0,
+      width: 0,
+      height: 0,
+    };
+
+    if let (Some(xlib), Some(display), Some(window_handle)) =
+      (&self.xlib, self.x11_display, self.x11_window)
+    {
```
amrbashir, line 608: We can probably skip using x11 and use `self.gtk_window` directly if it is a child

src/webkitgtk/mod.rs
```diff
@@ -600,6 +600,68 @@ impl InnerWebView {
...
+        }
+      }
+    } else if let Some(window) = &self.gtk_window {
+      if self.is_child {
+        let position = window.position();
+        bounds.x = position.0;
+        bounds.y = position.1;
+      }
+
+      let size = window.size();
```
amrbashir, line 625: if we are not a child, we should be using `self.webview.allocated_size` instead

amrbashir (changes requested):

src/webkitgtk/mod.rs
lucasfernog: I tried that, but I use i3wm and by default it expands the window to fill the tile, so the bounds wrong without using x11. if we don't really care about tiling window managers then yeah we can skip to gtk directly

lucasfernog:
also fixed an issue with the webview size on macOS. let me know if you want a separate PR

src/webkitgtk/mod.rs
amrbashir: it is fine to keep then

amrbashir (approved):

PR #1093, opened 2023-11-22
feat: add tracing for `evaluate_script`, `ipc_handler` and `custom_protocols`

lucasfernog-crabnebula:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

### Checklist

### Other information

This PR adds tracing spans to the most performance-relevant APIs of wry so downstream consumers can instrument them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,11 +32,12 @@ devtools = [ ]
 transparent = [ ]
 fullscreen = [ ]
 linux-headers = [ "webkit2gtk/v2_36" ]
+tracing = [ "dep:tracing" ]
 
 [dependencies]
 libc = "0.2"
 log = "0.4"
-tracing = "0.1"
+tracing = { version = "0.1", optional = true }
 once_cell = "1"
 serde = { version = "1.0", features = [ "derive" ] }
 serde_json = "1.0"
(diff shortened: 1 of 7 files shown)


src/webview/wkwebview/mod.rs
```diff
@@ -158,6 +163,8 @@ impl InnerWebView {
             let _: () = msg_send![body_stream, close];
           }
 
+          tracing::debug!("done reading body");
```
amrbashir: is this a leftover? seems like other platforms don't have a similar `debug!` call

JonasKruckenberg (approved):

Cargo.toml
```diff
@@ -36,6 +36,7 @@ linux-headers = [ "webkit2gtk/v2_36" ]
 [dependencies]
 libc = "0.2"
 log = "0.4"
+tracing = "0.1"
```
amrbashir: should we add this behind a feature flag (disabled by default)? so ultimately tauri uses could opt-into disabling the tracing if they wish so

amr-crabnebula (approved):

amrbashir (approved):

PR #1127, opened 2023-12-20
feat: add bevy example

amrbashir:


amrbashir:
looks like this will require some dependencies on Linux, which I don't want to enforce in this repo, gonna close

PR #1138, opened 2024-01-12
servo-wry-demo: fix build on Windows and update README

delan:
I managed to get the Servo demo working on Linux and even Windows, at least without the seamless window decorations. This patch updates the README accordingly, and fixes some small compile errors on Windows.

Untitled5106

Untitled5107


wusyong (approved):
Thanks for contributing on Windows! I'm actually porting to this as well!

PR #1144, opened 2024-01-22
fix(deps): update raw-window-handle to v0.6

### LINKED ISSUE #1143: Update raw-window-handle dependency to v0.6

raw-window-handle crate dependency of this crate is out of date. It is better to update it to the latest because

- some crates require the latest version of raw-window-handle. For example, wgpu v0.19 requires raw-window-handle v0.6 and doesn't work with v0.5
- v0.6 provides safer interface. `WindowHandle` manages lifetime not to make the underlying handle invalid. It is better to make users migrate to the safer interface
- Currently wry has both v0.6 and v0.5 as dev dependency redundantly because winit enables `rwh_06` by default

Bump up raw-window-handle to v0.6. This means the following breaking changes:

- wry's APIs will require `HasWindowHandle` trait instead of `HasRawWindowHandle` since `HasRawWindowHandle` was deprecated at v0.6
- `wry::raw_window_handle` will re-export v0.6

- Support both v0.6 and v0.5 as winit does. This would increase maintenance cost but possible
- Keep v0.5 until v0.6 is mandatory due to some reason

**Additional context**

Nothing.

rhysd:
Fix #1143
Close #1062 

I confirmed

- `cargo check`
- `cargo test`
- `winit`, `simple`, `wgpu` examples

on the following targets

- `x86_64-apple-darwin`
- `x86_64-pc-windows-msvc`
- `x86_64-unknown-linux-gnu`

And I confirmed `cargo check` on the following targets (because I don't have the environments)

- `aarch64-apple-darwin`
- `x86_64-linux-android`
- `aarch64-apple-ios`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ serde_json = "1.0"
 thiserror = "1.0"
 url = "2.4"
 http = "0.2"
-raw-window-handle = { version = "0.5", features = [ "std" ] }
+raw-window-handle = { version = "0.6", features = [ "std" ] }
 
 [target."cfg(any(target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
 javascriptcore-rs = { version = "=1.1", features = [ "v2_28" ], optional = true }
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -117,6 +117,6 @@ tao-macros = { version = "0.1.0" }
 [dev-dependencies]
 http-range = "0.1.5"
 pollster = "0.3.0"
-tao = { version = "0.24", default-features = false, features = [ "rwh_05" ] }
-wgpu = "0.18.0"
-winit = { version = "0.29", features = [ "rwh_05" ] }
+tao = { version = "0.24", default-features = false, features = [ "rwh_06" ] }
```
rhysd: I was not sure this line should be simplified to `tao = "0.24"` so I preserved the configuration as much as possible.

amrbashir: let's simplify it

rhysd: Sure.

amrbashir (approved):

amrbashir:
Thanks

PR #1167, opened 2024-02-12
feat(windows) add print_to method in wry

huangmingg:
First PR to get the ball rolling for https://github.com/tauri-apps/plugins-workspace/issues/293

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -303,6 +303,8 @@ impl InnerWebView {
 
   pub fn print(&self) {}
 
+  pub fn print_to(&self, _printer_name: &str) {}
+
   pub fn url(&self) -> Url {
     let (tx, rx) = bounded(1);
     MainPipe::send(WebViewMessage::GetUrl(tx));
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1301,6 +1301,13 @@ impl WebView {
     Ok(())
   }
 
+  /// Silently print to printer with default settings
+  /// ## Platform-specific
+  /// - **Windows**: Available on webview >= 16 only.
+  pub fn print_to(&self, printer_name: &str) -> Result<()> {
+    self.webview.print_to(printer_name)
+  }
+
   /// Open the web inspector which is usually called dev tool.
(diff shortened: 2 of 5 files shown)


amrbashir:
I feel like advanced printing functionality could be left out of `wry`, since tauri has [`with_webview`](https://docs.rs/tauri/latest/tauri/window/struct.Window.html#method.with_webview) and could be used to implement the needed APIs in the plugin rather directly in `wry`.

cc @wusyong @lucasfernog

wusyong:
I prefer to use [with_webview](https://docs.rs/tauri/latest/tauri/window/struct.Window.html#method.with_webview) too. And this one seems to be platform specific.

huangmingg:
What are the advantages of pushing this feature downstream to `tauri`? The ability to silent print from a webview sounds like the responsibility of `wry`.

It also seems using `with_webview ` will require exposing more fields downstream to tauri, such as ICoreWebView2Environment of windows to create the printer setting.

amrbashir:
> What are the advantages of pushing this feature downstream to `tauri`? The ability to silent print from a webview sounds like the responsibility of `wry`.

advanced printing functionality doesn't sound like a thing that should be maintained in `wry`, it should instead be in its own plugin so it is easier to maintain both projects and easier to contribute to.

> It also seems using `with_webview ` will require exposing more fields downstream to tauri, such as ICoreWebView2Environment of windows to create the printer setting.

You can still get the webview2 environment though the controller.

huangmingg:
How can i get the webview2 environment via the controller?

FabianLars:
It's on the webview, not the controller but you can use this method to get it: https://docs.rs/webview2-com/latest/webview2_com/Microsoft/Web/WebView2/Win32/trait.ICoreWebView2_2_Impl.html#tymethod.Environment

amrbashir:
yes you can get ICoreWebview2 from controller and get the environment from ICoreWebview2

lucasfernog:
I agree this should be left out of wry, at least until it is supported on more platforms

PR #1182, opened 2024-03-05
refactor!: refactor windows and linux backends for better maintainability

### LINKED ISSUE #1173: Crash on multiwebview example in WIndows
**Describe the bug**
When running multiwebview example somtimes it crashes with
```console
Error: WebView2Error(WindowsError(Error { code: HRESULT(0x8007139F), message: "The group or resource is not in the correct state to perform the requested operation." }))
```

**Steps To Reproduce**
```console
git clone https://github.com/tauri-apps/wry
cd wry
cargo run --example multiwebview
```

**Expected behavior**
It should work without crash

```console
OS: Windows 11
Rustc: 
cargo 1.75.0 (1d8b05cdd 2023-11-20)
rustc 1.75.0 (82e1608df 2023-12-21)
```

**Additional context**
It works well on `MacOS`
it doesn't happen when adding only single webview
it happens when adding 2 or more

amrbashir:

### DIFF
--- a/examples/winit.rs
+++ b/examples/winit.rs
@@ -71,7 +71,7 @@ fn main() -> wry::Result<()> {
           event: WindowEvent::Resized(size),
           ..
         } => {
-          _webview.set_bounds(wry::Rect {
+          let _ = _webview.set_bounds(wry::Rect {
             x: 0,
             y: 0,
             width: size.width,
--- a/examples/winit.rs
+++ b/examples/winit.rs
@@ -71,12 +71,14 @@ fn main() -> wry::Result<()> {
           event: WindowEvent::Resized(size),
           ..
         } => {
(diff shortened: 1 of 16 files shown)


examples/winit.rs
```diff
@@ -71,7 +71,7 @@ fn main() -> wry::Result<()> {
           event: WindowEvent::Resized(size),
           ..
         } => {
-          _webview.set_bounds(wry::Rect {
+          let _ = _webview.set_bounds(wry::Rect {
```
wusyong: LGTM overall, but I think all example should unwrap the call instead. We could find which part gets the error in this way.

amrbashir: good idea, done

wusyong (approved):

PR #1183, opened 2024-03-05
refactor(ipc): use http::Request on with_ipc_handler

lucasfernog:
This pull request changes the ipc_handler closure to get a new IpcRequest type instead of only receive the message string.
On macOS, iOS and Windows, we can read the request URL which enhances the security checks to verify the request source (currently only the url() function can be used, which does not work on contexts where an iframe is used).

For custom protocols, I believe the Origin header can be used instead of this.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,3 +2,11 @@
 # target = "aarch64-linux-android"
 [target.x86_64-apple-darwin]
 rustflags = ["-C", "link-arg=-mmacosx-version-min=10.12"]
+
+[target.x86_64-pc-windows-msvc]
+linker = "lld"
+rustflags = [
+  "-Lnative=/Users/lucas/.xwin/crt/lib/x86_64",
+  "-Lnative=/Users/lucas/.xwin/sdk/lib/um/x86_64",
+  "-Lnative=/Users/lucas/.xwin/sdk/lib/ucrt/x86_64",
+]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,11 +2,3 @@
 # target = "aarch64-linux-android"
 [target.x86_64-apple-darwin]
 rustflags = ["-C", "link-arg=-mmacosx-version-min=10.12"]
-
-[target.x86_64-pc-windows-msvc]
-linker = "lld"
-rustflags = [
-  "-Lnative=/Users/lucas/.xwin/crt/lib/x86_64",
-  "-Lnative=/Users/lucas/.xwin/sdk/lib/um/x86_64",
-  "-Lnative=/Users/lucas/.xwin/sdk/lib/ucrt/x86_64",
-]
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
+use http::Request;
 use http::{
   header::{HeaderName, HeaderValue, CONTENT_TYPE},
   Request,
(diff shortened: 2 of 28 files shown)


.cargo/config.toml
```diff
@@ -2,3 +2,11 @@
...
 [target.x86_64-apple-darwin]
 rustflags = ["-C", "link-arg=-mmacosx-version-min=10.12"]
+
+[target.x86_64-pc-windows-msvc]
+linker = "lld"
+rustflags = [
+  "-Lnative=/Users/lucas/.xwin/crt/lib/x86_64",
+  "-Lnative=/Users/lucas/.xwin/sdk/lib/um/x86_64",
+  "-Lnative=/Users/lucas/.xwin/sdk/lib/ucrt/x86_64",
+]
```
amrbashir: leftover?

src/lib.rs
```diff
@@ -1025,6 +1025,26 @@ impl<'a> WebViewBuilder<'a> {
   }
 }
 
+/// Request from a IPC message. See [`WebViewBuilder::with_ipc_handler`] for more information.
+pub struct IpcRequest {
```
amrbashir: instead of a new type, can't we just pass the url as a second argument with `Option<String>`?

src/webkitgtk/mod.rs
```diff
@@ -244,7 +244,9 @@ impl InnerWebView {
 
       if let Some(js) = msg.js_value() {
         if let Some(ipc_handler) = &ipc_handler {
-          ipc_handler(js.to_string());
+          ipc_handler(IpcRequest {
```
amrbashir: we can pass a clone of the webview here and get the url

src/android/binding.rs
```diff
@@ -308,7 +308,7 @@ pub unsafe fn ipc(mut env: JNIEnv, _: JClass, arg: JString) {
 
       let arg = arg.to_string_lossy().to_string();
       if let Some(ipc) = IPC.get() {
-        (ipc.handler)(arg)
+        (ipc.handler)(IpcRequest { body: arg })
```
amrbashir: we can also support this on android by storing a reference to the webview on `Ipc` class and getting the url from it

amrbashir (changes requested):

.cargo/config.toml
lucasfernog: aaaaaaaaaaaaaaaaaaaaaaa

src/lib.rs
lucasfernog: maybe we should use http::Request instead? we might add new parameters to this closure in the future (like a closure to provide a response to the webview) and it's more type safe to have a dedicated type for multiple parameters

src/android/kotlin/RustWebViewClient.kt
```diff
@@ -48,6 +49,7 @@ class RustWebViewClient(context: Context): WebViewClient() {
     }
 
     override fun onPageFinished(view: WebView, url: String) {
+        currentUrl = url
```
amrbashir, line 52: we should use https://developer.android.com/reference/android/webkit/WebView#getUrl() instead

lucasfernog: getUrl can only be used on the main thread, and i feel it's a bottleneck to the IPC

lucasfernog: tried it directly, just crashes :(

amrbashir: then let's leave a comment and document why we went with this manual approach

amrbashir (dismissed):

lucasfernog:
@amrbashir sorry had to push a docs update :D

amrbashir (approved):

PR #1186, opened 2024-03-11
fix(mac): suppress duplicated media permission prompt

### LINKED ISSUE #1101: getDisplayMedia() does not work on MacOS Sonoma 14.0+
**Describe the bug**
If you try getting the user's screen capture it fails with a `NotAllowedError`, this is also an issue reported in tauri, https://github.com/tauri-apps/tauri/issues/2600

**Steps To Reproduce**
- Change the simple.rs example to display any webpage that has a getDisplayMedia call (Gmeet, etc)
- Run `cargo run --example simple`
- Try to share screen from the video calling app
- It will fail

**Expected behavior**
Ideally, the display media should be captured and shown on the screen, just like camera and audio work out of the box

**Screenshots**
Copied the screen shot from the same issue
image

OS: macOS 14.0 (23A344)
Rustc: 1.72.1

**Would you want to assign yourself to resolve this bug?**
- [ ] Yes
- [x] No

pewsheen:
Issue: https://github.com/tauri-apps/tauri/issues/8979
Original fix: #1101 

- #1101 fixed the issue that the prompt does not show up on macOS 14.0 hence the request always being denied.
- In https://github.com/tauri-apps/tauri/issues/8979, the behavior seems to have changed back on macOS 14.2; we should also revert the fix.

I've tested it on my macOS 14.2.1. It would be better to have a second check.

A minimal testing repo: https://github.com/pewsheen/tauri-media-permission/

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -804,16 +804,10 @@ impl InnerWebView {
             run_file_upload_panel as extern "C" fn(&Object, Sel, id, id, id, id),
           );
 
-          // Only disable media dialogs on macOS < 14.0
-          // https://tauri.app/v1/references/webview-versions/
-          let webview_system_version = platform_webview_system_version()?.parse::<i32>();
-          if webview_system_version.is_err() || webview_system_version.unwrap() < 19 {
-            // Disable media dialogs
-            ctl.add_method(
-              sel!(webView:requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:),
-              request_media_capture_permission as extern "C" fn(&Object, Sel, id, id, id, id, id),
-            );
-          }
+          ctl.add_method(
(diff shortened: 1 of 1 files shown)


wusyong (approved):

stolinski:
Is there a chance that this broke getDisplayMedia? I am unable to get the getDisplayMedia permissions prompt and when adding a 
```
        const stream = await navigator.mediaDevices.getDisplayMedia({
          video: true, // Screen capture
          audio: false, // No audio capture
        });
```
to your test repo for tauri-media-permissions, it once again gives the same 
"Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission." 
error without a permissions prompt.

pewsheen:
What's your OS version and did you run the repo by `tauri dev` or the bundled `.app` application?

stolinski:
Edit: Moving this to https://github.com/tauri-apps/tauri/issues/2600

Issue #1195, opened 2024-03-20
Fix `getDispayMedia()` `getUserMedia()` permission prompt on macOS


pewsheen:
Current Status on macOS 14 
- `getUserMedia()`: Show permission prompt for camera and microphone **twice** (application level and webview level)
- `getDispayMedia()`: Normal

but with `requestMediaCapturePermissionForOrigin` delegated in wry
- `getUserMedia()`: Normal
- `getDispayMedia()`: Permission denied due to not showing picker for a display device (`SCContentSharingPicker`)

A possible fix for now is delegating both `requestMediaCapturePermissionForOrigin` and `requestDisplayCapturePermissionForOrigin`. But when we are calling `getDispayMedia()`, it can choose a screen only.

Upstream tracking: https://bugs.webkit.org/show_bug.cgi?id=271688

Workaround PR: https://github.com/tauri-apps/wry/pull/1196

**---  outdated ---**

The media permission prompt was suppressed when macOS 14.0 introduced

Issue: https://github.com/tauri-apps/wry/issues/1101, https://github.com/tauri-apps/tauri/issues/2600
PR: https://github.com/tauri-apps/wry/pull/1111

_note:_ #1111 brings a regression issue: https://github.com/tauri-apps/tauri/issues/9178

After an OS update, which could be 14.2 (not sure), the original issue was solved by Apple, so we revert the fix:

Issue: https://github.com/tauri-apps/tauri/issues/8979
PR: https://github.com/tauri-apps/wry/pull/1186

We need to find the affected OS version (14.0 ~ 14.x.x) and apply the version check like https://github.com/tauri-apps/wry/pull/1111

**TODO**
- [ ] Do version check for macOS 14.0 (#1196)
- [ ] Find the affected macOS version range

FabianLars:
i think i can test 14.1.2 later today

FabianLars:
I can confirm that it only works on 14.1.2 after i changed `(14, 0, _)` to `(14, 0, _) | (14, 1, _)`
(is that the best way to do the pattern matching? 😅)

FabianLars:
Looks like we already know that it's fixed in 14.2.1 if 8979 is indeed the same issue so i'm just gonna try to check 14.2.0 and not the latest 14.2 patch ver

FabianLars:
[...]

PR #1196, opened 2024-03-20
fix(mac): expose getDisplayMedia() permission decision handler

pewsheen:
Related issue: https://github.com/tauri-apps/wry/issues/1195

Temporary fix for the issue. Needs to find the affected range in the future.


src/wkwebview/mod.rs
```diff
@@ -806,10 +808,13 @@ impl InnerWebView {
             run_file_upload_panel as extern "C" fn(&Object, Sel, id, id, id, id),
           );
 
-          ctl.add_method(
-            sel!(webView:requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:),
-            request_media_capture_permission as extern "C" fn(&Object, Sel, id, id, id, id, id),
-          );
+          // https://github.com/tauri-apps/wry/issues/1101
+          if let (14, 0, _) = operating_system_version() {
```
pewsheen: Should we add `#[cfg(target_os = "macos")]` check here?

wusyong: yeah it probably should

wusyong (dismissed):

src/wkwebview/mod.rs
```diff
@@ -806,10 +808,19 @@ impl InnerWebView {
...
-          ctl.add_method(
-            sel!(webView:requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:),
-            request_media_capture_permission as extern "C" fn(&Object, Sel, id, id, id, id, id),
-          );
+          #[cfg(target_os = "macos")]
+          match operating_system_version() {
+            (14, 0, _) => {
+              // Do not suppress media request on 14.0:
+              // https://github.com/tauri-apps/wry/issues/1101
+            }
```
FabianLars:
```suggestion
            (14, 0, _) | (14, 1, _) => {
              // Do not suppress media request on 14.0 & 14.1:
              // https://github.com/tauri-apps/wry/issues/1101
            }
```
https://github.com/tauri-apps/wry/issues/1195#issuecomment-2009402427

FabianLars (changes requested):
(just blocking this because of ongoing discussion in #1195)

pewsheen:
@FabianLars Your private API (or maybe just not document yet?) hint is mega helpful

After I read some WebKit source code
https://github.com/WebKit/WebKit/blob/b7a7805ab3a2a1b719af3112657934e837ee3df5/Source/WebKit/UIProcess/Cocoa/UIDelegate.mm#L1260

```objective-c
bool respondsToRequestMediaCapturePermission = [delegate respondsToSelector:@selector(webView:requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:)];
bool respondsToRequestUserMediaAuthorizationForDevices = [delegate respondsToSelector:@selector(_webView:requestUserMediaAuthorizationForDevices:url:mainFrameURL:decisionHandler:)];
bool respondsToRequestDisplayCapturePermissionForOrigin = [delegate respondsToSelector:@selector(_webView:requestDisplayCapturePermissionForOrigin:initiatedByFrame:withSystemAudio:decisionHandler:)];
```

I guess they are using `_webView:requestDisplayCapturePermissionForOrigin:initiatedByFrame:withSystemAudio:decisionHandler:` for `getDisplayMedia` and yes!

The reason why it's not working on your commit is because the typo LOL

```diff
- _webView:requestDisplayCapturePermissionForSecurityOrigin:initiatedByFrame:withSystemAudio:decisionHandler:
+ _webView:requestDisplayCapturePermissionForOrigin:initiatedByFrame:withSystemAudio:decisionHandler:
```

But it's still not showing the window / screen select panel, just window OR screen.

FabianLars:
nooo, not a freaking typo, I just can't 😂

pewsheen:
Just update the implementation, will expose a closure and let them pick the decision

pewsheen:
Maybe it can just store the decision instead of a closure, since there's only a capture type parameter?

src/wkwebview/display_capture.rs
```diff
@@ -0,0 +1,115 @@
...
+        (*ui_delegate).set_ivar(
+          "display_capture_decision_handler",
+          handler as *mut _ as *mut c_void,
+        );
+      }
+    }
+  }
+}
+
+pub(crate) fn drop_decision_hanlder(webview: *mut Object) {
```
FabianLars, line 79:
```suggestion
pub(crate) fn drop_decision_handler(webview: *mut Object) {
```
(the usages of this function have the same typo)

FabianLars:
> Maybe it can just store the decision instead of a closure, since there's only a capture type parameter?

I'm not 100% sure i really understand. When we use a variable only the developer could decide right? The good thing about the handler is that wkwebview will call that itself and then the dev can implement something that will let the actual app user decide if they want a screen or window prompt. I don't see how we could mimic this with just a variable.

pewsheen:
> > Maybe it can just store the decision instead of a closure, since there's only a capture type parameter?
> 
> I'm not 100% sure i really understand. When we use a variable only the developer could decide right? The good thing about the handler is that wkwebview will call that itself and then the dev can implement something that will let the actual app user decide if they want a screen or window prompt. I don't see how we could mimic this with just a variable.

In this case, the developer will need to implement something like a prompt to let the user choose to capture a screen or a window and call a setter like `set_capture_target(Window or Screen)`, and then call the `getDisplayMedia()`.

Closure is good and may not need an extra set step like the variable approach. But I think it will be a bit more complicated managing the lifetime and ownership when implementing logic inside the closure.

FabianLars:
Hmm, another problem that comes to mind is that this makes it harder for use cases where you load external websites that you don't control since then you don't really know when you ask the user what they want to share. I still think this is still a bit of a problem even if it's your website because then you _have_ to communicate between js and rust to know when to prompt the user.

> Closure is good and may not need an extra set step like the variable approach. But I think it will be a bit more complicated managing the lifetime and ownership when implementing logic inside the closure.

Hmm, yeah that could be a problem. I thought/hoped that the main thing to do here would be to simply call rfd in that closure and leave it at that 🤔

PR #1208, opened 2024-04-01
fix(wkwebview): menu shortcuts

thewh1teagle:
Fix global menu shortcuts on macOS with multiwebview window, when adding webview as child.
Improved version of https://github.com/tauri-apps/wry/pull/1156#issuecomment-2029180822


lucasfernog (approved):
works and I couldn't find a regression :) nice one!

PR #1214, opened 2024-04-05
fix (macos) APP crash when a web custom protocol request is aborted before or during response

### LINKED ISSUE #1142: Reloading webview crashes in an edge case [macOS]
**Describe the bug**
I have a custom menu item to reload webview contents (in tauri). It uses following code: 
```
window.with_webview(|webview| {
          #[cfg(target_os = "macos")]
          unsafe {
            let () = msg_send![webview.inner(), reload];
          }
        });
```

This worked fine until ~last month which started to cause a panic with this message: 
CleanShot 2024-01-21 at 13 20 46@2x

**Steps To Reproduce**
I tried doing the same thing in custom_protocol example, but I can't reproduce this. Can you guide me through debugging what might cause this based on the stack trace, I'd be happy to help narrow down the issue and provide a Wry example / or move the issue to tauri.

more context: I added this code to wry example to trigger reloads which didn't panic: 
```
  event_loop.run(move |event, _, control_flow| {
    *control_flow = ControlFlow::Wait;

    match event {
      Event::WindowEvent { window_id, ref event, .. } => match event {
        WindowEvent::Focused(_) => {
          #[cfg(target_os = "macos")]
          unsafe {
            let webview = _webview.webview();
            let () = msg_send![webview, reload];
          };
        }
        _ => {}
      }
      _ => {}
    }
```

**Expected behavior**
To reload the app without panic.

**Screenshots**
N/A

OS: macOS 14.2.1 (23C71)
Rustc: rustc 1.73.0 (cc66ad468 2023-10-03)

**Additional context**
N/A

### LINKED ISSUE #1189: [macOS] APP crash when a web custom protocol request is aborted before or during response
### Describe the bug

When initiating a `register_uri_scheme_protocol` request from the web side, if the web aborts the request before or during the response, the macOS APP crashes.

### Reproduction

- Register a `register_asynchronous_uri_scheme_protocol`.
- From the web side, initiate a request using this protocol and abort the request before the response.
- The APP crashes 100% of the time.

### Expected behavior

- Before responding, the `responder` should check if the request has been canceled. If the request has been canceled, do not send the response.
- Or, during the response, catch exceptions to prevent the APP from crashing and quitting abruptly when the request is terminated.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 13.6.3 X64
    ✔ rustc: 1.76.0 (07dca489a 2024-02-04)
    ✔ cargo: 1.76.0 (c84b36747 2024-01-18)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-apple-darwin (default)
    - node: 18.17.1
    - pnpm: 8.15.4
    - npm: 9.6.7

[-] Packages
    - tauri [RUST]: 2.0.0-beta.10
    - tauri-build [RUST]: 2.0.0-beta.8
    - wry [RUST]: 0.37.0
    - tao [RUST]: 0.26.1
    - @tauri-apps/api [NPM]: 2.0.0-beta.5
    - @tauri-apps/cli [NPM]: 2.0.0-beta.9

[-] App
    - build-type: bundle
    - CSP: default-src x-http: x-https: blob: data: filesystem: ws: http: https: 'unsafe-eval' 'unsafe-inline'
    - frontendDist: ../feiyu/dist
```

### Stack trace

```text
thread 'tokio-runtime-worker' panicked at /Users/mac/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wry-0.37.0/src/wkwebview/mod.rs:262:28:
Uncaught exception <NSException: 0x10428cf30>
```

### Additional context

Related code: 
https://github.com/tauri-apps/wry/blob/5789bf759ce94e4dad5ff26a08fe81521658a4e4/src/wkwebview/mod.rs#L272
图片

Related issue: https://github.com/tauri-apps/tauri/issues/9177

thomastschurtschenthaler:
closes 1189

I am new to rust and object-c, so not sure if everything is alright with the memory management, but it works well for my tests so far. 
Testfile attached - every third request is aborted. 
[main.rs.zip](https://github.com/tauri-apps/wry/files/14887689/main.rs.zip)

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -250,9 +250,9 @@ impl InnerWebView {
                   let stoppedtaskslock:id = * (* _this).get_ivar("stoppedtaskslock");
                   let _: () = msg_send![stoppedtaskslock, lock];
                   let stoppedtasks:id = *(* _this).get_ivar("stoppedtasks");
-                  let ctns:bool = msg_send![stoppedtasks, containsObject: task];
+                  let hash:id = msg_send![task, hash];
+                  let ctns:bool = msg_send![stoppedtasks, containsObject: hash];
                   if !ctns {
-                    let _: () = msg_send![stoppedtaskslock, unlock];
                     let content = sent_response.body();
                     // default: application/octet-stream, but should be provided by the client
                     let wanted_mime = sent_response.headers().get(CONTENT_TYPE);
@@ -289,9 +289,9 @@ impl InnerWebView {
                     let () = msg_send![task, didFinish];
                   } else {
                     // Remove stopped task
-                    let _: () = msg_send![stoppedtasks, removeObject: task];
-                    let _: () = msg_send![stoppedtaskslock, unlock];
-                  }  
+                    let _: () = msg_send![stoppedtasks, removeObject: hash];
+                  }
+                  let _: () = msg_send![stoppedtaskslock, unlock];
                 },
               );
 
@@ -315,7 +315,8 @@ impl InnerWebView {
         let stoppedtaskslock:id = * (* this).get_ivar("stoppedtaskslock");
         let _: () = msg_send![stoppedtaskslock, lock];
         let stoppedtasks:id = * (* this).get_ivar("stoppedtasks");
-        let _: () = msg_send![stoppedtasks, addObject: task];
+        let hash:id = msg_send![task, hash];
+        let _: () = msg_send![stoppedtasks, addObject: hash];
         let _: () = msg_send![stoppedtaskslock, unlock];
       }
     }
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -240,46 +240,58 @@ impl InnerWebView {
             // Finish
             let () = msg_send![task, didFinish];
           };
-
+          let _this:*const Object = &(* this) as *const Object;
           // send response
           match http_request.body(sent_form_body) {
             Ok(final_request) => {
               let responder: Box<dyn FnOnce(HttpResponse<Cow<'static, [u8]>>)> = Box::new(
                 move |sent_response| {
-                  let content = sent_response.body();
-                  // default: application/octet-stream, but should be provided by the client
-                  let wanted_mime = sent_response.headers().get(CONTENT_TYPE);
-                  // default to 200
-                  let wanted_status_code = sent_response.status().as_u16() as i32;
-                  // default to HTTP/1.1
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing! I normally keep quiet about objc heavy macos issues/prs but since i spent the whole day on the startTask code yesterday i do have some opinions this time.

I personally would like to reduce our reliance on ivars where possible, even if that means having to use global variables (via once_cell::Lazy for example). To make this easier, i saw that others stored just the hash of the object, for example like [here](https://github.com/arnaudelub/flutter_polywebview/blob/76def884f29ac13065a538a10789b761c168c5ff/ios/Classes/CustomeSchemeHandler.swift#L44). I did not hear anything bad about this approach in this context so i assume it's safe enough.

Also, i wonder if we have to check this multiple times and not only once before it actually starts sending it. For that to make sense we'd probably have to chunk the didReceive call so maybe that's overkill 🤔I don't know how fast the didReceive call is, and how it behaves if stopTask is triggered _just before_ we call didReceive (but after the check you added) 🤔

thomastschurtschenthaler:
I made changes to use the tasks hash value in the store as you suggested, also the lock ist now released only after the call to task->didFinish.
still works fine in my tests.
Though I wonder the swift code you linked doesn't use any lock/sync for the dictionary at all..

morajabi:
I don't know what I'm doing wrong but as soon as I run tauri with this change it crashes.

thomastschurtschenthaler:
I managed to solve the sync/lock problem by sending the task response on the main thread (via NSObject.performSelectorOnMainThread).

FabianLars:
Won't this cause issues with large payloads?

thomastschurtschenthaler:
It should not as the response data payload is already provided as byte array by the custom thread.
Calls to the task like didReceiveResponse and didReceiveData seem to run on the main thread anyways - thus my issues with deadlocks when I locked the main thread in the stop_task call.
I did some tests with large data (3 parallel requests with about 3.5GB response payload) and the performance seems not to be affected.
For streaming the Responder API would probably need to provide some way to send multiple smaller chunks of data - and then call didReceiveData for each chunk.

lucasfernog:
your fix did work indeed and it was quite clever - but it looks like the objc2 migration already fixed this - I can't reproduce the crash with your main.rs file anymore on latest dev.

FabianLars:
let's close it then and revisit it if needed :)

thanks again for the PR!

PR #1215, opened 2024-04-05
fix(macos): Prevent NSExceptions when dropping a webview with registered protocols

### LINKED ISSUE #7898: [bug] Crash when closing a window `NSException`
### Describe the bug

When a window closes we get an exception
```
thread 'tokio-runtime-worker' panicked at 'Uncaught exception <NSException: 0x6000010509c0>', /Users/etorreborre/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wry-0.33.0/src/webview/wkwebview/mod.rs:238:28
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
This sometimes crashes the application right away and sometimes it crashes a bit later.

### Reproduction

You should be able to reproduce the issue by following these steps:

 1. checkout the `ockam` repository: https://github.com/build-trust/ockam
 2. run `cargo build`
 3. run `cd implementations/rust/ockam/ockam_app; cargo tauri dev -c tauri.conf.dev.json`
 4. click on `Enroll` in the desktop tray and authenticate with Github
 5. when your Ockam project is ready click on "Create a service" and enter a name for a service + an email address (you don't need to start any service on your machine)
 6. then the service name you created you appear in the menu. Click on it, and select "Share", and enter an email address + "Create"

=> this generally triggers the crash

### Expected behavior

We expect no crash at all :-).

### Platform and versions

```text
❯ cargo tauri info                                                                                                                                                                                                                                                                                                                                                                                      

[✔] Environment
    - OS: Mac OS 13.5.2 X64
    ✔ rustc: 1.71.1 (eb26296b5 2023-08-03)
    ✔ Cargo: 1.71.1 (7f1d04c00 2023-07-29)
    ✔ rustup: 1.25.2 (1980-01-01)
    ✔ Rust toolchain: 1.71.1-aarch64-apple-darwin (environment override by RUSTUP_TOOLCHAIN)
    - node: 18.14.1
    - pnpm: 8.6.5
    - yarn: 1.22.19
    - npm: 9.3.1

[-] Packages
    - @tauri-apps/api [NPM]: not installed!
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.14
```
```


### Stack trace

```text
Even with `RUST_BACKTRACE=1` there is no additional stacktrace displayed
```


### Additional context

_No response_

FabianLars:
fixes https://github.com/tauri-apps/tauri/issues/7898

This one was quiet hard to fix, i really tried to not rely on a global variable but always ended up with either nsexceptions or stack traces for invalid memory access. The problem is that when you're dropping the webview, like tauri does on window close, is that protocol handlers may still be running. In Tauri this is even more of an issue because it uses a protocol for the whole IPC implementation so even the close-requested event will cause problems here.

If the webview is dropped before the handler finished, either _webview or task (or both) will be undefined, but in a way that objc's is_null still returns false, but if you check its class it will be returned as `nil`. Relying on that unfortunetely doesn't work because it requires 2 calls to the object / class (first to get the class, second to get the class name) and the second one really often causes memory issues. Also the timing is unpredictable, sometimes it fails on the didReceive call, sometimes on the didFinish call, rarely on the didReceiveResponse call.

Since wry consumers can't know when the handler was finished (especially the responder part inside wry) i believe it would be unreasonable to expect users to make sure handlers are done before dropping the webview and instead it's something we should make sure.

----

P.S. I kinda hope someone comes up with a better solution because i'm not the biggest fan of the global webview vec myself, it's just the only thing i found that works **reliably**.

P.P.S. Also semi related to https://github.com/tauri-apps/wry/pull/1214 which is why i'm opening this pr right now and not waiting for tomorrow morning 🙃 

----

Edit: Another thing that just came to mind, which i didn't investigate is whether returning early from the function will leak memory? For example the NSData instance when we don't end up calling didReceive. Even if this leaks memory, i'm not sure how to handle that in objc so some input on that topic would be appreciated :)

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -18,6 +18,7 @@ use cocoa::{
   foundation::{NSDictionary, NSFastEnumeration, NSInteger},
 };
 use dpi::{LogicalPosition, LogicalSize};
+use once_cell::sync::Lazy;
 use raw_window_handle::{HasWindowHandle, RawWindowHandle};
 
 use std::{
@@ -71,6 +72,8 @@ const ACCEPT_FIRST_MOUSE: &str = "accept_first_mouse";
 
 const NS_JSON_WRITING_FRAGMENTS_ALLOWED: u64 = 4;
 
+static WEBVIEW_IDS: Lazy<Mutex<Vec<usize>>> = Lazy::new(Default::default);
+
 pub(crate) struct InnerWebView {
   pub webview: id,
   pub manager: id,
@@ -86,6 +89,7 @@ pub(crate) struct InnerWebView {
   drag_drop_ptr: *mut Box<dyn Fn(crate::DragDropEvent) -> bool>,
   download_delegate: id,
   protocol_ptrs: Vec<*mut Box<dyn Fn(Request<Vec<u8>>, RequestAsyncResponder)>>,
+  webview_id: usize,
 }
 
 impl InnerWebView {
@@ -172,6 +176,8 @@ impl InnerWebView {
         #[cfg(feature = "tracing")]
         let span = tracing::info_span!("wry::custom_protocol::handle", uri = tracing::field::Empty)
           .entered();
+        let webview_id = *this.get_ivar::<usize>("webview_id");
+
         let function = this.get_ivar::<*mut c_void>("function");
         if !function.is_null() {
           let function =
@@ -271,14 +277,25 @@ impl InnerWebView {
 
                   let urlresponse: id = msg_send![class!(NSHTTPURLResponse), alloc];
                   let response: id = msg_send![urlresponse, initWithURL:url statusCode: wanted_status_code HTTPVersion:NSString::new(&wanted_version) headerFields:headers];
+                  if !WEBVIEW_IDS.lock().unwrap().contains(&webview_id) {
+                    return;
+                  }
                   let () = msg_send![task, didReceiveResponse: response];
 
                   // Send data
                   let bytes = content.as_ptr() as *mut c_void;
                   let data: id = msg_send![class!(NSData), alloc];
                   let data: id = msg_send![data, initWithBytesNoCopy:bytes length:content.len() freeWhenDone: if content.len() == 0 { NO } else { YES }];
+
+                  if !WEBVIEW_IDS.lock().unwrap().contains(&webview_id) {
+                    return;
+                  }
                   let () = msg_send![task, didReceiveData: data];
+
                   // Finish
+                  if !WEBVIEW_IDS.lock().unwrap().contains(&webview_id) {
+                    return;
+                  }
                   let () = msg_send![task, didFinish];
                 },
               );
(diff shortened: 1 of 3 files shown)


src/wkwebview/mod.rs
```diff
@@ -71,6 +72,8 @@ const ACCEPT_FIRST_MOUSE: &str = "accept_first_mouse";
 
 const NS_JSON_WRITING_FRAGMENTS_ALLOWED: u64 = 4;
 
+static WEBVIEW_IDS: Lazy<Mutex<Vec<usize>>> = Lazy::new(Default::default);
```
pewsheen: I'm okay with the global variable approach. Just slightly concerned that the ID will conflict when removing and adding a webview immediately. The ID may be the same.

I remember that muda has a similar way of dealing with the ID in a global variable: https://github.com/tauri-apps/muda/blob/dev/src/util.rs

Maybe using a Counter + HashSet will be better?

pewsheen:
Is #1214 handling the same issue? I would prefer the fix here and just check if the webview_id is still on the running list.

src/wkwebview/mod.rs
FabianLars: i don't think this should be possible but better safe than sorry, i'll copy the counter.

FabianLars:
> Is https://github.com/tauri-apps/wry/pull/1214 handling the same issue? I would prefer the fix here and just check if the webview_id is still on the running list.

No. My PR fixes a panic/mem issue when the InnerWebview struct is dropped while protocols are registered.

1214 fixes an issue when the dev aborts a request to a protocol via AbortHandler for example (a page reload may causes this too). This will make wkwebview send the stopTask event which is **not triggered** for the issue i'm trying to fix here.
For 1214 i do still think we should rustify it a bit more and use Lazy (like i said in the comment over there), if we end up using the PR.

src/wkwebview/mod.rs
```diff
@@ -1245,6 +1275,11 @@ pub fn platform_webview_version() -> Result<String> {
 
 impl Drop for InnerWebView {
   fn drop(&mut self) {
+    WEBVIEW_IDS
+      .lock()
+      .unwrap()
+      .retain(|i| i != &self.webview_id);
```
pewsheen: Can it just `.remove(i)` ?

FabianLars: yes, i missed that one

FabianLars:
don't mind the last commits, i'm so done for today 😂

amrbashir (commented):
I think the reason is `is_null` still returns false is that because it is not set to `std::ptr::null`, I have seen a similar issue in https://github.com/tauri-apps/wry/pull/1206

I think we should fix it like this:
1. pass the webview pointer to the responder closure
2. when dropping the webview, we explicitly set the value at the webview ptr address to a null ptr or just a unit (`()``
3. check inside the responder if the value pointed by the webview ptr is null

FabianLars:
Thank you, i'll see if i can make that work.

FabianLars:
Hmm, i may be doing something wrong, but i can't make it work like this. It also looks like the webview in the start_task and the one in drop point to different addresses.

amrbashir:
cc @wusyong

lucasfernog (approved):

wusyong:
@lucasfernog Are you sure it's working? I believe @FabianLars mentioned it didn't work as intended.

FabianLars:
The PR's current implementation works for me.
I just couldn't make Amr's suggestion work (didn't push anything to the branch to not break the PR).

wusyong:
Ah sorry for the misunderstanding.

PR #1217, opened 2024-04-09
fix: use TARGET env var instead of cfg_aliases so cross compiling works

lucasfernog:
I've been facing random issues when cross compiling wry. When running `cargo xwin build --target x86_64-pc-windows-msvc`, sometimes the build script enables the gtk alias. So I thought its better to detect the target platform manually with the TARGET env var.

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -112,7 +112,6 @@ impl InnerWebView {
     pl_attrs: super::PlatformSpecificWebViewAttributes,
     _web_context: Option<&mut WebContext>,
   ) -> Result<Self> {
-    println!("N");
     let ns_view = match window.window_handle()?.as_raw() {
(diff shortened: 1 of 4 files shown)


amrbashir (dismissed):

wusyong:
android target will fail to build it seems.

lucasfernog:
fix pushed

src/wkwebview/mod.rs
```diff
@@ -112,6 +112,7 @@ impl InnerWebView {
     pl_attrs: super::PlatformSpecificWebViewAttributes,
     _web_context: Option<&mut WebContext>,
   ) -> Result<Self> {
+    println!("N");
```
pewsheen: 👀

amrbashir (approved):

PR #1226, opened 2024-04-19
feat(macos): add option to set a DataStoreIdentifier for WKWebView

santikid:
Both WebView2 for Windows and WebkitGTK support setting a custom data directory path to store cookies, IndexedDB and other data (set in ```src/web_context.rs -> WebContextData```). On macOS, WKWebView does not support this so setting data_directory in wry or tauri has no effect.

This PR adds the ability to specify a UUID that is used to call [dataStoreForIdentifier](https://developer.apple.com/documentation/webkit/wkwebsitedatastore/4183560-datastoreforidentifier?language=objc), which creates a data directory in ~/Library/WebKit/[application]/WebsiteDataStore/[UUID] instead of using the default ~/Library/WebKit/[application]/WebsiteData .

Discussion: #1198 

tauri-apps/tauri#8637

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1032,6 +1032,36 @@ impl<'a> WebViewBuilder<'a> {
   }
 }
 
+#[cfg(target_os = "macos")]
+#[derive(Clone)]
+pub(crate) struct PlatformSpecificWebViewAttributes {
+  data_store_identifier: Option<[u8; 16]>,
+}
+
+#[cfg(target_os = "macos")]
+impl Default for PlatformSpecificWebViewAttributes {
+  fn default() -> Self {
+    Self {
+      data_store_identifier: None,
+    }
+  }
+}
+
+#[cfg(target_os = "macos")]
+pub trait WebViewBuilderExtMacOS {
+  /// Initialize the WebView with a custom data store identifier. This can be used as a replacement for
+  /// data_directory not being available in WKWebView.
+  fn with_data_store_identifier(self, identifier: [u8; 16]) -> Self;
+}
+
+#[cfg(target_os = "macos")]
+impl WebViewBuilderExtMacOS for WebViewBuilder<'_> {
+  fn with_data_store_identifier(mut self, identifier: [u8; 16]) -> Self {
+    self.platform_specific.data_store_identifier = Some(identifier);
+    self
+  }
+}
+
 #[cfg(windows)]
 #[derive(Clone)]
 pub(crate) struct PlatformSpecificWebViewAttributes {
@@ -1654,7 +1684,6 @@ pub enum PageLoadEvent {
   target_os = "netbsd",
   target_os = "openbsd",
   target_os = "ios",
-  target_os = "macos",
 ))]
 #[derive(Default)]
 pub(crate) struct PlatformSpecificWebViewAttributes;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1049,8 +1049,10 @@ impl Default for PlatformSpecificWebViewAttributes {
 
 #[cfg(any(target_os = "macos", target_os = "ios",))]
 pub trait WebViewBuilderExtDarwin {
-  /// Initialize the WebView with a custom data store identifier. This can be used as a replacement for
-  /// data_directory not being available in WKWebView on macOS and iOS.
+  /// Initialize the WebView with a custom data store identifier.
+  /// Can be used as a replacement for data_directory not being available in WKWebView.
+  ///
+  /// Only available on macOS >= 14 and iOS >= 17
   fn with_data_store_identifier(self, identifier: [u8; 16]) -> Self;
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1032,13 +1032,13 @@ impl<'a> WebViewBuilder<'a> {
   }
 }
 
-#[cfg(target_os = "macos")]
+#[cfg(any(target_os = "macos", target_os = "ios",))]
 #[derive(Clone)]
 pub(crate) struct PlatformSpecificWebViewAttributes {
   data_store_identifier: Option<[u8; 16]>,
 }
 
-#[cfg(target_os = "macos")]
+#[cfg(any(target_os = "macos", target_os = "ios",))]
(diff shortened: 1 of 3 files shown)


santikid:
converted to draft because this needs to be added to/tested on iOS as well

santikid:
just tried on iOS with the newest tauri beta, seems to work well

amrbashir (commented):
Instead of adding a new option just for macOS, can't we reuse the data_direcotry passed to the WebContext and hash it into an identifier?

santikid:
It may be a little confusing if a data_directory is specified and there is nothing at the given path, but since as you said this is only an issue on macOS that's probably a good trade-off.

Do you think this can be done without adding any external crates for hashing? The built-in std::hash::DefaultHasher is [not guaranteed  to stay consistent over releases.](https://doc.rust-lang.org/std/hash/struct.DefaultHasher.html)

I guess some non-cryptographic hash implementation could be vendored?

amrbashir:
I see, maybe it is fine to just add a separate option as you proposed and downstream tauri could choose to hash the data_directory and reuse it.

pewsheen:
[`datastoreforidentifier`](https://developer.apple.com/documentation/webkit/wkwebsitedatastore/4183560-datastoreforidentifier?language=objc) seems support macOS 14+ and iOS 17+ only. Should we do a version check or add the version limitation to the document?

cc @FabianLars

FabianLars:
if it crashes on earlier macos version then imo ideally both, a version check and documentation.

santikid:
It does crash on macOS < 14. For macOS, a version check could be done with `NSAppKitVersionNumber` in cocoa::appkit, but I couldn't figure out how to do it on iOS.

Alternatively, we could just check for the initWithIdentifier method in [_WKWebsiteDataStoreConfiguration.h](https://github.com/WebKit/WebKit/blob/fd62a381b01ca67d2d9f159a3858e67c32d9f381/Source/WebKit/UIProcess/API/Cocoa/_WKWebsiteDataStoreConfiguration.h). I don't fully get why it doesn't exist on WKWebsiteDataStore though.

```rust
let class = class!(_WKWebsiteDataStoreConfiguration);
println!("{:?}", class.instance_methods().iter().map(|x| x.name()).collect::<Vec<_>>());
let method = class.instance_method(Sel::register("initWithIdentifier:"));
println!("initWithIdentifier available: {}", method.is_some());
```

Would appreciate some input on this.

**EDIT:** Turns out the above snippet works on macOS 13.0, returning false for initWithIdentifier.is_some, but is broken on 13.6, where it prints true even though the API isn't there yet.

pewsheen:
> It does crash on macOS < 14. For macOS, a version check could be done with `NSAppKitVersionNumber` in cocoa::appkit, but I couldn't figure out how to do it on iOS.
> 
> Alternatively, we could just check for the initWithIdentifier method in [_WKWebsiteDataStoreConfiguration.h](https://github.com/WebKit/WebKit/blob/fd62a381b01ca67d2d9f159a3858e67c32d9f381/Source/WebKit/UIProcess/API/Cocoa/_WKWebsiteDataStoreConfiguration.h). I don't fully get why it doesn't exist on WKWebsiteDataStore though.
> 
> ```rust
> let class = class!(_WKWebsiteDataStoreConfiguration);
> println!("{:?}", class.instance_methods().iter().map(|x| x.name()).collect::<Vec<_>>());
> let method = class.instance_method(Sel::register("initWithIdentifier:"));
> println!("initWithIdentifier available: {}", method.is_some());
> ```
> 
> Would appreciate some input on this.
> 
> **EDIT:** Turns out the above snippet works on macOS 13.0, returning false for initWithIdentifier.is_some, but is broken on 13.6, where it prints true even though the API isn't there yet.

Thanks for the testing! It's so confusing that the API existed, but you can't use it :\

I think we could use [`NSOperatingSystemVersion`](https://developer.apple.com/documentation/foundation/nsoperatingsystemversion) and combine `#[cfg(target_os = "ios")]`, `#[cfg(target_os = "macos")]` to check iOS and macOS version

This is a code snippet from another PR:
```rs
pub fn operating_system_version() -> (u64, u64, u64) {
   unsafe {
     let process_info: id = msg_send![class!(NSProcessInfo), processInfo];
     let version: NSOperatingSystemVersion = msg_send![process_info, operatingSystemVersion];
     (
       version.majorVersion,
       version.minorVersion,
       version.patchVersion,
     )
   }
 }

#[cfg(target_os = "macos")]
if operating_system_version().0 >= 14 {
  todo!();
}
#[cfg(target_os = "ios")]
if operating_system_version().0 >= 17 {
  todo!();
}
```

santikid:
Thanks @pewsheen !

>  It's so confusing that the API existed, but you can't use it :\

My understanding of "darwin" APIs / objc is pretty limited, but this does seem very strange. 😅

src/wkwebview/util.rs
```diff
@@ -16,3 +17,16 @@ impl Counter {
     self.0.fetch_add(1, Ordering::Relaxed)
   }
 }
+
+#[cfg(any(target_os = "macos", target_os = "ios"))]
```
pewsheen:
```suggestion
```

We didn't public `util` mod outside wkwebview, it should be safe to remove this

src/lib.rs
```diff
@@ -1032,6 +1032,38 @@ impl<'a> WebViewBuilder<'a> {
...
+    }
+  }
+}
+
+#[cfg(any(target_os = "macos", target_os = "ios",))]
+pub trait WebViewBuilderExtDarwin {
+  /// Initialize the WebView with a custom data store identifier.
+  /// Can be used as a replacement for data_directory not being available in WKWebView.
+  ///
+  /// Only available on macOS >= 14 and iOS >= 17
```
pewsheen:
```suggestion
  /// - **macOS / iOS**: Available on macOS >= 14 and iOS >= 17
```

Just follow the other document's style

pewsheen (commented):
It looks good to me, just some small suggestions.

cc @wusyong

pewsheen (approved):

FabianLars (approved):

PR #1251, opened 2024-05-09
disable initialization script injection into subframes on macOS

chippers:
This matches the current behavior of Linux and (mostly) Windows. Windows differs slightly since it still injects subscripts to subframes on the same origin as the main frame.


tweidinger (approved):

pewsheen (approved):

pewsheen:
@chippers Do we need to patch `dev` branch as well?

PR #1258, opened 2024-05-12
feat: Add screenshot functionality

atlanticaccent:
Based on work in PR #226 - adds screenshot functionality to macOS, linux and Windows. Android is unsupported.

Compared to 226, this PR does not include the region selection functionality. [This comment in the Webview2 repo](https://github.com/MicrosoftEdge/WebView2Feedback/issues/529#issuecomment-1021262814) indicates that it may now be possible to have parity across all three platforms in selecting regions, and so make it worthwhile to re-add.

This PR does not (yet) improve on the API presented in 226.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,8 @@ os-webview = [
   "dep:gtk",
   "soup3",
   "x11-dl",
-  "gdkx11"
+  "gdkx11",
+  "cairo-rs"
 ]
 tracing = [ "dep:tracing" ]
 
@@ -59,6 +60,7 @@ javascriptcore-rs = { version = "=1.1.2", features = [ "v2_28" ], optional = tru
 webkit2gtk = { version = "=2.0.1", features = [ "v2_38" ], optional = true }
 webkit2gtk-sys = { version = "=2.0.1", optional = true }
 gtk = { version = "0.18", optional = true }
+cairo-rs = { version = "0.18", features = ["png"], optional = true }
 soup3 = { version = "0.5", optional = true }
 x11-dl = { version = "2.21", optional = true }
 gdkx11 = { version = "0.18", optional = true }
--- a/examples/screenshot.rs
+++ b/examples/screenshot.rs
@@ -20,7 +20,14 @@ fn main() -> wry::Result<()> {
     .build(&event_loop)
     .unwrap();
   // Build the webview
-  let webview = WebViewBuilder::new(&window)
+  let webview = WebViewBuilder::new_as_child(&window)
+    .with_bounds(wry::Rect {
+      position: dpi::Position::Logical(dpi::LogicalPosition { x: 50.0, y: 50.0 }),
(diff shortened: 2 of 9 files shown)


andrewjanssen:
I'm getting an `EXC_BAD_ACCESS` error when I call `screenshot`, on the following line. MacOS 15.0.1.

```
let image: id = msg_send![ns_image, CGImageForProposedRect:nil context:nil hints:nil];
```

This is crashing on `A::invoke` in:

```
pub unsafe fn send_unverified<T, A, R>(obj: *const T, sel: Sel, args: A) -> Result<R, MessageError>
where T: Message, A: MessageArguments, R: Any {
    let receiver = obj as *mut T as *mut Object;
    let msg_send_fn = msg_send_fn::<R>();
    objc_try!({
        A::invoke(msg_send_fn, receiver, sel, args)
    })
}
```

`deprecated_screenshot` runs successfully for me but the WebView doesn't appear in the screenshot: the image contains every window except the WebView. This is fixed by using only the `kCGWindowListOptionIncludingWindow` option and removing `kCGWindowListOptionOnScreenAboveWindow` and `kCGWindowListExcludeDesktopElements`. (It does capture the correct region, even after the WebView window is moved.)

Thanks for your work here!

atlanticaccent:
> I'm getting an `EXC_BAD_ACCESS` error when I call `screenshot`, on the following line. MacOS 15.0.1.
> 
> ```
> let image: id = msg_send![ns_image, CGImageForProposedRect:nil context:nil hints:nil];
> ```
> 
> This is crashing on `A::invoke` in:
> 
> ```
> pub unsafe fn send_unverified<T, A, R>(obj: *const T, sel: Sel, args: A) -> Result<R, MessageError>
> where T: Message, A: MessageArguments, R: Any {
>     let receiver = obj as *mut T as *mut Object;
>     let msg_send_fn = msg_send_fn::<R>();
>     objc_try!({
>         A::invoke(msg_send_fn, receiver, sel, args)
>     })
> }
> ```
> 
> `deprecated_screenshot` runs successfully for me but the WebView doesn't appear in the screenshot: the image contains every window except the WebView. This is fixed by using only the `kCGWindowListOptionIncludingWindow` option and removing `kCGWindowListOptionOnScreenAboveWindow` and `kCGWindowListExcludeDesktopElements`. (It does capture the correct region, even after the WebView window is moved.)
> 
> Thanks for your work here!

Hi! Yes, I pretty much gave up on the macOS implementation and in my own project currently have it disabled entirely. I believe I did have the "newer" version working at some point and lost it, or it might actually work right now but require granting the running application an Accessibility permission so it can capture properly - possibly both.

Either way, the performance is also catastrophic using either method, which further contributed to me abandoning the work. 

I've since had some ideas about alternatives (grabbing the window handle and using something like `wgpu` to render the window texture to a usable image format), so I may come back to this PR if they work out/people are still interested in this.

PR #1259, opened 2024-05-12
feat: add MacOS print options

### LINKED ISSUE #713: MacOS: window.print() margins and dialog options differ
**Describe the bug**
When printing using the window.print() method in the context of MacOS, page margins as well as dialog options differ from Safari. Margins are significantly larger and paper size and orientation options are absent.

**Steps To Reproduce**
Initialise Tauri Vanilla app, call window.print() method.

**Expected behavior**
Default margins and all options available.

**Screenshots**
image
image

OS: macOS 12.6
Rustc: 1.57

- [ ] Yes
- [x] No

**Additional context**
The @page rule is not implemented with Safari. Setting the printable area to "position: fixed;" and "top: 0; right: 0; left: 0;" does not influence the margins. Using negative inset will result in cutting the content off.

augustoccesar:
### Description
Currently, when printing on a MacOS, it defaults to a quite big margin, which does not reflect the side of it when printing on the web. The current code just creates a new NSPrintInfo from the default, which for some reason has a large margin.

This changes so that the default margin is 0 and add the option to change it when printing the webview on MacOs. This is not a field that can be changed by the user on the appkit modal, so it has to be defined before the modal is open.

Other options could be added on top of these ones, but for this initial commit/PR decided to go with adding the margins since it seems to be the one with bigger discrepancy compared to the web version.

Closes #713 (Not sure if should mark this as really close it since it really fixes only the margin part. The difference on the modal doesn't seem to be customisable as far as I understood from the docs).

### References:
- https://developer.apple.com/documentation/appkit/nsprintinfo

### Screenshots:
Using the new `print.rs` in the examples folder.


pewsheen:
Hi, can you add a changelog (https://github.com/tauri-apps/wry/tree/dev/.changes) and do a `cargo fmt`?

augustoccesar:
> Hi, can you add a changelog (https://github.com/tauri-apps/wry/tree/dev/.changes) and do a `cargo fmt`?

Done! ✅ 
Put it as a `minor`, but let me know if that was classified incorrectly for the project.
It does fix what I believe to be a bug (the large margins when printing on MacOS), but also adds a new feature (the printing with extra options), so the `minor` felt more correct.

examples/print.rs
amrbashir, line 1: no need to add a print example, let's remove

.changes/macos_print_options.md
```diff
@@ -0,0 +1,8 @@
+---
+"wry": minor
+---
+
+Default the margin when printing on MacOS to 0 so it is closer to the behavior
+of when printing on the web. It also add a new function to WebViewExtMacOS
+called print_with_options which allows the user to modify the margins that
+will be sent down to the AppKit print operation (NSPrintInfo).
```
amrbashir: let's split these into two separate change files

src/wkwebview/mod.rs
```diff
@@ -79,6 +81,30 @@ const NS_JSON_WRITING_FRAGMENTS_ALLOWED: u64 = 4;
...
+impl Default for PrintMargin {
+  fn default() -> Self {
+    PrintMargin {
+      top: 0.0,
+      right: 0.0,
+      bottom: 0.0,
+      left: 0.0,
+    }
+  }
+}
```
amrbashir: just derive, also derive Copy and Clone
```suggestion
#[derive(Debug, Default, Copy, Clone)]
pub struct PrintMargin {
  pub top: f32,
  pub right: f32,
  pub bottom: f32,
  pub left: f32,
}
```

src/wkwebview/mod.rs
```diff
@@ -79,6 +81,30 @@ const NS_JSON_WRITING_FRAGMENTS_ALLOWED: u64 = 4;
...
+
+#[derive(Debug, Default)]
```
amrbashir:
```suggestion
#[derive(Debug, Default, Clone)]
```

amrbashir (changes requested):

examples/print.rs
augustoccesar: Changed on 921d297ecf288bfbfdeb2d09c338cd845fa0bc6d

.changes/macos_print_options.md
augustoccesar: Changed on 2a08ba857c5221c1c198f49c6762c9be27d65d86

src/wkwebview/mod.rs
augustoccesar: Changed on ca1ac37fc3ef8d4c4b31fe89dc0c22c56caed710

src/wkwebview/mod.rs
augustoccesar: Changed on 3868f6522fb6d7f8ba07b644470e642a4e34b447

.changes/macos_default_print_margins.md
```diff
@@ -0,0 +1,6 @@
+---
+"wry": minor
+---
+
+Default the margin when printing on MacOS to 0 so it is closer to the behavior
+of when printing on the web.
```
amrbashir:
```suggestion
Default the margin when printing on MacOS to 0 so it is closer to the behavior of when printing on the web.
```

.changes/macos_print_margin_options.md
```diff
@@ -0,0 +1,6 @@
+---
+"wry": minor
+---
+
+Add a new function to WebViewExtMacOS called print_with_options which
+allows the user to modify the margins that will be sent down to the AppKit print operation (NSPrintInfo).
```
amrbashir:
```suggestion
Add `WebViewExtMacOS::print_with_options` which allows to modify the margins that will be used on the print dialog.
```

amrbashir (approved):
Thank you

PR #1276, opened 2024-05-22
fix(linux): async custom protocol crashing sending response

lucasfernog:
Fixes `GTK may only be used from the main thread. thread 'tokio-runtime-worker' panicked at .cargo/registry/src/index.crates.io-6f17d22bba15001f/webkit2gtk-2.0.1/src/auto/uri_scheme_response.rs:26`

Ref https://github.com/tauri-apps/tauri/issues/9794

### DIFF
--- a/src/webkitgtk/web_context.rs
+++ b/src/webkitgtk/web_context.rs
@@ -21,7 +21,7 @@ use std::{
 };
 use webkit2gtk::{
   ApplicationInfo, AutomationSessionExt, CookiePersistentStorage, DownloadExt, LoadEvent,
-  SecurityManagerExt, URIRequest, URIRequestExt, URISchemeRequestExt, URISchemeResponse,
+  SecurityManagerExt, URIRequest, URIRequestExt, URISchemeRequestExt, URISchemeResponse, URISchemeRequest,
   URISchemeResponseExt, UserContentManager, WebContext, WebContextExt as Webkit2gtkContextExt,
   WebView, WebViewExt,
 };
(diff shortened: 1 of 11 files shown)


src/webkitgtk/web_context.rs
```diff
@@ -361,28 +361,33 @@ where
...
-            .headers()
-            .get(CONTENT_TYPE)
-            .and_then(|h| h.to_str().ok());
-
-          let response = URISchemeResponse::new(&input, buffer.len() as i64);
-          response.set_status(http_response.status().as_u16() as u32, None);
-          if let Some(content_type) = content_type {
-            response.set_content_type(content_type);
-          }
+          gtk::glib::idle_add(move || {
```
amrbashir: should we use [`MainContext::invoke`](https://docs.rs/glib/0.18.5/glib/struct.MainContext.html#method.invoke) instead? seems like `idle_add` might wait until the loop is idle which may impact performance, also `MainContext::invoke` will call the closure directly if it is already on the main thread.

lucasfernog: oh i'll try that

lucasfernog: sorted @amrbashir

lucasfernog: i knew that function existed - just couldn't find it in tao/wry

amrbashir: > sorted @amrbashir

did you forget to push :joy:?

lucasfernog: 🤣🤣🤣🤣 in my defense i was testing this while holding a baby

amrbashir: no worries, anything for the baby ❤️

lucasfernog: finally pushed

amrbashir (approved):

PR #1279, opened 2024-05-29
fix(macos): suppress incorrect `dead_code` warning from rustc 1.78

rhysd:
This PR suppresses the following warning reported by `cargo check`. From 1.78, rustc enhanced the dead code check and it wrongly detected the unused field. The field is actually used from Objective-C world.

```
warning: field `0` is never read
    --> src/wkwebview/mod.rs:1441:15
     |
1441 | struct NSData(id);
     |        ------ ^^
     |        |
     |        field in this struct
     |
     = note: `#[warn(dead_code)]` on by default
help: consider changing the field to be of unit type to suppress this warning while preserving the field numbering, or remove the field
     |
1441 | struct NSData(());
     |               ~~
```

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -1438,6 +1438,7 @@ impl From<NSData> for NSString {
   }
 }
 
+#[allow(dead_code)] // rustc complains `id` is unused, but it is actually used from Objective-C
 struct NSData(id);
 
 /// Converts from wry screen-coordinates to macOS screen-coordinates.


.changes/dead-code-warning.md
```diff
@@ -0,0 +1,5 @@
+---
```
amrbashir: I think the warning is only emitted when developing this crate, so no need to add a change file

amrbashir (changes requested):

.changes/dead-code-warning.md
rhysd: Thanks. I confirmed that this warning was not reported when it was used as dependency. I will remove it.

rhysd: Removed: fdee0a2659f8f7b5e540e629b6324624657c304c

amrbashir (approved):

PR #1284, opened 2024-06-04
Revert "fix(macos): check WKURLSchemeTask is valid before using"

wusyong:
Reverts tauri-apps/wry#1282

Didn't realize there's another slimilar one in discussion.

### DIFF
--- a/.changes/fix-macos-mitigate-async-command-panic.md
+++ b/.changes/fix-macos-mitigate-async-command-panic.md
@@ -1,5 +1,0 @@
----
-"wry": patch
----
-
-On macOS, mitigate an issue that could cause a panic when running an async command.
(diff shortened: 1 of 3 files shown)


jason-crabnebula (approved):

pewsheen (approved):

FabianLars:
to be fair though, imo jason's solution was better because it didn't use so much objc 😂

pewsheen:
> to be fair though, imo jason's solution was better because it didn't use so much objc 😂

I was about to ask for your opinion lol

This still crashes in some circumstances, so we want to revert it first and try something else.

FabianLars:
hmm, that's unfortunate. Did you check if the community pr also crashes in the same context?

pewsheen:
> hmm, that's unfortunate. Did you check if the community pr also crashes in the same context?

Community PR works, and I was testing the example that PR provided.
It looks like the `task` will be released even inside the didResponse method. Retain `task` until the response closure ends would fix it. I'll reopen the PR again with the fix.

update:
Community PR does not work for the issue on tauri

PR #1286, opened 2024-06-04
feat: Added visionOS support for Apple Vision Pro

eugenehp:
Based on Added visionOS support based on https://github.com/rust-lang/rust/pull/121419

Preliminary work to add visionOS support in Tauri https://github.com/tauri-apps/tauri/issues/9969


eugenehp:
Ok, looks like CI needs some love. Going to look into it a bit later.

PR #1291, opened 2024-06-11
chore(ci): run tests on Intel Mac

rhysd:
`macos-latest` runner is now run on arm64 macOS. (`macos-latest` and `macos-14` now mean the same runner.)

https://github.com/actions/runner-images?tab=readme-ov-file#available-images

So the build workflow of this repository is now running only arm64 macOS and no tests are run on x86_64 macOS (Intel Mac).

This PR modifies the build workflow to run tests on both arm64 and x86_64 macOS as follows:

target v.s. host arch

| | aarch64-apple-darwin | x86_64-apple-darwin |
|-|-|-|
| aarch64 | ✅ | ✅ |
| x86_64 | :x: | ✅ |

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
           - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
           - { target: x86_64-apple-darwin, os: macos-latest }
           - { target: aarch64-apple-ios, os: macos-latest }
-          - { target: aarch64-apple-darwin, os: macos-14 }
+          - { target: x86_64-apple-darwin, os: macos-12 }
           - { target: aarch64-linux-android, os: ubuntu-latest }
 
     runs-on: ${{ matrix.platform.os }}


.github/workflows/build.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
           - { target: x86_64-apple-darwin, os: macos-latest }
           - { target: aarch64-apple-ios, os: macos-latest }
-          - { target: aarch64-apple-darwin, os: macos-14 }
+          - { target: x86_64-apple-darwin, os: macos-12 }
```
amrbashir, line 19: Why not `macos-13`? also can't we cross-compile x86_64 on macos-14?

FabianLars: > also can't we cross-compile x86_64 on macos-14?

That's what we are doing on line 17 so technically with this PR we aren't building _for_ arm anymore

rhysd: > Why not macos-13? 

Because older macOS is better in terms of checking compatibility. If my understanding is correct, x86_64 binary built on macOS 12 is available on both macOS 12 and 13, but the one built on macOS 13 is only available on macOS 13.

> can't we cross-compile x86_64 on macos-14?

Yes, we can do. And actually the combination of x86_64 target and arm64 host is tested as I showed in above table. However, it doesn't mean this crate can be built on x86_64 host. So checking it with macos-12 (x86_64 host) runner is necessary.

amrbashir (approved):

.github/workflows/build.yml
FabianLars: > but the one built on macOS 13 is only available on macOS 13.

Nope, that's a Linux problem. On macOS this doesn't matter.

> Yes, we can do. And actually the combination of x86_64 target and arm64 host is tested as I showed in above table. 

Yeah, but the arm64 _target_ isn't tested anymore now...

> However, it doesn't mean this crate can be built on x86_64 host.

Honestly i doubt that this is much of an issue on macOS, imo compiling for the target is good enough, but let's keep the x86 host until github removes it.

rhysd: > Nope, that's a Linux problem. On macOS this doesn't matter.

Ah, thanks. I think that's my misunderstanding.

> Yeah, but the arm64 target isn't tested anymore now...

The combination is not tested, but I believe compiling aarch64 binary on x86_64 host is rare case so I didn't include it in the matrix. If you think it would be better to include, I can happily make another PR.

> Honestly i doubt that this is much of an issue on macOS, imo compiling for the target is good enough, but let's keep the x86 host until github removes it.

Yeah, I didn't encounter any problem specific to x86_64. However x86_64 rustc and aarch64 rustc are completely different binaries so I feel that testing both is a good idea.

PR #1296, opened 2024-06-17
chore(deps): Upgrade to `ndk 0.9` and delete unused `ndk-sys`/`ndk-context` dependencies

MarijnS95:
Closes #1060
Closes #1061

The `ndk` crate received some marginal API upgrades, besides fixing soundness issues.  Specifcally, `ForeignLooper::add_fd_with_callback()` now signifies that the incoming file descriptor is a `BorrowedFd` and the callback is executed on a different thread (where the looper is polled on the `ThreadLooper`) and must hence be `Send.`.

It appears the `ndk-sys` and `ndk-context` crates are not used directly, hence they are removed from `Cargo.toml` here.

### DIFF
--- a/.changes/ndk-0.9.md
+++ b/.changes/ndk-0.9.md
@@ -1,0 +1,6 @@
+---
+"wry": minor
+---
+
+**Breaking change**: Upgrade `ndk` crate to `0.9` and delete unused `ndk-sys` and `ndk-context` dependencies.  Types from the `ndk` crate are used in public API surface.
+**Breaking change**: The public `android_setup()` function now takes `&ThreadLooper` instead of `&ForeignLooper`, signifying that the setup function must be called on the thread where the looper is attached (and the `JNIEnv` argument is already thread-local as well).
--- a/src/android/main_pipe.rs
+++ b/src/android/main_pipe.rs
@@ -15,10 +15,10 @@ use std::{os::unix::prelude::*, sync::atomic::Ordering};
 use super::{find_class, EvalCallback, EVAL_CALLBACKS, EVAL_ID_GENERATOR, PACKAGE};
 
 static CHANNEL: Lazy<(Sender<WebViewMessage>, Receiver<WebViewMessage>)> = Lazy::new(|| bounded(8));
-pub static MAIN_PIPE: Lazy<[RawFd; 2]> = Lazy::new(|| {
+pub static MAIN_PIPE: Lazy<[OwnedFd; 2]> = Lazy::new(|| {
   let mut pipe: [RawFd; 2] = Default::default();
   unsafe { libc::pipe(pipe.as_mut_ptr()) };
-  pipe
+  unsafe { pipe.map(|fd| OwnedFd::from_raw_fd(fd)) }
 });
 
 pub struct MainPipe<'a> {
@@ -32,7 +32,13 @@ impl<'a> MainPipe<'a> {
   pub(crate) fn send(message: WebViewMessage) {
     let size = std::mem::size_of::<bool>();
     if let Ok(()) = CHANNEL.0.send(message) {
-      unsafe { libc::write(MAIN_PIPE[1], &true as *const _ as *const _, size) };
+      unsafe {
+        libc::write(
+          MAIN_PIPE[1].as_raw_fd(),
+          &true as *const _ as *const _,
+          size,
+        )
+      };
     }
   }
 
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -17,13 +17,14 @@ use jni::{
   JNIEnv,
 };
 use kuchiki::NodeRef;
-use ndk::looper::{FdEvent, ForeignLooper};
+use ndk::looper::{FdEvent, ThreadLooper};
 use once_cell::sync::OnceCell;
 use raw_window_handle::HasWindowHandle;
 use sha2::{Digest, Sha256};
 use std::{
   borrow::Cow,
   collections::HashMap,
+  os::fd::{AsFd as _, AsRawFd as _},
   sync::{atomic::AtomicI32, mpsc::channel, Mutex},
 };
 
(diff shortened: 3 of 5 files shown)


src/android/main_pipe.rs
```diff
@@ -15,10 +15,10 @@ use std::{os::unix::prelude::*, sync::atomic::Ordering};
 use super::{find_class, EvalCallback, EVAL_CALLBACKS, EVAL_ID_GENERATOR, PACKAGE};
 
 static CHANNEL: Lazy<(Sender<WebViewMessage>, Receiver<WebViewMessage>)> = Lazy::new(|| bounded(8));
-pub static MAIN_PIPE: Lazy<[RawFd; 2]> = Lazy::new(|| {
+pub static MAIN_PIPE: Lazy<[OwnedFd; 2]> = Lazy::new(|| {
   let mut pipe: [RawFd; 2] = Default::default();
   unsafe { libc::pipe(pipe.as_mut_ptr()) };
-  pipe
+  unsafe { pipe.map(|fd| OwnedFd::from_raw_fd(fd)) }
```
MarijnS95, line 21: It might be more convenient to use `rustix` here.

src/android/mod.rs
```diff
@@ -108,10 +109,10 @@ pub unsafe fn android_setup(
   };
 
   looper
-    .add_fd_with_callback(MAIN_PIPE[0], FdEvent::INPUT, move |_| {
+    .add_fd_with_callback(MAIN_PIPE[0].as_fd(), FdEvent::INPUT, move |fd, _event| {
       let size = std::mem::size_of::<bool>();
       let mut wake = false;
-      if libc::read(MAIN_PIPE[0], &mut wake as *mut _ as *mut _, size) == size as libc::ssize_t {
+      if libc::read(fd.as_raw_fd(), &mut wake as *mut _ as *mut _, size) == size as libc::ssize_t {
         main_pipe.recv().is_ok()
```
MarijnS95, line 119: Note that the compiler points out that `main_pipe` is not `Send`, so this won't compile.

amrbashir: It is probably safe to just manually implement `Send` for main_pipe, it is sent here and never accessed anywhere else. It is a workaround for now, hopefully I can find time to switch to android-activity crate instead.

src/android/main_pipe.rs
amrbashir: let's keep what works for now, I'd rather migrate to something based on android-activity crate in the future.

amrbashir (dismissed):

amrbashir (dismissed):
Thank you, just waiting for https://github.com/tauri-apps/tao/pull/955

src/android/mod.rs
MarijnS95: @amrbashir don't ever push such horrendous hacks to external contributors' PRs. I can't even refute your safety assessment because there is none to begin with.

This `Send` requirement stems from the `JNIEnv`, and the documentation clearly says **in bold** that they are thread-local and **cannot be shared between threads**:
https://developer.android.com/training/articles/perf-jni

(I hope this is not how you typically handle safety-related compiler errors in `wry`/`tao` :see_no_evil:)

MarijnS95: In any case, the correct fix is to pass the `ThreadLooper` which [does not have the `Send` bound](https://github.com/rust-mobile/ndk/pull/469). Looking through the `tao` implementation (for which I'll send an update-PR shortly - had that in the queue while resolving this PR first) it calls this `android_setup()` function on the thread where both the `JNIEnv` are and looper are local so this is the only correct thing to do.

amrbashir: > @amrbashir don't ever push such horrendous hacks to external contributors' PRs.

Apologies

> I can't even refute your safety assessment because there is none to begin with.
> 
> This `Send` requirement stems from the `JNIEnv`, and the documentation clearly says **in bold** that they are thread-local and **cannot be shared between threads**: [developer.android.com/training/articles/perf-jni](https://developer.android.com/training/articles/perf-jni)

I know that `MainPipe` is not `Send` but it is only ever accessed on the same thread and never shared between other threads so that's why it is probably safe to just slap a `unsafe impl Send for MainPipe` and as a good measure because I didn't trust my assessment fully, I made sure to test it on a real application which seemed to work fine. 


> (I hope this is not how you typically handle safety-related compiler errors in `wry`/`tao` 🙈)

As I said before, our current android implementation is all a hack and experimental, and we currently don't have the time to rewrite a better implementation so I myself just try to get by if possible until I find time to write a better implementation based on android-activity crate. 


Anyway, thanks for fixing this properly, we appreciate it.

amrbashir (changes requested):
Could you also add a change file in `.changes` directory?

.changes/ndk-0.9.md
```diff
@@ -0,0 +1,7 @@
+---
+"wry": major
```
amrbashir:
```suggestion
"wry": minor
```

amrbashir (changes requested):

amrbashir (approved):

src/android/mod.rs
MarijnS95: > but it is only ever accessed on the same thread and never shared between other threads so that's why it is probably safe to just slap a `unsafe impl Send for MainPipe` and as a good measure because I didn't trust my assessment fully, I made sure to test it on a real application which seemed to work fine.

You didn't say that _"it is only ever accessed on the same thread"_ in your original assessment. This is the kind of info that should have gone into a `// SAFETY:` comment.

> As I said before, our current android implementation is all a hack and experimental

That doesn't seem to have been stated anywhere, just that the `unsafe impl Send` was a workaround until switching to a different kind of implementation i.e. based on `android-activity`. But you're right, it doesn't take much reading of the code to realize this :wink: 

At least we managed to patch it up somewhat.

amrbashir:
thank you

PR #1299, opened 2024-06-21
fix(windows): set physical values in bounds method

### LINKED ISSUE #10053: [bug] Manual WebView resizing does not work correctly
### Describe the bug

Manually resizing the WebView(set_size) results in incorrect positioning of the WebView.
However, if the WebView coordinates are set to (0., 0.), it works correctly.

https://youtu.be/02gJWTb1Hjg

### Reproduction

https://github.com/kanatapple/tauri-manual-resize

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.19045 X64
    ✔ WebView2: 125.0.2535.92
    ✔ MSVC:
        - Visual Studio Professional 2017
        - Visual Studio Professional 2019
    ✔ rustc: 1.77.2 (25ef9e3d8 2024-04-09)
    ✔ cargo: 1.77.2 (e52e36006 2024-03-26)
    ✔ rustup: 1.27.0 (bbb9276d2 2024-03-08)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.11.0
    - yarn: 1.22.11
    - npm: 10.2.4

[-] Packages
    - tauri [RUST]: 2.0.0-beta.22
    - tauri-build [RUST]: 2.0.0-beta.17
    - wry [RUST]: 0.40.1
    - tao [RUST]: 0.28.1
    - tauri-cli [RUST]: 1.2.3
    - @tauri-apps/api [NPM]: 2.0.0-beta.13
    - @tauri-apps/cli [NPM]: 2.0.0-beta.20

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

kanatapple:
https://github.com/tauri-apps/wry/blob/f56ae29ee76611c554e1f550c2457d6431d38575/src/webview2/mod.rs#L1193-L1198
bounds.position/size expects Logical position/size, but position_point and rect are Physical values.
So conversion is necessary.

closes tauri-apps/tauri#10053


pewsheen (dismissed):

wusyong:
Thanks for helping fix the bug! Could you a change file before I can merge it?

kanatapple:
@wusyong 
I added a change file.

src/webview2/mod.rs
```diff
@@ -1178,10 +1178,12 @@ impl InnerWebView {
   }
 
   pub fn bounds(&self) -> Result<Rect> {
-    let mut bounds = Rect::default();
+    let dpi = unsafe { util::hwnd_dpi(self.hwnd) };
+    let scale_factor = util::dpi_to_scale_factor(dpi);
```
amrbashir: we don't need these, since we can just use Physical position without converting to logical

src/webview2/mod.rs
```diff
@@ -1190,22 +1192,17 @@ impl InnerWebView {
...
       unsafe { MapWindowPoints(self.hwnd, *self.parent.borrow(), position_point) };
 
-      bounds.position = LogicalPosition::new(position_point[0].x, position_point[0].y).into();
-      bounds.size = LogicalSize::new(
-        (rect.right - rect.left) as u32,
-        (rect.bottom - rect.top) as u32,
-      )
-      .into();
+      bounds.position = PhysicalPosition::new(position_point[0].x, position_point[0].y)
+        .to_logical::<f64>(scale_factor)
```
amrbashir: no need to convert to logical

src/webview2/mod.rs
```diff
@@ -1190,22 +1192,17 @@ impl InnerWebView {
...
       unsafe { self.controller.Bounds(&mut rect) }?;
-      bounds.size = LogicalSize::new(
-        (rect.right - rect.left) as u32,
-        (rect.bottom - rect.top) as u32,
-      )
-      .into();
     }
 
+    bounds.size = PhysicalSize::new(rect.right - rect.left, rect.bottom - rect.top)
+      .to_logical::<f64>(scale_factor)
```
amrbashir: no need to convert to logical

amrbashir (changes requested):

src/webview2/mod.rs
kanatapple: @amrbashir 
https://github.com/tauri-apps/tauri/issues/10053

Have you seen this issue?
This is caused by not converting to logical.

amrbashir: the field `position` and `size` in `Rect` are of type `Position` and `Size` which can be constructed directly from `Physical` or `Logical` type, it is the responsibility of the user to convert the position/size later to their preferred variant.

pewsheen: I think Amr means Rect can take PhysicalPosition dierctly, and we can convert it later when needed.

amrbashir: you can do

```rs
Rect {
	position: PhysicalPositon::new().into(),
}
```
as well as:
```rs
Rect {
	position: LogicalPosition::new().into(),
}
```

and the user of `.bounds()` function can convert to their needed types:
```rs
let bounds = webview.bounds()?;
let position = bounds.position.to_logical(scale_factor);
```

kanatapple: Is this what you mean?
```
bounds.position = PhysicalPosition::new(position_point[0].x, position_point[0].y).into();
```

The original code is wrong because it uses logical, right?
```
bounds.position = LogicalPosition::new(position_point[0].x, position_point[0].y).into();
```

amrbashir: yeah this is what I mean

amrbashir: I still haven't tested this so I am not sure yet if the correct value should be `LogicalPosition` or `PhysicalPosition` or maybe the bug could in `set_bounds` not `bounds` where it should be passing logical values to win32 APIs instead of physical.

kanatapple:
@amrbashir 
I just set physical values without converting.

kanatapple:
Sorry I don't understand the process, is there anything else I should do?

src/webview2/mod.rs
pewsheen: I got a bit confused by the definition of `Logical` lol

~- (set) `SetWindowPos` in `set_bound` should be `Logical`, meaning it should have DPI awareness.~

- (set) `SetWindowPos` in `set_bound` should be `Physical`, since our app is DPI unawareness.
- (get) The position we get from `GetClientRect` should be `Physical` since our app is DPI unawareness. (not quite sure)

~- So the `LogicalPosition` for  `SetWindowPos` should be the `PhysicalPosition` from `MapWindowPoints` * `dpi_scale_factor`, which gets a larger value if the DPI is higher.~

It looks different from macOS's implementation because macOS will get a smaller coordinate if the DPI is higher.

pewsheen: ~I'm not sure what we get from WIN32 API is already scaled now :cry: , but I think we should use `Logical` in `set_bounds`~

Update: oh, I forgot we are DPI unawareness, so it should be `Physical`, Windows will scale it by the DPI factor

kanatapple:
@amrbashir 
Is there anything else I should do? (except fixing covector)
Sorry to bother you, but please check.

amrbashir:
Apologies, I got busy with some other stuff, I will review and test today.

.changes/fix-set-physical-values-in-bounds.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": patch
+---
+
+Set physical values in bounds method
```
amrbashir:
```suggestion
Fix `Webview::bounds` returning logical values where it should have been physical.
```

amrbashir (approved):
Thank you

kanatapple:
@amrbashir 
Thank you for the review and merge!

PR #1303, opened 2024-06-27
Enable Change Comments On Fork-based PRs

jbolda:
## Motivation

PRs from forks were failing as they don't have the permissions to comment to the PR. With changes upstream in covector and adding a workflow_run, we can enable these comments to be securely created.


jbolda:
Note we will switch back to a tagged CI run once the next version is merged in: https://github.com/jbolda/covector/pull/317.

amrbashir (approved):

PR #1306, opened 2024-07-01
enhance(windows): support webview2 version before 101

Legend-Master:
Support WebView2 version older than 101.0.1210.39 and document `incognito` and `theme` will not work for versions before it

Reference: https://github.com/tauri-apps/tauri/issues/10157

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -963,6 +963,8 @@ impl<'a> WebViewBuilder<'a> {
   ///
   /// ## Platform-specific:
   ///
+  /// - Windows: Requires WebView2 Runtime version 101.0.1210.39 or higher, does nothing on older versions,
+  /// see https://learn.microsoft.com/en-us/microsoft-edge/webview2/release-notes/archive?tabs=dotnetcsharp#10121039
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -470,6 +470,8 @@ pub struct WebViewAttributes {
   ///
   /// ## Platform-specific:
   ///
+  /// - Windows: Requires WebView2 Runtime version 101.0.1210.39 or higher, does nothing on older versions,
+  /// see https://learn.microsoft.com/en-us/microsoft-edge/webview2/release-notes/archive?tabs=dotnetcsharp#10121039
```
pewsheen, line 474: I think we should also mention the requirement on `with_incognito` (line 961).

I don't know if the document is enough. Since personal information is quite sensible, the developer may want to know if the mode is enabled or if they must detect it in other ways.

Legend-Master: Since we documented it requires version `101.0.1210.39` or higher and always have no effect on older versions, devs can provide fallbacks or something else by comparing the version with `webview_version`

amrbashir (approved):

Issue #1318, opened 2024-07-18
winit example panics on macOS


johanholmerin:
**Describe the bug**
The winit example panics on macOS whenever the window is unfocused. I've tried updating winit but get the same error.

**Steps To Reproduce**
Run `cargo run --example winit` and then focus another window

**Expected behavior**
The example not to panic

**Screenshots**

**Platform and Versions (please complete the following information):**
OS: macOS Sonoma 14.5
Rustc: 1.79.0

**Additional context**
```log
thread 'main' panicked at /Users/dev/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.4.1/src/runtime/mod.rs:1015:17:
ivar _state not found on class WryWebViewParent
stack backtrace:
   0: rust_begin_unwind
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/panicking.rs:652:5
   1: core::panicking::panic_fmt
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/panicking.rs:72:14
   2: objc2::runtime::ivar_offset
             at /Users/dev/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.4.1/src/runtime/mod.rs:1015:17
   3: objc2::declare::ivar::IvarType::__offset
             at /Users/dev/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.4.1/src/declare/ivar.rs:97:9
   4: objc2::declare::ivar::Ivar<T>::as_inner_ptr
             at /Users/dev/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.4.1/src/declare/ivar.rs:218:31
   5: <objc2::declare::ivar::Ivar<T> as core::ops::deref::Deref>::deref
             at /Users/dev/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.4.1/src/declare/ivar.rs:271:18
   6: winit::platform_impl::platform::view::WinitView::reset_modifiers
             at /Users/dev/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.15/src/platform_impl/macos/view.rs:905:13
   7: winit::platform_impl::platform::window_delegate::WinitWindowDelegate::window_did_resign_key
[...]

PR #1323, opened 2024-07-22
fix(example): workaround to prevent winit crash when unfocused

pewsheen:
It's a workaround for issue https://github.com/tauri-apps/wry/issues/1318

### Root cause
In Wry, we replace the content view with `WryWebViewParent` to pass key events to the application menu. When the window is unfocused, winit will call `WindowDelegate::window_did_resign_key` and get the `contentView`. It assumes the `contentView` is `WinitView` and gets its ivar `_state`. The ivar does not exist in our `WryWebViewParent`, so the exception is thrown.

### Workaround
`WebViewBuilder::new_as_child` will not create `WryWebViewParent` to replace the `contentView`, so it will be safe to use in winit.

### DIFF
--- a/.changes/example-winit-unfocus-crash.md
+++ b/.changes/example-winit-unfocus-crash.md
@@ -1,5 +1,0 @@
----
-"wry": patch
----
-
-Use a workaround for winit example to prevent crash when unfocus the window.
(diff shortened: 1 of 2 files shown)


.changes/example-winit-unfocus-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": patch
```
wusyong: No need for patch if it's example.

wusyong (approved):

PR #1326, opened 2024-07-24
fix(linux): don't share UserContentManager between webviews (dev)

### LINKED ISSUE #1308: IPC broken when sharing context between WebViews on Linux
**Describe the bug**
If I spawn multiple WebViews with a shared WebContext, IPC callbacks (from JS to Rust) are routed incorrectly. It is as if each window sent the same event separately.

**Steps To Reproduce**

Apply the following changes to `examples/multiwindow.rs`. It makes the windows use a shared WebContext.
<details><summary>Diff</summary>
<p>

```diff
diff --git a/examples/multiwindow.rs b/examples/multiwindow.rs
index 1a95c09..5b50a2a 100644
--- a/examples/multiwindow.rs
+++ b/examples/multiwindow.rs
@@ -8,7 +8,7 @@ use tao::{
   event_loop::{ControlFlow, EventLoopBuilder, EventLoopProxy, EventLoopWindowTarget},
   window::{Window, WindowBuilder, WindowId},
 };
-use wry::{http::Request, WebView, WebViewBuilder};
+use wry::{http::Request, WebContext, WebView, WebViewBuilder};
 
 enum UserEvent {
   CloseWindow(WindowId),
@@ -20,11 +20,13 @@ fn main() -> wry::Result<()> {
   let event_loop = EventLoopBuilder::<UserEvent>::with_user_event().build();
   let mut webviews = HashMap::new();
   let proxy = event_loop.create_proxy();
+  let mut web_context = WebContext::new(None);
 
   let new_window = create_new_window(
     format!("Window {}", webviews.len() + 1),
     &event_loop,
     proxy.clone(),
+    &mut web_context,
   );
   webviews.insert(new_window.0.id(), (new_window.0, new_window.1));
 
@@ -47,6 +49,7 @@ fn main() -> wry::Result<()> {
           format!("Window {}", webviews.len() + 1),
           event_loop,
           proxy.clone(),
+          &mut web_context,
         );
         webviews.insert(new_window.0.id(), (new_window.0, new_window.1));
       }
@@ -69,6 +72,7 @@ fn create_new_window(
   title: String,
   event_loop: &EventLoopWindowTarget<UserEvent>,
   proxy: EventLoopProxy<UserEvent>,
+  web_context: &mut WebContext,
 ) -> (Window, WebView) {
   let window = WindowBuilder::new()
     .with_title(title)
@@ -114,6 +118,7 @@ fn create_new_window(
   };
 
   let webview = builder
+    .with_web_context(web_context)
     .with_html(
       r#"
         <button onclick="window.ipc.postMessage('new-window')">Open a new window</button>
```

</p>
</details> 

[Full example](https://gist.github.com/ollpu/76f7cea00788c0753209c95b3e6979e3)

Then:
- Start the example with `cargo run --example multiwindow`
- Open a new window using the button
- Try to change the title of either of the windows. Both windows change their title.
- Press the "new window" button again. This time two new windows open.

**Expected behavior**
As with the original example,
- only the current window changes its title based on the text field
- only one new window is opened each time the button is pressed

**Screenshots**
image

**Platform and Versions:**
OS: Ubuntu 22.04.4
Rustc: rustc 1.77.2 (25ef9e3d8 2024-04-09)

**Additional context**
I primarily want to share the WebContext between WebViews so that the `BroadcastChannel` API would work. It seems that on Linux / WebKitGTK, it doesn't work if the views are in separate contexts. (Same with live localStorage changes, etc.)

In Tauri the WebViews are currently always set up with separate contexts, **unless** `TAURI_AUTOMATION=true` (Tauri v1) or `TAURI_WEBVIEW_AUTOMATION=true` (Tauri v2) is set, in which case this bug is exhibited.

ollpu:
Fixes #1308

Init scripts are registered to the `UserContentManager`, so if it is shared between `WebView`s, the init scripts conflict. IPC handlers are also tied to the `UserContentManager`. Use a unique `UserContentManager` for each `WebView` instead of sharing it across the `WebContext`.

Corresponding PR on the v0.24 branch: #1325

### DIFF
--- a/.changes/linux-shared-webcontext-ipc.md
+++ b/.changes/linux-shared-webcontext-ipc.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix Linux IPC handler and initialization scripts when sharing a WebContext between multiple WebViews.
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -28,7 +28,7 @@ use webkit2gtk::{
   AutoplayPolicy, InputMethodContextExt, LoadEvent, NavigationPolicyDecision,
   NavigationPolicyDecisionExt, NetworkProxyMode, NetworkProxySettings, PolicyDecisionType,
   PrintOperationExt, SettingsExt, URIRequest, URIRequestExt, UserContentInjectedFrames,
-  UserContentManagerExt, UserScript, UserScriptInjectionTime,
+  UserContentManager, UserContentManagerExt, UserScript, UserScriptInjectionTime,
   WebContextExt as Webkit2gtkWeContextExt, WebView, WebViewExt, WebsiteDataManagerExt,
   WebsiteDataManagerExtManual, WebsitePolicies,
 };
@@ -266,7 +266,7 @@ impl InnerWebView {
     Self::attach_handlers(&webview, web_context, &mut attributes);
 
     // IPC handler
-    Self::attach_ipc_handler(webview.clone(), web_context, &mut attributes);
+    Self::attach_ipc_handler(webview.clone(), &mut attributes);
 
     // Drag drop handler
     if let Some(drag_drop_handler) = attributes.drag_drop_handler.take() {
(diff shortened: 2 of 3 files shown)


ollpu:
Note: This doesn't completely fix opening multiple windows with `TAURI_WEBVIEW_AUTOMATION=true` in Tauri, because it tries to register the `tauri://` custom protocol for each webview separately.

amrbashir (approved):

PR #1340, opened 2024-08-15
fix(android): custom protocol not triggered on redirect

lucasfernog:
The shouldInterceptRequest function is not called on redirects, which means any redirect to a custom protocol URL will fail to load, so we must apply a small check in the onReceivedError function.

This changes the Android implementation to handle URL load errors to retry loading the URL to give a chance of the custom protocol function to handle the request.


lucasfernog:
fyi this is required for some oauth2 flows

amrbashir (approved):

PR #1342, opened 2024-08-16
fix(android): avoid rebuilds if nothing changed

Flakebi:
Unconditionally overwriting files where the build reruns if they changed leads to rebuilds every time.
Only overwrite a file if its content is different to not rebuild in such a case.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -93,10 +93,7 @@ fn main() {
 
         let out_path = kotlin_out_dir.join(file.file_name());
         // Overwrite only if changed to not trigger rebuilds
-        if fs::read_to_string(&out_path)
-          .map(|o| o != out)
-          .unwrap_or(false)
-        {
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -92,7 +92,13 @@ fn main() {
...
 
         let out_path = kotlin_out_dir.join(file.file_name());
-        fs::write(&out_path, out).expect("Failed to write kotlin file");
+        // Overwrite only if changed to not trigger rebuilds
+        if fs::read_to_string(&out_path)
+          .map(|o| o != out)
+          .unwrap_or(false)
+        {
+          fs::write(&out_path, out).expect("Failed to write kotlin file");
+        }
```
lucasfernog:
```suggestion
        if fs::read_to_string(&out_path).map_or(true, |o| o != out) {
          fs::write(&out_path, out).expect("Failed to write kotlin file");
        }
```

lucasfernog (approved):

PR #1352, opened 2024-09-03
fix(android): file picker result processing

lucasfernog:
The Android file picker result callback implementation is incorrectly checking the expected minimum result length. This only impacts usage on SDK 34+ because on older releases the `FileChooserParams.parseResult` properly resolves the response.


lucasfernog:
From my debugging it looks like before api 34 the data was shared as clipData only for multiple file selection, so our broken if() condition didn’t trigger issues

amrbashir (approved):

PR #1354, opened 2024-09-05
fix(android): error loading resource redirects webview to resource url

millermk:
Currently, the `onReceivedError` override is checking for `net::ERR_CONNECTION_REFUSED` errors and retrying them using `view.loadUrl()`. However, based on [the documentation](https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20android.webkit.WebResourceRequest,%20android.webkit.WebResourceError)), `onReceivedError` will be called if there is an error when loading an image, iframe, or other resource, not just the main page. In these cases, the request should not be retried this way.

Therefore, this PR adds a check for `request.isForMainFrame()`.

To give an example of the issue, we experienced a case where a custom DNS server blocked loading of a certain external stylesheet, which caused the webview to be redirected to the url of that stylesheet, resulting in an unexpected user-visible error 

```
Web page not available.
The web page at example.com/styles.css could not be loaded because:
new::ERR_CONNECTION_REFUSED"
```

### DIFF
--- a/.changes/fix-android-error-handling-page-visible.md
+++ b/.changes/fix-android-error-handling-page-visible.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fixes Android webview error page flashing when a redirect to the app is performed.
--- a/.changes/fix-request-error-handling-android.md
+++ b/.changes/fix-request-error-handling-android.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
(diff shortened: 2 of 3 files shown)


lucasfernog:
while testing this I also found some other bugs in the implementation... had to push a few really weird (familiar if you're an Android dev) workarounds

matthewlst:
Thanks Lucas for your testing and improvements!

amrbashir (approved):

PR #1360, opened 2024-09-10
fix(android): skip duplicated Content-Type/Content-Length header to fix tauri-apps/tauri#8857

Steve-xmh:
As title follows, currently Android binding need to pass `Content-Type` separately when creating `WebResourceReponse` objects, but the `response_headers` doesn't skip `Content-Type` and `Content-Length` headers, which will be append into auto-generated headers with comma, which causes issue tauri-apps/tauri#8857 (e.g. `application/wasm` becomes `application/wasm, application/wasm`)

This PR fixed it by adding single conditional statement to skip these headers inside for loop when putting header into objects.

### DIFF
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -216,6 +216,8 @@ fn handle_request(
       let response_headers = {
         let headers_map = JMap::from_env(env, &obj)?;
         for (name, value) in headers.iter() {
+          // WebResourceResponse will automatically generate Content-Type and
+          // Content-Length headers so we should skip them to avoid duplication.
           if name == CONTENT_TYPE || name == CONTENT_LENGTH {
             continue;
           }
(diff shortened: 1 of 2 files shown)


src/android/binding.rs
```diff
@@ -216,6 +216,9 @@ fn handle_request(
       let response_headers = {
         let headers_map = JMap::from_env(env, &obj)?;
         for (name, value) in headers.iter() {
+          if name == CONTENT_TYPE || name == CONTENT_LENGTH {
```
amrbashir, line 221: let's add a comment to explain why we are skipping these

amrbashir (changes requested):
Thank you, could you also add a change file in `.changes` directory?

src/android/binding.rs
Steve-xmh: It's added now :D

amrbashir (approved):

amrbashir:
Thank you

PR #1365, opened 2024-09-21
feat: Allow injecting JavaScript code into subframes

### LINKED ISSUE #1313: Allow injecting JS into all Webview subframes
Hi, I have a use-case where I'd like to support plugins for my Tauri app using iframes. I was planning on injecting some initial JS to initialize the frame content with themes and metadata before the first render, instead of relying on `postMessage` which would cause flickering. I built a working prototype of this on Windows before realizing that it was only working because a. I was on Windows and b. I was serving the iframe content from the same origin.

I would like to be able to opt-in individual initialization scripts to running on subframes within a Webview. I found the TODOs related to this feature ([like here](https://github.com/tauri-apps/wry/blob/93eddc31df613cd42688e85f434c6abde9c1250b/src/wkwebview/mod.rs#L1119)) so this is clearly something that the team _intends_ to support at some point, but I'd like to see what I could do to push it forward. I'd be happy to implement it myself but I'm not exactly clear on what the team was envisioning with these TODOs and I don't have access to all of the platforms Wry supports to test on.

Norbiros:
Resolves #1313

Based on changes from [Modrinth's fork](https://github.com/modrinth/wry)


src/lib.rs
```diff
@@ -891,6 +895,12 @@ impl<'a> WebViewBuilder<'a> {
     self
   }
 
+  /// Whether JavaScript code should be injected to all subframes
```
amrbashir: I think this option will need to be more granular, i.e. choose which script should be injected into subframes and which is not. cc @tweidinger

amrbashir (changes requested):

src/lib.rs
Norbiros: I'm not sure how to do that, because I don't see any options in here: `      let script: id = msg_send![userscript, initWithSource:NSString::new(js) injectionTime:0 forMainFrameOnly:for_main_frame_only];`. Maybe that can be own separate feature added in future?

amrbashir: Instead of adding a new field, you can just modify [`initialization_scripts`](https://github.com/Norbiros/wry/blob/c65ca8486966a996f640aa0c2510ee1bf6b07c3e/src/lib.rs#L358) field and change its type from `Vec<String>` to `Vec<(String, bool)>` where the bool defines whether to inject into subframes or not.

Then you can add a new method called `with_initialization_script_subframes`

Norbiros:
I am unable to change my changes on platforms other than linux (webkitgtk), sorry

Norbiros:
Any updates @amrbashir?

amrbashir (approved):

amrbashir:
Thank you @Norbiros

Issue #1367, opened 2024-09-24
Reloading a webview initialized with `with_html` results in an empty page


just-be-dev:
**Describe the bug**

When initializing a webview with `with_html`, the webview is blank after reload. 

**Steps To Reproduce**
```rust
fn main() -> wry::Result<()> {
    use tao::{
        event::{Event, WindowEvent},
        event_loop::{ControlFlow, EventLoop},
        window::WindowBuilder,
    };
    use wry::WebViewBuilder;

    let event_loop = EventLoop::new();
    let window = WindowBuilder::new().build(&event_loop).unwrap();

    let _webview = WebViewBuilder::new(&window)
        .with_html("<h1>Hello, World!</h1>")
        .build()?;

    event_loop.run(move |event, _, control_flow| {
        *control_flow = ControlFlow::Wait;

        match event {
            Event::WindowEvent {
                event: WindowEvent::CloseRequested,
                ..
            } => *control_flow = ControlFlow::Exit,
            _ => (),
        }
    });
}
```

Right click the webview and select reload. 

**Expected behavior**

When reloading, the initial HTML should be present. 

**Screenshots**

**Platform and Versions (please complete the following information):**
OS: MacOS Sonoma 14.6.1
Rustc: rustc 1.78.0 (9b00956e5 2024-04-29)

**Additional context**

I'm sure that I could _prevent_ reloading via the navigation handler, but that seems slightly undesirable given that sometimes refreshing is an important recovery method for when things go really wrong in the UI. I would programmatically inject the HTML on refresh, but it's not clear to me exactly how to go about doing that. [`.load_url`](https://docs.rs/wry/latest/wry/struct.WebView.html#method.load_url) with a data URL?

just-be-dev:
For extra context, I stumbled across this as I've been working on https://github.com/zephraph/webview which is a deno binding (ish) for Wry. The HTML usecase is important, but as I was working through building out some example apps I had one that was refreshing (for reasons I haven't yet determined), but that makes the app lose its content.

amrbashir:
looking at our usage https://github.com/tauri-apps/wry/blob/aa42ca437262729afd0e3188f00fabec675f0dbe/src/wkwebview/mod.rs#L1166 and the docs for `loadHTMLString`, there is nothing weird we are doing, so I suspect this just how `loadHTMLString` behaves. 

We don't recommend using `wiht_html` and instead register a custom protocol to serve the html and use `with_url`.

[...]

PR #1368, opened 2024-09-25
feat: add `WebView::load_html` for loading HTML after intialization

just-be-dev:
This PR aims to implement `load_html`, a companion of `load_url` to balance out the current implementation. Currently, the webview can be initialized with either a URL or HTML via `.with_url` or `.with_html` in the builder. After creation `.load_url` can be used to load a different URL, but the same can't be said for `.load_html`. 

In #1367 I noted that refreshing a webview initialized with `.load_html` causes the webview to be left with an empty screen. While that's not ideal, if this method existed it'd be easy enough to simply call `load_html` with the same payload as what the webview was initialized with to get it back in the correct state.

### DIFF
--- a/.changes/load-html.md
+++ b/.changes/load-html.md
@@ -1,0 +1,5 @@
+---
+"wry": minor
+---
+
+Add `Webview::load_html`.
--- a/src/android/main_pipe.rs
+++ b/src/android/main_pipe.rs
@@ -300,6 +300,12 @@ impl<'a> MainPipe<'a> {
               .call_method(webview, "clearAllBrowsingData", "()V", &[])?;
           }
         }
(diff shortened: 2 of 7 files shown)


just-be-dev:
I've got a PR in my library that tests the code path for this PR: https://github.com/zephraph/webview/pull/56. That's what I've used to test the implementation.

just-be-dev:
I got someone at [Recurse](https://www.recurse.com/) to verify that it works on windows as well. Still need to test on linux, but I'll get someone for that soon. My library isn't really setup to be ran on iOS or android so I'd appreciate extra scrutiny there. Overall the implementation seems straight forward though.

amrbashir (approved):
Thank you

PR #1372, opened 2024-09-26
fix(macos): do not ignore errors on sending HTTP responses

rhysd:
All errors on sending HTTP responses are ignored on macOS.

https://github.com/tauri-apps/wry/blob/eafaadb9b35f74f07ef83da3d3e738c9ae631f7c/src/wkwebview/mod.rs#L345

This PR changes to check the errors. When some error happens, it now causes a panic.

### DIFF
--- a/.changes/dont-ignore-response-errors-macos.md
+++ b/.changes/dont-ignore-response-errors-macos.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+On macOS, do not ignore errors on sending HTTP resopnses. Errors now cause panics.
--- a/src/error.rs
+++ b/src/error.rs
@@ -55,8 +55,6 @@ pub enum Error {
   #[cfg(target_os = "android")]
   #[error(transparent)]
   CrossBeamRecvError(#[from] crossbeam_channel::RecvError),
-  #[error("Custom protocol task is invalid.")]
-  CustomProtocolTaskInvalid,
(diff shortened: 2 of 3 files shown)


.changes/dont-ignore-response-errors-macos.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": patch
+---
+
+On macOS, do not ignore errors on sending HTTP resopnses. Errors now cause panics.
```
amrbashir, line 5: I don't think our users would appreciate panics, we should instead just `log::error` the result

rhysd: Hmm, does it mean users can recover the error when it occurs? This is an internal error and users don't know why it happened. And there is no other way to receive a response via custom protocol. For unrecoverable errors, I believe panics are a correct way to handle them.

https://doc.rust-lang.org/beta/book/ch09-03-to-panic-or-not-to-panic.html

amrbashir: @lucasfernog what do you think?

FabianLars: I don't think the quote from the rust book applies well to gui applications. gui apps should really never crash like this, or at least not from inside a library (from someone else than the app dev), but of course the dev not having a way to react to those errors is quite problematic too, hmm 🤔

rhysd:
I apologize for the huge delay. I've now noticed this PR is still open.

PR #1383, opened 2024-10-08
example: Add GPUI example

huacnlee:
```bash
cargo run --example gpui
```

- Ref https://github.com/zed-industries/zed/pull/24545 wait this PR to support run on Window.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,17 +19,51 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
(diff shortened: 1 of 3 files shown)


amrbashir:
GPUI is still in early development and I don't think we should have a dedicated example for it, not yet at least

clearlysid:
I understand why this PR is closed but had a follow-on question: would it be possible to go the other way round and render some gpui-driven UI into a tao/wry/tauri-provided native window (something with `HasWindowHandle`)?

I imagine it might be useful to some Tauri apps to avoid the whole webview abstraction for *some* perf-sensitive UI work here and there.

FabianLars:
In theory yes but as far as i can tell this would require changes in gpui since it currently seems to hide all the windowing internally.

In the meantime one could probably fork it so that it uses tao and then create something like https://github.com/tauri-apps/tauri-egui (i have no idea how much work that would be)

clearlysid:
Oooh, thanks for pointing me there! `tao` ↔ `egui`  seems like it could work for my use-case, but that implementation seems old. Let me try to shake off the cobwebs a little 🫡

FabianLars:
oh yeah, that's pretty old by now and essentially unmaintained. it would be very much appreciated if someone could pick it up again (the prior maintainers don't work on tauri anymore) but it _may_ be easier to wait until we switch back to winit (or have a fresh fork) but who knows how long that'll take.
Anyway, good luck 🙏

clearlysid:
Got a crude version working! https://github.com/clearlysid/egui-tao

Still thinking about what a good way to make a "low-overhead" (ie.. doesn't involve maintaining multiple forks) integration is.

> until we switch back to winit

Are there plans for this in-place already? Given how widely adopted winit is in the rest of the Rust ecosystem, I think this will simplify a lot of things. Especially if it now supports all the things that Tauri needs, which it may not have before.


Either ways, let's tune off of this thread since we're wayy off-topic and continue on Discord/elsewhere. Cheers!

PR #1384, opened 2024-10-08
refactor!: add webview id to protocol handlers and improve API ergonomics

amrbashir:

### DIFF
--- a/.changes/custom-protocol-label.md
+++ b/.changes/custom-protocol-label.md
@@ -17,7 +17,7 @@ We also made a few changes to the builder, specficially `WebViewBuilder::new` an
 - Added `WebViewAttributes.id` field to specify an id for the webview.
 - Added `WebViewBuilder::with_id` method to specify an id for the webview.
 - Added `WebViewAttributes.context` field to specify a shared context for the webview.
-- **Breaking** Changed `WebViewAttributes.custom_protocols` field,`WebViewBuilder::with_custom_protocol` method and `WebViewBuilder::with_async_custom_protocol` method handler function to take `Option<WebViewId>` as the first argument to check which webview made the request to the protocol.
+- **Breaking** Changed `WebViewAttributes.custom_protocols` field,`WebViewBuilder::with_custom_protocol` method and `WebViewBuilder::with_async_custom_protocol` method handler function to take `WebViewId` as the first argument to check which webview made the request to the protocol.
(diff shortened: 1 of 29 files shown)


lucasfernog (dismissed):

.changes/custom-protocol-label.md
```diff
@@ -0,0 +1,28 @@
...
+- Added `Error::ContextDuplicateCustomProtocol` enum variant.
+- On Linux, return an error in `WebViewBuilder::build` if registering a custom protocol multiple times.
+- Added `WebContext::is_custom_protocol_registered` to check if a protocol has been regsterd for this web context.
+- Added `WebViewId` alias type.
+- **Breaking** Changed `WebViewAttributes` to have a lifetime parameter.
+- Added `WebViewAttributes.id` field to specify an id for the webview.
+- Added `WebViewBuilder::with_id` method to specify an id for the webview.
+- Added `WebViewAttributes.context` field to specify a shared context for the webview.
+- **Breaking** Changed `WebViewAttributes.custom_protocols` field,`WebViewBuilder::with_custom_protocol` method and `WebViewBuilder::with_asynchronous_custom_protocol` method handler function to take `WebViewId` as the first argument to check which webview made the request to the protocol.
+- **Breaking** Changed `WebViewBuilder::with_web_context` to be a static method to create a builder with a webcontext, instead of it being a setter method. It is now an alternative to `WebviewBuilder::new`
```
MarijnS95, line 21: Was there a specific reason to change this? While simple to solve, it it's annoying and seems unnecessary to deal with while migrating through version upgrades. While at it, it keeps the "builder pattern" `with_` prefix rather than being named `new_with_web_context()` or something? `with_attributes()` should probably also receive a `new_` prefix?

amrbashir: I agree I could've named it a bit better, but we can change that in upcoming release.

The reason why is it changed to a web context, is that so `with_custom_protocol` or `with_async_custom_protocol` could check (on Linux) if a protocol with the same name already registered (limitation of webkit2gtk)

amrbashir: `new_` prefix -> https://github.com/tauri-apps/wry/pull/1558

MarijnS95: Right I see now, those functions set something on the `context` (if not `None`) so it wouldn't make sense to set or write to it afterwards. In addition, it also touches on other fields in the builder struct, so moving it to the `context` wouldn't have been possible. Thanks for the explanation, I retract my statement that this seemed like unnecessary churn.

PR #1385, opened 2024-10-08
feat: add `WebView::focus_parent` method.

huacnlee:
The blur method need in some case, for example the WebView is a child in window (The WebView is not full size in the Window). When we click the main area, we want the WebView to lose the focus handle.

- [x] WKWebView
- [x] WebView2 (need to test)
- [ ] WebkitGTK, Android (I don't find the API doc how to implement it, so keep is do nothing.)

I use this for add Wry into GPUI, here is the bug fix:

https://github.com/huacnlee/gpui-component/pull/320

## Before

https://github.com/user-attachments/assets/cc614334-47fe-4722-8745-aea783fbab4a

## After

After called `cx.blur` on mouse click outside the WebView area.

On macOS, now the IME bug is gone.

https://github.com/user-attachments/assets/363bb016-229f-4766-84f7-6a4838e490bf


huacnlee:
Windows as test passed.

https://github.com/user-attachments/assets/320c5c1c-cfac-4165-af8a-bc9708952460

amrbashir (changes requested):
I don't think the `blur` name here is correct. In my head, blur means move focus back to previous focus-holder, for example when switching from one webview to another. The current implementation moves focus to parent window only and thus not the same as blur.

I think it should be named more accurately and should be called `focus_parent` or something. 

At which point, wouldn't it be better to just call `focus` on the hosting window?

huacnlee:
I can rename it do `focus_parent`.

This should be provided a method from wry, because I can't get the parent view in outside.

amrbashir:
I guess we can include it then.

amrbashir:
>  WebkitGTK, Android (I don't find the API doc how to implement it, so keep is do nothing.)

Android can be documented as unsupported. As for webkit2gtk, try calling [`widget.parent_window()`](https://docs.gtk.org/gtk3/method.Widget.get_parent_window.html) and then call focus on it.

huacnlee:
Can I just keep Android unsupport? the reason is:

1. I found that the `focus` method also not did supported in Android.
2. I don't know how to test it, to make sure the change is correct.

amrbashir (dismissed):
Thank you

amrbashir (approved):

PR #1394, opened 2024-10-16
feat: add `WebView::cookies` and `WebView::cookies_for_url` 

### LINKED ISSUE #518: want an API can get webview cookie
I want use wry load a remote url, wait user login, and get cookie to use user cookie to call thrid-service api, so I need get cookie on user login sucesss.

I use electron session to finish it, the sesion can give all webview  cookies, cache, proxy setting...
but I can't found tauri have similar API

I found wry can preload script , I try init script and send event to main process, but the remote url will run 302 redict and callback , my init script will loss.

- [ x ] No

**Additional context**
electron session : https://www.electronjs.org/docs/latest/api/session

amrbashir:
closes #518
ref: https://github.com/tauri-apps/tauri/issues/11330

### DIFF
--- a/.changes/cookies-api.md
+++ b/.changes/cookies-api.md
@@ -1,5 +1,5 @@
 ---
-"wry": "patch:bug"
+"wry": "patch"
 ---
 
 Add `WebView::cookies` and `WebView::cookies_for_url` APIs.
--- a/.changes/cookies-api.md
+++ b/.changes/cookies-api.md
@@ -1,0 +1,5 @@
+---
+"wry": "patch:bug"
+---
+
+Add `WebView::cookies` and `WebView::cookies_for_url` APIs.
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: MIT
 
 use dpi::{LogicalPosition, LogicalSize};
+use ffi::CookieManageExt;
 use gdkx11::{
   ffi::{gdk_x11_window_foreign_new_for_display, GdkX11Display},
   X11Display,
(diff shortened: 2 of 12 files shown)


src/wkwebview/mod.rs
```diff
@@ -825,6 +821,96 @@ r#"Object.defineProperty(window, 'ipc', {
...
+            // cookie is secure and url is https
+            (cookie.secure() && url.scheme() == "https") ||
+            // or cookie is secure and is localhost
+            (
+              cookie.secure() && url.scheme() == "http" && 
+              (url.domain() == Some("localhost") || url.domain().and_then(|d| Ipv4Addr::from_str(d).ok()).map(|ip| ip.is_loopback()).unwrap_or(false))
+            ) ||
+            // or cookie is not secure
+            (!cookie.secure())
+          )
```
amrbashir: not sure if this is too restrictive, tried to check how webkit2gtk cookies_for_url is implementation but I think their implementation is a bit more relaxed ? https://github.com/WebKit/WebKit/blob/c305a86965236fd8052623faaf8804b1424e7f6b/Source/WebKit/UIProcess/API/APIHTTPCookieStore.cpp#L69

lucasfernog: when in doubt just ask @tweidinger though I think we should just match the behavior from other platforms if possible

tweidinger: I think we do match too strict on the path depending on the cookie policy.
For example `/` in the cookie would not match a subpath cookie but it should match any path.

tweidinger: Just checking the domain match and checking if secure cookie or not sounds acceptable if we don't want to implement the cookie specs fully 😁 .

amrbashir: alright, pushed a change to match on domain and secure or not. (basically just removed matching the path)

src/webkitgtk/mod.rs
```diff
@@ -804,6 +805,95 @@ impl InnerWebView {
...
+              .map(Self::cookie_from_soup_cookie)
+              .collect()
+          });
+          let _ = tx.send(cookies);
+        })
+      });
+
+    gtk::main_iteration();
+
+    rx.recv()?.map_err(Into::into)
```
lucasfernog:
```suggestion
    loop {
      gtk::main_iteration();

      if let Ok(response) = rx.try_recv() {
        return response.map_err(Into::into);
      }
    }
```

src/webkitgtk/mod.rs
```diff
@@ -804,6 +805,95 @@ impl InnerWebView {
...
```
lucasfernog:
```suggestion
    loop {
      gtk::main_iteration();

      if let Ok(response) = rx.try_recv() {
        return response.map_err(Into::into);
      }
    }
```

.changes/cookies-api.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": "patch:bug"
```
lucasfernog:
```suggestion
"wry": "patch"
```

lucasfernog (approved):

PR #1400, opened 2024-10-23
fix(windows): handle window borders when resizing to parent

amrbashir:
ref: https://github.com/tauri-apps/tao/pull/993
ref: https://github.com/rust-windowing/winit/pull/3712/


amrbashir:
should probably wait for https://github.com/tauri-apps/tao/pull/993

PR #1403, opened 2024-10-28
feat: Allow setting extension path for Windows and Linux

SpikeHD:
Introduces `with_extension_path`, which is now used by Windows and Linux to load extensions from.

Linux doesn't include `browser_extensions_enabled` because there is no setting to turn them on or off in WebkitGTK, only to set their directory. I assume if no directory is set, then no extensions are loaded from anywhere, so the option would be redundant.

### DIFF
--- a/.changes/extension-path.md
+++ b/.changes/extension-path.md
@@ -1,0 +1,5 @@
+---
+"wry": "patch"
+---
+
+Add `with_extension_path` API to Windows and Linux.
--- a/.changes/extension-path.md
+++ b/.changes/extension-path.md
@@ -2,4 +2,4 @@
 "wry": "patch"
 ---
 
-Add `with_extension_path` API to Windows and Linux.
+Add `WebViewBuilder::with_extension_path` API to Windows and Linux.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -244,6 +244,15 @@ use self::webview2::*;
 #[cfg(target_os = "windows")]
 use webview2_com::Microsoft::Web::WebView2::Win32::ICoreWebView2Controller;
 
+#[cfg(any(
+  target_os = "linux",
+  target_os = "dragonfly",
+  target_os = "freebsd",
+  target_os = "netbsd",
+  target_os = "openbsd",
+))]
+use webkit2gtk::WebContextExt;
+
 use std::{borrow::Cow, collections::HashMap, path::PathBuf, rc::Rc};
 
 use http::{Request, Response};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1402,6 +1402,18 @@ impl WebViewBuilderExtAndroid for WebViewBuilder<'_> {
   target_os = "netbsd",
   target_os = "openbsd",
 ))]
+#[derive(Default)]
+pub(crate) struct PlatformSpecificWebViewAttributes {
+  extension_path: Option<PathBuf>,
+}
+
+#[cfg(any(
+  target_os = "linux",
+  target_os = "dragonfly",
+  target_os = "freebsd",
+  target_os = "netbsd",
+  target_os = "openbsd",
+))]
 pub trait WebViewBuilderExtUnix<'a> {
   /// Consume the builder and create the webview inside a GTK container widget, such as GTK window.
   ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -244,15 +244,6 @@ use self::webview2::*;
 #[cfg(target_os = "windows")]
 use webview2_com::Microsoft::Web::WebView2::Win32::ICoreWebView2Controller;
 
-#[cfg(any(
-  target_os = "linux",
-  target_os = "dragonfly",
-  target_os = "freebsd",
-  target_os = "netbsd",
-  target_os = "openbsd",
-))]
-use webkit2gtk::WebContextExt;
-
 use std::{borrow::Cow, collections::HashMap, path::PathBuf, rc::Rc};
 
 use http::{Request, Response};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1261,7 +1261,7 @@ pub trait WebViewBuilderExtWindows {
   fn with_browser_extensions_enabled(self, enabled: bool) -> Self;
 
   /// Swt the path from which to load extensions from. Extensions stored in this path should be unpacked.
-  /// 
+  ///
   /// Does nothing if browser extensions are disabled. See [`with_browser_extensions_enabled`](Self::with_browser_extensions_enabled)
   fn with_extension_path(self, path: impl Into<PathBuf>) -> Self;
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1440,8 +1440,10 @@ impl<'a> WebViewBuilderExtUnix<'a> for WebViewBuilder<'a> {
   }
 
   fn with_extension_path(mut self, path: impl Into<PathBuf>) -> Self {
-    self.inner.platform_specific.extension_path = Some(path.into());
-    self
+    self.and_then(|mut b| {
+      b.platform_specific.extension_path = Some(path.into());
+      Ok(b)
+    })
   }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1915,16 +1915,6 @@ pub enum PageLoadEvent {
   Finished,
 }
 
-#[cfg(any(
-  target_os = "linux",
-  target_os = "dragonfly",
-  target_os = "freebsd",
-  target_os = "netbsd",
-  target_os = "openbsd",
-))]
-#[derive(Default)]
-pub(crate) struct PlatformSpecificWebViewAttributes;
-
 #[cfg(test)]
 mod tests {
   use super::*;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1260,9 +1260,9 @@ pub trait WebViewBuilderExtWindows {
   /// see https://learn.microsoft.com/en-us/microsoft-edge/webview2/release-notes/archive?tabs=dotnetcsharp#10221055
   fn with_browser_extensions_enabled(self, enabled: bool) -> Self;
 
-  /// Determines the path from which to load extensions from. Extensions stored in this path should be unpacked.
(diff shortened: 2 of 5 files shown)


.changes/extension-path.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": "patch"
+---
+
+Add `with_extension_path` API to Windows and Linux.
```
amrbashir:
```suggestion
Add `WebViewBuilder::with_extension_path` API on Windows and Linux.
```

src/lib.rs
```diff
@@ -1257,6 +1259,11 @@ pub trait WebViewBuilderExtWindows {
   /// Requires WebView2 Runtime version 1.0.2210.55 or higher, does nothing on older versions,
   /// see https://learn.microsoft.com/en-us/microsoft-edge/webview2/release-notes/archive?tabs=dotnetcsharp#10221055
   fn with_browser_extensions_enabled(self, enabled: bool) -> Self;
+
+  /// Determines the path from which to load extensions from. Extensions stored in this path should be unpacked.
```
amrbashir:
```suggestion
  /// Set the path from which to load extensions from. Extensions stored in this path should be unpacked.
```

src/lib.rs
```diff
@@ -1257,6 +1259,11 @@ pub trait WebViewBuilderExtWindows {
...
+  /// 
+  /// Does nothing if browser extensions are disabled.
```
amrbashir:
```suggestion
  /// Does nothing if browser extensions are disabled. See [`with_browser_extensions_enabled`](Self::with_browser_extensions_enabled)
```

src/webkitgtk/mod.rs
```diff
@@ -242,6 +242,13 @@ impl InnerWebView {
       }
     }
 
+    // Extension loading
+    if let Some(extension_path) = pl_attrs.extension_path {
+      if let Some(extension_path) = extension_path.to_str() {
+        web_context.os.set_web_extensions_directory(extension_path);
+      }
+    }
```
amrbashir: no need to convert to str here
```suggestion
    if let Some(extension_path) = &pl_attrs.extension_path {
      web_context.os.set_web_extensions_directory(extension_path);
    }
```

src/webkitgtk/web_context.rs
```diff
@@ -92,6 +92,10 @@ impl WebContextImpl {
     self.automation = flag;
     self.context.set_automation_allowed(flag);
   }
+
+  pub fn set_web_extensions_directory(&mut self, path: &str) {
+    self.context.set_web_extensions_directory(path);
+  }
```
amrbashir:
```suggestion
  pub fn set_web_extensions_directory(&mut self, path: &Path) {
    self.context.set_web_extensions_directory(path.to_string_lossy());
  }
```

src/webview2/mod.rs
```diff
@@ -499,6 +499,15 @@ impl InnerWebView {
...
     }
 
+    // Extension loading
+    unsafe {
+      if pl_attrs.browser_extensions_enabled {
+        if let Some(extension_path) = pl_attrs.extension_path {
+          Self::load_extensions(&webview, &extension_path)?;
+        }
+      }
+    }
```
amrbashir: let's move the unsafe block a little bit below

```suggestion
    if pl_attrs.browser_extensions_enabled {
      if let Some(extension_path) = pl_attrs.extension_path {
        unsafe { Self::load_extensions(&webview, &extension_path)? };
      }
    }
```

src/webview2/mod.rs
```diff
@@ -1174,6 +1183,31 @@ impl InnerWebView {
...
+
+  
+  #[inline]
+  unsafe fn load_extensions(webview: &ICoreWebView2, extension_path: &PathBuf) -> Result<()> {
+    let profile = webview
+      .cast::<ICoreWebView2_13>()?
+      .Profile()?
+      .cast::<ICoreWebView2Profile7>()?;
+
+    if let Some(extension_path) = extension_path.to_str() {
```
amrbashir: no need to call `to_str` here, just pass ito `fs::read_dir` directly.

src/webview2/mod.rs
```diff
@@ -1174,6 +1183,31 @@ impl InnerWebView {
...
+      // Iterate over all folders in the extension path
+      for entry in fs::read_dir(extension_path)? {
+        let path = entry?.path();
+        let path_str = path.to_str();
+
+        if let Some(path_str) = path_str {
```
amrbashir: no need to convert to str her as well, just do
```rs
let path_hs = HSTRING::from(path);
```

amrbashir (changes requested):

src/webview2/mod.rs
```diff
@@ -1174,6 +1184,24 @@ impl InnerWebView {
...
+  unsafe fn load_extensions(webview: &ICoreWebView2, extension_path: &PathBuf) -> Result<()> {
+    let profile = webview
+      .cast::<ICoreWebView2_13>()?
+      .Profile()?
+      .cast::<ICoreWebView2Profile7>()?;
+
+    // Iterate over all folders in the extension path
+    for entry in fs::read_dir(extension_path)? {
+      let path = entry?.path();
+      let path_hs = HSTRING::from(path);
```
amrbashir:
```suggestion
      let path_hs = HSTRING::from(path.as_path());
```

amrbashir (dismissed):

amrbashir:
Thank you

amrbashir:
Oh looks like you didn't sign your commits and I can't merge the PR unless they are signed. You need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928). `git rebase --exec 'git commit --amend --no-edit -n -S' -i dev`

SpikeHD:
One day I will remember to add `-S` when committing. Today was not that day

amrbashir:
> One day I will remember to add `-S` when committing. Today was not that day

🤣 , I have my global `.gitconfig` to always sign so I don't forget
```
[commit]
    gpgsign = true
```

amrbashir (approved):

zhangsan2lisi:
```
use std::path::PathBuf;
use tao::event::{Event, WindowEvent};
use tao::event_loop::{ControlFlow, EventLoop};
use tao::window::WindowBuilder;
use wry::{WebContext, WebViewBuilder, WebViewBuilderExtWindows};
fn main() {
        .build(&event_loop)
        .expect("failed to create window");
    let path = PathBuf::from("test");
    let mut context = WebContext::new(Some(path));
    let webview = WebViewBuilder::with_web_context(&mut context)
        .with_url("https://google.com/")
        // .with_browser_extensions_enabled(true)
        .with_extension_path(PathBuf::from("Extensions/6.6.3_0"))
        .build(&window)
        .expect("could not create webview");
    webview.open_devtools();

    event_loop.run(move |event, _, control_flow| {
        *control_flow = ControlFlow::Poll;

        match event {
            Event::WindowEvent {
                window_id,
                ..
            } => {
                *control_flow = ControlFlow::Exit;
            }
            _ => {}
        }
    });
}
```
@SpikeHD  @amrbashir 
First of all, thank you for implementing this function

I attempted to launch a window to load the following extension: https://microsoftedge.microsoft.com/addons/detail/vuejs-devtools/olofadcdnkkjdfgjcmjaadnlehnnihnl
 I have already unpacked and repackaged the extension. When tested using Chromium in Electron with the following arguments, it works perfectly fine:
```
args: [ 
  `--disable-extensions-except=${extensionPath}`,
  `--load-extension=${extensionPath}`,
  "--no-sandbox",
  "--disable-sync",
]
```

However, when I use Wry on Windows and call `.with_extension_path()` to load this extension, it fails. Additionally, the extension folder (olofadcdnkkjdfgjcmjaadnlehnnihnl) does not appear in the test\EBWebView\Default\Extensions directory.

When I attempt to enable both `.with_browser_extensions_enabled(true)` and `.with_extension_path(PathBuf::from("Extensions/6.6.3_0"))` simultaneously, it results in the following error:
`WebView2Error(WindowsError(Error { code: HRESULT(0x80004003), message: "Invalid pointer" })).`

Even when I only enable` .with_browser_extensions_enabled(true)`, the extension still does not work.

Of course, I’ve also tried using the `with_additional_browser_args `function, but the results were equally disappointing.

By the way, I noticed that using the` --proxy-server `argument in the browser arguments works the same as the `with_proxy_config `function. Was this intentionally designed to allow --proxy-server to work, or was it simply overlooked and not filtered out?

How can I resolve this issue? Thank you again for helping me.

SpikeHD:
I've done some digging, and it seems `ICoreWebView2Profile7::AddBrowserExtension` doesn't like being passed `None` as the completion handler, which is what's causing `Invalid pointer`.

Changing that to have an empty handler function fixes your example:

`src/webview2/mod.rs L1195`
```rust
    // Iterate over all folders in the extension path
    for entry in fs::read_dir(extension_path)? {
      let path = entry?.path();
      let path_hs = HSTRING::from(path.as_path());
      let handler = ProfileAddBrowserExtensionCompletedHandler::create(Box::new(|_, _| Ok(())));

      profile.AddBrowserExtension(&path_hs, &handler)?;
    }
```

Keep in mind @zhangsan2lisi that the path you are providing should be a *folder* where all extensions reside, even if it is just one.

For example, this would fail:

```
extension/
├─ manifest.json
```
```rust
let webview = WebViewBuilder::with_web_context(&mut context)
    .with_extension_path(PathBuf::from("extension"))
```
but this would succeed:
```
extensions/
├─ my_extension/
│  ├─ manifest.json
```
```rust
let webview = WebViewBuilder::with_web_context(&mut context)
    .with_extension_path(PathBuf::from("extensions"))
```
because the function reads the files/folders within the provided directory as extensions, not the folder itself.

I will make a PR soon to address the `Invalid Pointer` issue, as well as rename the function to `with_extensions_path` so it is less misleading/confusing.

zhangsan2lisi:
@SpikeHD  Thank you for your help. I have successfully used it.

PR #1421, opened 2024-11-19
fix: Prevent `Invalid Pointer` when loading extensions on Windows, rename `with_extension_path` -> `with_extensions_path`

SpikeHD:
Fixes loading extensions on Windows by providing a handler function to `AddBrowserExtension`, rather than `None`, which causes a `Invalid Pointer` error.

Also renames the function `with_extension_path` -> `with_extensions_path`, as the name might be misleading/confusing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1302,7 +1302,7 @@ pub trait WebViewBuilderExtWindows {
   /// Set the path from which to load extensions from. Extensions stored in this path should be unpacked.
   ///
   /// Does nothing if browser extensions are disabled. See [`with_browser_extensions_enabled`](Self::with_browser_extensions_enabled)
-  fn with_extension_path(self, path: impl Into<PathBuf>) -> Self;
+  fn with_extensions_path(self, path: impl Into<PathBuf>) -> Self;
 }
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1302,7 +1302,7 @@ pub trait WebViewBuilderExtWindows {
   /// Set the path from which to load extensions from. Extensions stored in this path should be unpacked.
   ///
   /// Does nothing if browser extensions are disabled. See [`with_browser_extensions_enabled`](Self::with_browser_extensions_enabled)
-  fn with_extension_path(self, path: impl Into<PathBuf>) -> Self;
+  fn with_extensions_path(self, path: impl Into<PathBuf>) -> Self;
```
amrbashir: let's revert this naming change so I can release the fix as a patch version, we can rename in another PR afterwards.

amrbashir (changes requested):

src/lib.rs
SpikeHD: Gotcha 👍

amrbashir (approved):

PR #1425, opened 2024-11-23
fix(macOS): fix race condition in webview handler

linw1995:
https://github.com/tauri-apps/tauri/issues/5519#issuecomment-2495380378

### DIFF
--- a/src/wkwebview/class/url_scheme_handler.rs
+++ b/src/wkwebview/class/url_scheme_handler.rs
@@ -261,8 +261,15 @@ extern "C" fn start_task(
                 }))
                 .unwrap();
 
-                webview.remove_custom_task_key(task_key);
-                Ok(())
+                {
+                  let ids = WEBVIEW_IDS.lock().unwrap();
+                  if ids.contains(webview_id) {
+                    webview.remove_custom_task_key(task_key);
+                    Ok(())
+                  } else {
+                    Err(crate::Error::CustomProtocolTaskInvalid)
+                  }
+                }
               }
 
               let _ = response(


FabianLars:
Does this really help? We already check this 2 instructions above in check_webview_id_valid 🤔

linw1995:
> Does this really help? We already check this 2 instructions above in check_webview_id_valid 🤔

work like a charm. 

```plain
  14:        0x103430874 - core::panicking::panic_fmt::ha4b80a05b9fff47a
                               at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/panicking.rs:74:14
  15:        0x102fa8054 - objc2::__macro_helpers::declared_ivars::get_initialized_ivar_ptr::h92376258763b2729
                               at /Users/linw1995/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.5.2/src/__macro_helpers/declared_ivars.rs:403:17
  16:        0x102f9dc44 - objc2::top_level_traits::DeclaredClass::ivars_mut::h2f3cb44ff3ce2dab
                               at /Users/linw1995/.cargo/registry/src/index.crates.io-6f17d22bba15001f/objc2-0.5.2/src/top_level_traits.rs:355:34
  17:        0x102fa0474 - wry::wkwebview::class::wry_web_view::WryWebView::remove_custom_task_key::hd2a6c5fea7c15784
                               at /Users/linw1995/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wry-0.47.1/src/wkwebview/class/wry_web_view.rs:149:5
```

Simply performing checks is not enough; it needs to be locked to prevent race conditions.

pewsheen:
I think this will mitigate the problem, however it could still happen if the webview is closed just before calling `remove_custom_task_key`.

MacOS expects us to capture the NSExecption to tell if the task is dead, but that's impossible for now. The async command will be unreliable on MacOS until Objc2 can bubble up an NSException instead of panicking so that we can remove those weird checks.

pewsheen (approved):

lucasfernog-crabnebula:
> I think this will mitigate the problem, however it could still happen if the webview is closed just before calling `remove_custom_task_key`.
> 
> MacOS expects us to capture the NSExecption to tell if the task is dead, but that's impossible for now. The async command will be unreliable on MacOS until Objc2 can bubble up an NSException instead of panicking so that we can remove those weird checks.

since we're holding the WEBVIEW_IDS lock this essentially prevents the webview from closing. I also tried without the lock and it didn't crash FYI

PR #1440, opened 2024-12-16
Improve custom URL scheme handler robustness

UdaraJay:
Fix for `wry_web_view.rs/get_custom_task_uuid` trying to access stale `WryWebView`. 

```
thread 'tokio-runtime-worker' panicked at .../wry/src/wkwebview/class/wry_web_view.rs:152:10:
tried to access uninitialized instance variable
```
This happens sometimes when closing a `WebviewWindow` or calling `win.close()`. It causes a runtime panic and crashes the application when it occurs. Fixed by adding runtime checks to make sure `webview_id` and `task_uuid` are valid before they are used.

Potentially related: https://github.com/tauri-apps/wry/pull/1425.

### DIFF
--- a/src/wkwebview/class/url_scheme_handler.rs
+++ b/src/wkwebview/class/url_scheme_handler.rs
@@ -59,8 +59,8 @@ extern "C" fn start_task(
 ) {
   unsafe {
     #[cfg(feature = "tracing")]
-          let span = tracing::info_span!(parent: None, "wry::custom_protocol::handle", uri = tracing::field::Empty)
-            .entered();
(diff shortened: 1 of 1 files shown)


zakhenry:
I can confirm this patch fixed a segfault in macos for me too. For me this occurred during shutdown of the application after closing windows and invoking `app.run_iteration(...)`

FabianLars (commented):
This should be fairly safe to merge as the only logic changes are the upfront validation check (that i don't think has that much of an impact but as long as we're not hitting issues with the frequency of the mutex locks we should keep it), and the removal of the unwrap which i'm always a fan of.

The only thing i'd like to change is to bring back the removed comments.

FabianLars (approved):

Issue #1246, opened 2024-05-09
Allow disabling background throttling/suspension on macOS


venkr:
`WKPreferences` can be passed a `inactiveSchedulingPolicy` (see: [Apple docs](https://developer.apple.com/documentation/webkit/wkpreferences/4173317-inactiveschedulingpolicy?changes=_9_5&language=objc)) 

This would fix https://github.com/tauri-apps/tauri/issues/5250 on macOS. Currently, the default policy is `suspend` which may affect various tasks when backgrounded.

pewsheen:
Closed: #1445 fixed the issue.

PR #1445, opened 2025-01-03
feat: disable background throttling

bastiankistner:
This PR relates to https://github.com/tauri-apps/tauri/issues/5250

It requires a change of tauri for which I've opened another PR: https://github.com/tauri-apps/tauri/pull/12181

I've added an option `disableBackgroundThrottling` to allow developers to choose if their view should automatically be throttled / suspended by the browser after a certain time of being hidden or minimised. In the worst case, this default behaviour of browsers can lead to broken applications running in the background. 

For instance if a websocket connection is the only reason not to suspend a "tab" and the machine goes into sleep mode and wakes up again, the connection is often lost and the browser won't have any reason to keep the background tab active and therefore unload it, which will not allow the hidden view to "heal" itself. 

This default behaviour effectively renders any browser capabilities unusable for long running background processes. One alternative might be moving logic from the frontend to the backend, but browsers have meanwhile become a great choice for local first apps and if the browser already offers the required features easily accessible, why force developers to leave a territory they feel comfortable with?

As of now, this is only available for WebKit based browsers on iOS 17.0+,iPadOS 17.0+,Mac Catalyst 17.0+, macOS 14.0+, visionOS 1.0+ ( see https://developer.apple.com/documentation/webkit/wkpreferences/inactiveschedulingpolicy-swift.enum?language=objc ).

It can be assumed that it will also be implemented for WebView2 (Windows). And it might be that it's already usable for linux and android, although I couldn't find any documentation about it and also cannot test it locally due to a lack of devices. 

As an alternative for windows and non-webkit browsers, one can setup a pending WebLock transaction, which, according to the docs, can also keep the browser from suspending it.

----

I needed to update a snapshot test and am unsure if this is at all related to this change. Please let me know if I did something wrong here.

### DIFF
--- a/.changes/disable-background-throttling.md
+++ b/.changes/disable-background-throttling.md
@@ -1,5 +1,5 @@
 ---
-'wry': 'minor:enhance'
+'wry': 'patch:enhance'
 ---
 
 Add an option to disable background throttling (currently for WebKit only).
--- a/.changes/disable-background-throttling.md
+++ b/.changes/disable-background-throttling.md
@@ -1,0 +1,5 @@
+---
+"wry": minor
+---
+
+Add an option to disable background throttling (currently for WebKit only).
--- a/.changes/disable-background-throttling.md
+++ b/.changes/disable-background-throttling.md
@@ -2,4 +2,4 @@
 'wry': 'minor:enhance'
 ---
 
-Add an option to disable background throttling (currently for WebKit only).
+Add an option to change the default background throttling policy (currently for WebKit only).
--- a/.changes/disable-background-throttling.md
+++ b/.changes/disable-background-throttling.md
@@ -1,5 +1,5 @@
 ---
-"wry": minor
+'wry': 'minor:enhance'
 ---
 
 Add an option to disable background throttling (currently for WebKit only).
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1210,9 +1210,9 @@ impl<'a> WebViewBuilder<'a> {
   /// - **macOS**: Supported since version 14.0+.
   ///
   /// see https://github.com/tauri-apps/tauri/issues/5250#issuecomment-2569380578
-  pub fn with_background_throttling(self, policy: Option<BackgroundThrottlingPolicy>) -> Self {
+  pub fn with_background_throttling(self, policy: BackgroundThrottlingPolicy) -> Self {
     self.and_then(|mut b| {
-      b.attrs.background_throttling = policy;
+      b.attrs.background_throttling = Some(policy);
       Ok(b)
     })
   }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -574,6 +574,18 @@ pub struct WebViewAttributes<'a> {
   /// This is only effective if the webview was created by [`WebView::new_as_child`] or [`WebViewBuilder::new_as_child`]
   /// or on Linux, if was created by [`WebViewExtUnix::new_gtk`] or [`WebViewBuilderExtUnix::new_gtk`] with [`gtk::Fixed`].
   pub bounds: Option<Rect>,
+
+  /// Whether background throttling should be disabled.
+  ///
+  /// By default, browsers throttle timers and even unload the whole tab (view) to free resources after roughly 5 minutes when
+  /// a view became minimized or hidden. This will permanently suspend all tasks until the documents visibility state
+  /// changes back from hidden to visible by bringing the view back to the foreground.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / Windows / Android**: Unsupported yet. But workarounds like a pending WebLock transaction might suffice.
+  /// see https://github.com/tauri-apps/tauri/issues/5250#issuecomment-2569380578
+  pub disable_background_throttling: bool,
 }
 
 impl<'a> Default for WebViewAttributes<'a> {
@@ -614,6 +626,7 @@ impl<'a> Default for WebViewAttributes<'a> {
         position: dpi::LogicalPosition::new(0, 0).into(),
         size: dpi::LogicalSize::new(200, 200).into(),
       }),
+      disable_background_throttling: false,
     }
   }
 }
@@ -1177,6 +1190,23 @@ impl<'a> WebViewBuilder<'a> {
   pub fn with_bounds(self, bounds: Rect) -> Self {
     self.and_then(|mut b| {
       b.attrs.bounds = Some(bounds);
+      Ok(b)
+    })
+  }
+
+  /// Set whether background throttling should be disabled.
+  ///
+  /// By default, browsers throttle timers and even unload the whole tab (view) to free resources after roughly 5 minutes when
+  /// a view became minimized or hidden. This will permanently suspend all tasks until the documents visibility state
+  /// changes back from hidden to visible by bringing the view back to the foreground.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / Windows / Android**: Unsupported yet. But workarounds like a pending WebLock transaction might suffice.
+  /// see https://github.com/tauri-apps/tauri/issues/5250#issuecomment-2569380578
+  pub fn with_disable_background_throttling(self, disable: bool) -> Self {
+    self.and_then(|mut b| {
+      b.attrs.disable_background_throttling = disable;
       Ok(b)
     })
   }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -583,7 +583,9 @@ pub struct WebViewAttributes<'a> {
   ///
   /// ## Platform-specific
   ///
-  /// - **Linux / Windows / Android**: Unsupported yet. But workarounds like a pending WebLock transaction might suffice.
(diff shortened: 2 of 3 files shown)


src/wkwebview/mod.rs
```diff
@@ -347,6 +347,18 @@ impl InnerWebView {
         webview
       };
 
+      // disable background throttling if attributes.disable_background_throttling is true
+      // which works for iOS 17.0+,iPadOS 17.0+,Mac Catalyst 17.0+, macOS 14.0+, visionOS 1.0+
+      if attributes.disable_background_throttling {
```
pewsheen: We may need to add a version check before setting the preference value in case an unsupported version will panic.

Reference:
https://github.com/tauri-apps/wry/blob/2ffcd494af480ddbade3d7568e4d133ba519d83f/src/wkwebview/mod.rs#L195

pewsheen (changes requested):

src/wkwebview/mod.rs
bastiankistner: Will look into that tomorrow! I assumed that if there is an unsupported version, it'll just be ignored. But that was just a (bad) guess. Thank you very much @pewsheen !

FabianLars: Sometimes that's the case, sometimes not 😂 In this case i would assume it'll be ignored as well. I have a vm with something older than 14 that i could check tomorrow.

pewsheen: And I think the attribute will affect iOS as well, so we might need to validate it on iOS, too 👀

FabianLars: @pewsheen was correct. Without a guard it panics on macos 12

bastiankistner: @FabianLars thank you for testing! I've added a version check

src/lib.rs
```diff
@@ -583,7 +583,9 @@ pub struct WebViewAttributes<'a> {
   ///
   /// ## Platform-specific
   ///
-  /// - **Linux / Windows / Android**: Unsupported yet. But workarounds like a pending WebLock transaction might suffice.
+  /// - **Linux / Windows / Android**: Unsupported. Workarounds like a pending WebLock transaction might suffice.
+  /// - **iOS / macOS**: Supported since version 17.0+.
```
FabianLars: wrong version for macos

src/lib.rs
```diff
@@ -1202,7 +1204,9 @@ impl<'a> WebViewBuilder<'a> {
...
```
FabianLars: same

src/wkwebview/mod.rs
```diff
@@ -349,14 +349,20 @@ impl InnerWebView {
...
-        // Set inactive scheduling policy to None enum value (2)
-        _preference.setValue_forKey(
-          Some(&NSNumber::numberWithInt(
-            WKInactiveSchedulingPolicy::None.0.try_into().unwrap(),
-          )),
-          ns_string!("inactiveSchedulingPolicy"),
-        );
+      #[cfg(any(target_os = "ios", target_os = "macos"))]
+      {
+        let is_supported_os = (cfg!(target_os = "ios") && os_major_version >= 17)
```
FabianLars, line 356: we _could_ replace `is_supported_os` with `custom_data_store_available` from above since that has the same version requirement but i'm not sure if that's good for visibility 🤔

FabianLars (commented):
just a minor doc thing left

src/lib.rs
bastiankistner: My bad, apologies. Also need to change that in tauri. Thanks a lot.

src/wkwebview/mod.rs
bastiankistner: I can do so, but indeed, that could leave the impression that it has something to do with the custom data store.

.changes/disable-background-throttling.md
```diff
@@ -0,0 +1,5 @@
+---
+'wry': 'minor:enhance'
```
FabianLars: ah sorry, last one. this can be a patch because wry is still 0.x

FabianLars (dismissed):

src/lib.rs
```diff
@@ -574,6 +574,21 @@ pub struct WebViewAttributes<'a> {
...
+  /// changes back from hidden to visible by bringing the view back to the foreground.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / Windows / Android**: Unsupported. Workarounds like a pending WebLock transaction might suffice.
+  /// - **iOS**: Supported since version 17.0+.
+  /// - **macOS**: Supported since version 14.0+.
+  ///
+  /// see https://github.com/tauri-apps/tauri/issues/5250#issuecomment-2569380578
+  pub disable_background_throttling: bool,
```
pewsheen: I'm thinking if we should use an enum such as `BackgroundThrottlePolicy` here, just like: https://developer.apple.com/documentation/webkit/wkpreferences/inactiveschedulingpolicy-swift.enum

It would give more flexibility to the developers.

And this property could be renamed to something like `background_throttling`

pewsheen (commented):
Sorry for the late question, but I think it would be nice to give it more flexibility.

Because it defaults to `suspend` instead of `throttling` on macOS, there might be developers who want to set it to `throttling`. Also, if we want to support other platforms in the future, it may require a breaking change to this property to support different policies.

pewsheen:
Related issue: https://github.com/tauri-apps/wry/issues/1246

bastiankistner:
@pewsheen @FabianLars I changed it to use a policy enum of 
- `disabled`
- `throttle`
- `suspend`

I tested it successfully with my app and it seems to work. But I had to add the dependency `wry` to tauri-utils to implement a From trait. I'm not very experienced with rust, it's been a while... So please let me know if there is anything I should improve.

I also updated the tauri PR.

src/lib.rs
```diff
@@ -1997,6 +2033,17 @@ pub enum PageLoadEvent {
   Finished,
 }
 
+/// Background throttling policy
+#[derive(Debug, Clone)]
+pub enum BackgroundThrottlingPolicy {
```
FabianLars, line 2038: @pewsheen Should this be non_exhaustive from the start if future compat was the main concern for the enum?

pewsheen: It should be okay for now, as long as the client won't build fail after we add policies.
And Wry will build fail if we forget to handle the new policy.

src/lib.rs
```diff
@@ -1181,6 +1197,26 @@ impl<'a> WebViewBuilder<'a> {
...
+  /// - **iOS**: Supported since version 17.0+.
+  /// - **macOS**: Supported since version 14.0+.
+  ///
+  /// see https://github.com/tauri-apps/tauri/issues/5250#issuecomment-2569380578
+  pub fn with_background_throttling(self, policy: Option<BackgroundThrottlingPolicy>) -> Self {
+    self.and_then(|mut b| {
+      b.attrs.background_throttling = policy;
+      Ok(b)
+    })
+  }
```
pewsheen:
```suggestion
  pub fn with_background_throttling(self, policy: BackgroundThrottlingPolicy) -> Self {
    self.and_then(|mut b| {
      b.attrs.background_throttling = Some(policy);
      Ok(b)
    })
  }
```

I assume the user calls `with_background_throttling` when they decide to set a non-default value, so `Option<>` may not be necessary here.

pewsheen (changes requested):

src/lib.rs
bastiankistner: Or we go back to the boolean `disableBackgroundThrottling` , [like electron does](https://www.electronjs.org/docs/latest/api/browser-window#:~:text=8859%2D1.-,backgroundThrottling,-boolean%20(optional)%20%2D%20Whether). And maybe add a `backgroundThrottlingPolicy` later that could be used for advanced cases. But then there'd be a dependency of the policy on the disable flag.

I would assume that almost no developer would decide for a semi-throttled/semi-unthrottled mode. In that case, they should rather focus on optimising the usage of timers in their code and keep it fully disabled.

pewsheen: > Or we go back to the boolean `disableBackgroundThrottling` , [like electron does](https://www.electronjs.org/docs/latest/api/browser-window#:~:text=8859%2D1.-,backgroundThrottling,-boolean%20(optional)%20%2D%20Whether). And maybe add a `backgroundThrottlingPolicy` later that could be used for advanced cases. But then there'd be a dependency of the policy on the disable flag.
> 
> I would assume that almost no developer would decide for a semi-throttled/semi-unthrottled mode. In that case, they should rather focus on optimising the usage of timers in their code and keep it fully disabled.

Because Wry is an underlying library, I prefer to give higher-level libs/applications more flexibility in designing their API.
For example, we could add `disableBackgroundThrottling` in Tauri as a public API, and let tauri-wry-runtime decide which policy should be used.

pewsheen (approved):

src/lib.rs
bastiankistner: @pewsheen changed it accordingly, thank you!

pewsheen:
@bastiankistner  Cool! Thanks for your contribution! ❤️

bastiankistner:
Thank you for your help!

PR #1446, opened 2025-01-06
feat: Add traffic light inset

### LINKED ISSUE #1056: Traffic light inset not working with wry webview
## Description

When applying `with_traffic_light_inset(...)` method to reexported tao windowbuilder without passing it to a webview the macos traffic lights are inset as expected (of course the webview isn't working because it is never called). But when using the method on the windowbuilder and then passing the newly built window to `WebviewBuilder::new(...)` no traffic light inset can be observed. Everything looks like `with_traffic_light_inset(...)` was never called.

I.e.

```rust
    .with_title("Hello World")
    .with_traffic_light_inset(Position::Logical(LogicalPosition::new(32., 60.)))
    .build(&event_loop)?;
  let _webview = WebViewBuilder::new(window)?
    .with_url("https://www.netflix.com/browse")?
    .build()?;
```

works when commenting out the webview part but doesn't when constructing the webview. `set_traffic_light_inset(...)` on `window` doesn't work too.

## Steps To Reproduce

Use the `simple` example and add the `.with_traffic_light_inset(Position::Logical(LogicalPosition::new(32., 60.)))` to the tao `WindowBuilder` call chain. 

**Expected behavior**

Works only on macos: I expect the traffic lights to be inset on the webview wry window.

Platform and Versions (please complete the following information):

OS: OSX Ventura (not 100% sure)
Rustc: 1.73 (I don't have my mac here but I'm pretty sure it was the latest version)

**Would you want to assign yourself to resolve this bug?**
- [x] Yes
- [ ] No

related to [this PR](https://github.com/tauri-apps/tauri/issues/4789)

FabianLars:
fixes #1056 

I really wanted to fix this in tao or maybe in tauri but i just wasn't able to fix it in tao and in tauri it would have resulted in much more code (overwriting delegates and shit, see https://github.com/tauri-apps/tauri/issues/4789

### DIFF
--- a/src/wkwebview/class/wry_web_view_parent.rs
+++ b/src/wkwebview/class/wry_web_view_parent.rs
@@ -69,19 +69,17 @@ impl WryWebViewParent {
     unsafe { msg_send_id![super(delegate), init] }
   }
 
+  #[cfg(target_os = "macos")]
   pub fn set_traffic_light_inset(&self, ns_window: &NSWindow, position: dpi::Position) {
-    #[cfg(target_os = "macos")]
-    {
-      let scale_factor = NSWindow::backingScaleFactor(ns_window);
-      let position = position.to_logical(scale_factor);
-      self
-        .ivars()
-        .traffic_light_inset
-        .replace(Some((position.x, position.y)));
+    let scale_factor = NSWindow::backingScaleFactor(ns_window);
+    let position = position.to_logical(scale_factor);
+    self
+      .ivars()
+      .traffic_light_inset
+      .replace(Some((position.x, position.y)));
 
-      unsafe {
-        inset_traffic_lights(ns_window, position.x, position.y);
-      }
+    unsafe {
+      inset_traffic_lights(ns_window, position.x, position.y);
     }
   }
 }
--- a/src/wkwebview/class/wry_web_view_parent.rs
+++ b/src/wkwebview/class/wry_web_view_parent.rs
@@ -2,16 +2,21 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
+use std::cell::Cell;
+
 use objc2::{
   declare_class, msg_send_id, mutability::MainThreadOnly, rc::Retained, ClassType, DeclaredClass,
 };
 #[cfg(target_os = "macos")]
 use objc2_app_kit::{NSApplication, NSEvent, NSView};
-use objc2_foundation::MainThreadMarker;
+use objc2_app_kit::{NSWindow, NSWindowButton};
+use objc2_foundation::{MainThreadMarker, NSRect};
 #[cfg(target_os = "ios")]
 use objc2_ui_kit::UIView as NSView;
 
-pub struct WryWebViewParentIvars {}
+pub struct WryWebViewParentIvars {
+  traffic_light_inset: Cell<Option<(f64, f64)>>,
+}
 
 declare_class!(
   pub struct WryWebViewParent;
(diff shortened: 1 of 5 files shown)


src/wkwebview/class/wry_web_view_parent.rs
```diff
@@ -41,15 +47,72 @@ declare_class!(
...
+        unsafe {inset_traffic_lights(&self.window().unwrap(), x, y)};
+      }
+    }
   }
 );
 
 impl WryWebViewParent {
   #[allow(dead_code)]
-  pub fn new(mtm: MainThreadMarker) -> Retained<Self> {
+  pub fn new(mtm: MainThreadMarker, traffic_light_inset: Option<(f64, f64)>) -> Retained<Self> {
```
pewsheen: hmm.. I would prefer not to list `traffic_light_inset` as a parameter and we can call `set_traffic_light_inset` after `NSView` created?

FabianLars:
i will open the tauri pr tomorrow

src/wkwebview/class/wry_web_view_parent.rs
```diff
@@ -2,16 +2,22 @@
...
+use objc2_app_kit::{NSApplication, NSEvent, NSView, NSWindow, NSWindowButton};
 use objc2_foundation::MainThreadMarker;
+#[cfg(target_os = "macos")]
+use objc2_foundation::NSRect;
 #[cfg(target_os = "ios")]
 use objc2_ui_kit::UIView as NSView;
 
-pub struct WryWebViewParentIvars {}
+pub struct WryWebViewParentIvars {
+  traffic_light_inset: Cell<Option<(f64, f64)>>,
```
pewsheen, line 20: Maybe add `#[cfg(target_os = "macos")]` here as well?

src/wkwebview/mod.rs
```diff
@@ -504,19 +507,30 @@ r#"Object.defineProperty(window, 'ipc', {
...
+          let scale_factor = ns_window.backingScaleFactor();
+
+          let parent_view = WryWebViewParent::new(
+            mtm,
+            pl_attrs
+              .traffic_light_inset
+              .map(|p| p.to_logical(scale_factor))
+              .map(|p| (p.x, p.y)),
+          );
+
```
pewsheen: How about we remove the inset from the parameter and set it after creation?

```rust
if let Some(inset) = pl_attrs.traffic_light_inset {
  parent_view.set_traffic_light_inset(&ns_window, inset);
}
```

pewsheen (commented):
Also the options seemed to conflict with `with_decoration` in Tao. I think we should mention that the app would panic if the title bar doesn't exist.

~Another question is whether the setting works on multiple monitors with different DPI?~
It looks good on different DPI

src/wkwebview/class/wry_web_view_parent.rs
FabianLars: that's what i did or wanted to do initially. can't remember how i ended up with this tbh. I'll check if that works just as well.

src/wkwebview/class/wry_web_view_parent.rs
FabianLars: this whole file isn't used on ios. i assume it was used in the past hence the target_os stuff.

anyway, i can add it for future compat in case we need it on ios at some point.

src/wkwebview/mod.rs
FabianLars: Just to confirm, this is the same as your other comment here https://github.com/tauri-apps/wry/pull/1446#discussion_r1904452162 right?

FabianLars:
> Also the options seemed to conflict with with_decoration in Tao. I think we should mention that the app would panic if the title bar doesn't exist.

Oh yes, absolutely. It requires a hidden title setting (which also means decorations are required). I meant to ~~copy paste the electron docs~~ document this but forgot about it because of another doc change.

src/wkwebview/mod.rs
pewsheen: yes, its the same

pewsheen (approved):

haasal:
Hi, I am the original guy who tried to work on this. I have no idea what you did to fix this but I thank you deeply from my heart. I just couldn't look at this mess any longer without becoming anxious 😅

**More importantly**: I have lost track of this whole thing. Does my my kind of broken inset method still exist in Tao? If yes we should probably remove it right?

FabianLars:
I couldn't have done it without you 🫂 

And no, we still need the Tao method. The method is also no broken (after all i use the very same impl here). It's still needed if the webview is injected via build_as_child (for multiwebview support for example). We probablyyy could overwrite this here as well, but imo windowing stuff should stay in the windowing library as much as possible.
Additionally, we still need the Tao implementation for what i fixed here as well, otherwise we would end up with the same flicker (but kinda in the opposite way). At least it did in my earlier tests.

PR #1451, opened 2025-01-08
fix(windows): fix Using the Weasel input method in the always-on-top window will get stuck

CDHJS:
ref tauri-apps/tauri#6879
rime/weasel#1435

Referenced the implementation from
[MicrosoftEdge/WebView2Samples/SampleApps/WebView2APISample/ViewComponent.cpp](https://github.com/MicrosoftEdge/WebView2Samples/blob/96ef509eb5f7e13e31a2e7b4c8d561310556490c/SampleApps/WebView2APISample/ViewComponent.cpp#L399-L405)
 lines 399 to 405.


CDHJS:
There might be other tasks to do, but I find it difficult to read Tao code.

amrbashir (approved):

PR #1452, opened 2025-01-11
fix(windows): Fix the issue where TAB cannot cycle through focus elements.

mzdk100:
This PR fixes a minor issue, which is discussed [here](https://github.com/DioxusLabs/dioxus/issues/2900):
Specifically, when a user switches focus within a web element using the keyboard's tab, if they continue to press the tab after switching to the last web element, the focus will jump to the control with the class name WRY_WEBVIEW. At this point, no matter how much they press the tab, they cannot continue switching web elements;
The solution to this problem is to first switch to another application (such as the desktop), and then switch to the dioxus application again. At this point, the system will reset the focus to the web document, and you can continue to use the tab to switch elements, but this requires the user to operate it themselves, and the experience is very poor.

Expected behavior
The control with the class name WRY_WEBVIEW should not be allowed to grab keyboard focus, so that the tab key can be used to cyclically switch web elements.

### DIFF
--- a/src/webview2/mod.rs
+++ b/src/webview2/mod.rs
@@ -179,8 +179,8 @@ impl InnerWebView {
       if msg == WM_SETFOCUS {
         // Fix https://github.com/DioxusLabs/dioxus/issues/2900
         // Get the first child window of the window
-        let child = GetWindow(hwnd, GW_CHILD).unwrap_or_default();
-        if child != HWND::default() {
+        let child = GetWindow(hwnd, GW_CHILD).ok();
+        if child.is_some() {
           // Set focus to the child window(WebView document)
           let _ = SetFocus(child);
         }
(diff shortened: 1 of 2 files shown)


src/webview2/mod.rs
```diff
@@ -176,6 +176,16 @@ impl InnerWebView {
       wparam: WPARAM,
       lparam: LPARAM,
     ) -> LRESULT {
+      if msg == WM_SETFOCUS {
+        // Fix https://github.com/DioxusLabs/dioxus/issues/2900
+        // Get the first child window of the window
```
FabianLars, line 181: This doesn't sound like the correct behavior for multiwebview windows but i don't reallyyy care about that at the moment and it's still better than losing focus completely.

FabianLars (approved):

mzdk100:
Yes, I will further fix the focus issue with multiwebview windows in the future, thank you.

PR #1470, opened 2025-01-27
chore: Set and test `rust-version`

### LINKED ISSUE #1469: document and test MSRV
Just noticed that after the `This bumps MSRV on macOS/iOS to 1.71.` comment in #1468. With the new msrv aware resolver we should set `rust-version` in all our crates.

Do we even have a msrv for wry in mind? Or should we just use 1.77 like in tauri?

FabianLars:
gonna check via ci if we can maybe use something lower than 1.77 

closes #1469


FabianLars:
okay nevermind then 😂 I mean, it's just a dev dependency but still

pewsheen (approved):

pewsheen:
Should this apply to tao as well?

FabianLars:
probably, yeah. Or really all our crates tbh. I'll open a few PRs later.

PR #1480, opened 2025-02-07
fix(android): Call toString on versionName

ahqsoftwares:
ref https://github.com/tauri-apps/tauri/issues/12412


FabianLars (dismissed):

ahqsoftwares:
@FabianLars Verified that the issue has been fixed after adding the forked wry in patch

FabianLars (approved):

PR #1484, opened 2025-02-10
fix(macOS): race conditions around WKURLSchemeTask

### LINKED ISSUE #1189: [macOS] APP crash when a web custom protocol request is aborted before or during response

alexmoon:
Fixes two issues which could cause race conditions when navigating between pages while custom protocol requests are in progress on macOS and iOS:
- The `obc2/catch-all` feature turns all Objective-C exceptions into panics, preventing `obc2::exception::catch(...)` from working.
- The `webview` and `task` objects passed to `start_task` were not retained, leading to use-after-free issues in the `responder`.
- `WryWebView::remove_custom_task_key` is called from both `stop_task` and `response` which may be in different threads, so `custom_protocol_task_ids` needs to be protected by a `Mutex`.

Closes #1189


Brendonovich:
For us, this PR fixes the non-fatal 'unintialized instance variable' panics we've been getting. We're still getting crashes from different places but this is helpful.

pewsheen (approved):
Tested:
- https://github.com/tauri-apps/tauri/issues/9933
- https://github.com/tauri-apps/wry/pull/1214

Btw, if we can ensure `task` lives through the responder, I think we could also remove `check_task_is_valid` as well.

alexmoon:
> Btw, if we can ensure `task` lives through the responder, I think we could also remove `check_task_is_valid` as well.

Yes, I think those checks are likely superfluous now, since the `task` is retained for the lifetime of the responder. I didn't want to risk introducing a different bug, so I left them in.

PR #1503, opened 2025-02-26
docs: `WebViewBuilder::with_asynchronous_custom_protocol` - copy over important missing docs that the custom uri schemes are implemented differently on windows and android

Simon-Laux:
Took me quite a while to figure this out, this should help others to understand it faster than me.

pr for the tauri side of things: https://github.com/tauri-apps/tauri/pull/12830


amrbashir (approved):
Thank you

PR #1504, opened 2025-02-28
fix(linux): fix examples/gtk_multiwebview.rs layout misaligned

ohyfun:
GTK requires a single gtk::Fixed container as the parent for all WebView instances.</br>
However, the current code creates a new gtk::Fixed for each WebView, leading to incorrect layout behavior.</br>

before 
befor
after
after


amrbashir (approved):

amrbashir:
Thank you

PR #1505, opened 2025-02-28
fix(android): BorrowError crash when loading assets on Android

lucasfernog:
There is a race condition on the Android asset loading where the custom asset loading function is checked by the Java side while Rust is still setting it up, leading to a `already mutably borrowed: BorrowError` panic in the withAssetLoader function. 
Using Mutex instead of RefCell is the safer approach. Performance impact shouldn't be relevant here, since it's only a small lookup overhead every IPC and asset loading call.

ref https://github.com/tauri-apps/tauri/issues/12812

### DIFF
--- a/.changes/fix-android-race-condition.md
+++ b/.changes/fix-android-race-condition.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix `already mutably borrowed: BorrowError` panic on webview initialization on Android.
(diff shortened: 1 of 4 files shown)


lucasfernog:
this is a bit difficult to reproduce, but it looks like it's not crashing anymore.

lucasfernog:
script i've used to verify it:
```
watch -n 2 'adb shell am force-stop com.tauri.api && sleep 1 && adb shell am start -a android.intent.action.MAIN -n com.tauri.api/.MainActivity'
```
basically reopening the app every 2 seconds to see if it loads properly

lucasfernog:
user reported a different crash, I'll investigate further tomorrow

lucasfernog:
looks like the reported crash is not related to this change (only happens in prod)

PR #1506, opened 2025-02-28
fix crate version in Cargo.lock

Simon-Laux:
this is annoying as rustanalyzer makes the git checkout dirty by updating the file. This change is missing on dev branch.


FabianLars:
That annoys me on every release :/

FabianLars (approved):

PR #1517, opened 2025-03-12
feat(windows): add config to disable context menus

FabianLars:
only opening this because settings changes require a page reload. The other platforms don't have such a setting* so it's in the platform attributes.

`*` Windows and Linux can simulate the same behavior via a context menu event handler but on macos afaik you can only modify the menu, not disable it.

### DIFF
--- a/.changes/default-context-menus.md
+++ b/.changes/default-context-menus.md
@@ -1,0 +1,5 @@
+---
+wry: "patch:enhance"
+---
+
+Added a Windows-only setting to disable all Webview2 context menus.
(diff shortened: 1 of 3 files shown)


amrbashir:
What's the benefit over just listening to `context-menu` event and calling `preventDefault`

FabianLars:
doesn't work if the js or the whole website loads or has not loaded yet. doesn't really care about it but it came up in https://discord.com/channels/616186924390023171/1348377818274336939 so i thought i propose it at least 🤷

amrbashir (approved):

PR #1520, opened 2025-03-15
fix(macos): crash setting traffic lights inset on undecorated window

lucasfernog:


FabianLars:
funny story, i just remembered why i kept the tauri PR on draft for so long

Issue #454, opened 2021-12-11
feat: synchronous communication between webview and rust


amrbashir:
Tauri has a lot of functions in its Js API that can and should be synchronous instead of async, for example the `homeDir` function in tauri's `path` module. Synchronous communication can also help boost the performance since it doesn't include any serializations (I think).

- Webview2:
  - [`AddHostObjectToScript()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2.addhostobjecttoscript?view=webview2-dotnet-1.0.1054.31) exposes full host objects to Js which can be used asynchronously and synchronously. But it's blocked by https://github.com/microsoft/windows-rs/issues/1094. [Here](https://learn.microsoft.com/en-us/microsoft-edge/webview2/how-to/winrt-js-conversion) are the types it can support.
  - [`CreateSharedBuffer`](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2environment?view=webview2-winrt-1.0.1661.34#createsharedbuffer) creates an ArrayBuffer that can be used by webview and rust. But both processes must close it manually to release memory so that it can be error-prone.
- WkWebview:
  - [`didReceiveScriptMessage`](https://developer.apple.com/documentation/webkit/wkscriptmessagehandlerwithreply/3585111-usercontentcontroller?language=objc) can also receive and return certain types.
- Webkit2gtk:
  - [`webkit_user_content_manager_register_script_message_handler_with_reply`](https://webkitgtk.org/reference/webkit2gtk/stable/signal.UserContentManager.script-message-with-reply-received.html) can create and return fully defined JS value.

**Would you assign yourself to implement this feature?**
- [X] Yes (partially)
- [ ] No

wusyong:
>     * Webkit2gtk not sure if it has something similar or not but here is a good [article](https://blogs.igalia.com/carlosgc/2013/09/10/webkit2gtk-web-process-extensions/), maybe?

The article is outdated and, according to webkitgtk dev, they are deprecating web extension. So sad news is we will rely on custom URL scheme unless someone can find a better solution for this.
The good news is webkit2gtk 0.36 will come with headers support!

nklayman:
[...]

PR #1522, opened 2025-03-17
fix: WebView2: disable "host objects"

WofWca:
We don't use them, and Microsoft recommends to disable them
if we don't use them.

This does not fix any known vulnerabilities.
It's just an extra security measure.

- [ ] Please verify that this won't break things. My assumption is that we don't use them in this library, thus it's safe to disable.

See also discussion on host objects:
https://github.com/tauri-apps/wry/issues/454.

### DIFF
--- a/.changes/windows-disable-host-objects.md
+++ b/.changes/windows-disable-host-objects.md
@@ -1,0 +1,5 @@
+---
+wry: "patch:sec"
+---
+
+Windows (WebView2): disable [host objects](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/security#restrict-web-content-functionality) for extra security.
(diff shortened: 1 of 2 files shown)


amrbashir (commented):
I would like to keep this option enabled, so users could still potentially expose host objects if they wish so.

I am not sure I understand if there is any possible risk here, because if you never exposed a host object, it is the same as disabling it.

WofWca:
OK, then how about adding an option `with_host_objects_enabled_on_windows(enabled: bool)`, off by default? Off by default because Wry is supposed to be cross-platform, and this option only affects Windows.

amrbashir:
It should be on by default, otherwise we may break some use-cases for some rare developers that rely on it being active.

We have `WebViewBuilderExtWindow` for windows-specific methods.

WofWca:
OK, I'll close this then.

amrbashir:
I am not opposed to adding the option just saying it should be on by default (for now at least).

WofWca:
Got it.
Either way the MR needs to be re-made.
And this would also need to be added to Tauri then.

PR #1529, opened 2025-03-25
fix(macos): fix unused variables warnings from `cargo check` when `tracing` feature is disabled

rhysd:
I noticed that the HEAD of `dev` branch caused warnings as follows with `cargo check`.

```
warning: unused variable: `e`
   --> src/wkwebview/class/url_scheme_handler.rs:195:26
    |
195 |               if let Err(e) = validate() {
    |                          ^ help: if this is intentional, prefix it with an underscore: `_e`
    |

warning: unused variable: `e`
   --> src/wkwebview/class/url_scheme_handler.rs:307:26
    |
307 |               if let Err(e) = response(
    |                          ^ help: if this is intentional, prefix it with an underscore: `_e`

warning: `wry` (lib) generated 2 warnings
```

This PR fixes the warnings. And it fixes clippy warnings reported on macOS as well.

### DIFF
--- a/.changes/fix-unused-warnings.md
+++ b/.changes/fix-unused-warnings.md
@@ -1,5 +1,0 @@
----
-wry: patch
----
-
-Fix the unused variable warnings from `cargo check` on macOS when `tracing` feature is not enabled.
--- a/.changes/fix-unused-warnings.md
+++ b/.changes/fix-unused-warnings.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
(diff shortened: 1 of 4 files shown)


amrbashir (dismissed):

amrbashir (dismissed):

amrbashir:
Thank you

amrbashir (approved):

rhysd:
https://github.com/tauri-apps/wry/actions/runs/14087133462/job/39454045644

The benchmark job seems to hang at running the first benchmark suite. It works fine on my machine so I cannot investigate the reason. I guess there are some issues related to xvfb.

PR #1418, opened 2024-11-14
fix(windows): fix init scripts running twicec

amrbashir:
closes tauri-apps/plugins-workspace#2051

PR #1531, opened 2025-03-26
fix(windows): init scripts sometimes get executed too late

Legend-Master:
Reference: https://github.com/tauri-apps/wry/pull/1418#discussion_r2005617414, https://github.com/tauri-apps/tauri/issues/12990

- Fix init script sometimes get executed too late on Windows WebView2
- Init scripts are always executed on all frames on Windows WebView2
- `WebViewAttributes::initialization_scripts` takes `Vec::<InitializationScript>` now instead of `Vec::<(String, bool)>`

@amrbashir @tweidinger

### DIFF
--- a/.changes/init-script-struct.md
+++ b/.changes/init-script-struct.md
@@ -1,0 +1,5 @@
+---
+wry: minor
+---
+
+**Breaking Change**: `WebViewAttributes::initialization_scripts` takes `Vec::<InitializationScript>` now instead of `Vec::<(String, bool)>`
--- a/.changes/windows-init-script-main-only.md
+++ b/.changes/windows-init-script-main-only.md
@@ -1,0 +1,5 @@
+---
+wry: minor
+---
+
+Init scripts are always executed on all frames on Windows WebView2
--- a/.changes/windows-init-script-timing.md
+++ b/.changes/windows-init-script-timing.md
@@ -1,0 +1,5 @@
+---
+wry: minor
+---
+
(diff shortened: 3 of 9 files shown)


amrbashir (dismissed):
LGTM

tweidinger:
I think we should preserve the selective injection if possible due to https://github.com/tauri-apps/tauri/security/advisories/GHSA-57fm-592m-34r7 being an issue in the past. Is there another way to fix the issue without reverting @Legend-Master ?

Legend-Master:
I can not think of another way at the moment, the method amr used is what I was thinking as well, but it doesn't quite work the way we expected it to

One possible way is we add in some checks in the javascript we would inject, and let that detect if we're in an iframe or not, but I don't know how feasible it is and I don't really like that to be honest

Maybe we can open a feature request in webview2's feedback repo (seems like there's [one](https://github.com/MicrosoftEdge/WebView2Feedback/issues/4375) already), but that won't fix the problem we currently have

Legend-Master (commented):
Also tried to do it in `add_NavigationStarting` and it didn't work either, I think we probably have to revert this as this is basically breaking every tauri app embedding scripts that relies on tauri's javascript APIs directly inside the html

lucasfernog:
we should really fix this

lucasfernog:
looks like the only way is to indeed inject on all frames

lucasfernog (approved):

PR #1532, opened 2025-03-26
fix(bench): fix building benchmarks

rhysd:
This PR fixes the [benches broken due to recent breaking changes](https://github.com/tauri-apps/wry/actions/workflows/bench.yml).


amrbashir (approved):
Thank you

rhysd:
https://github.com/tauri-apps/wry/actions/runs/14087133462/job/39454045644

The benchmark job seems to hang at running the first benchmark suite. It works fine on my machine so I cannot investigate the reason. I guess there are some issues related to xvfb.

PR #1533, opened 2025-03-26
fix(macos): fix warnings reported from `cargo check --release`

rhysd:
This PR fixes the following two warnings reported by `cargo check --release`.

```
warning: unused import: `NSObject`
  --> src/wkwebview/mod.rs:37:24
   |
37 |   runtime::{AnyObject, NSObject, ProtocolObject},
   |                        ^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused variable: `error`
  --> src/wkwebview/download.rs:90:3
   |
90 |   error: &NSError,
   |   ^^^^^ help: if this is intentional, prefix it with an underscore: `_error`
   |

warning: `wry` (lib) generated 2 warnings (run `cargo fix --lib -p wry` to apply 1 suggestion)
```

These warnings are only caused on release build because some implementation is switched by `debug_assertions` gate. This PR also adds the check by `cargo check --release` in CI workflow so that we can notice warnings.

### DIFF
--- a/.github/workflows/clippy-fmt.yml
+++ b/.github/workflows/clippy-fmt.yml
@@ -37,6 +37,10 @@ jobs:
 
       - run: cargo clippy --all-targets
 
+      - run: cargo check --release
+        env:
+          RUSTFLAGS: -D warnings
+
   fmt:
     runs-on: ubuntu-latest
     steps:
--- a/src/wkwebview/download.rs
+++ b/src/wkwebview/download.rs
@@ -87,13 +87,13 @@ pub(crate) fn download_did_finish(this: &WryDownloadDelegate, download: &WKDownl
 pub(crate) fn download_did_fail(
   this: &WryDownloadDelegate,
   download: &WKDownload,
-  error: &NSError,
+  _error: &NSError,
   _resume_data: &NSData,
 ) {
   unsafe {
     #[cfg(debug_assertions)]
     {
-      let description = error.localizedDescription().to_string();
+      let description = _error.localizedDescription().to_string();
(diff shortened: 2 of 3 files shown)


.github/workflows/clippy-fmt.yml
```diff
@@ -37,6 +37,10 @@ jobs:
 
       - run: cargo clippy --all-targets
 
+      - run: cargo check --release
+        env:
+          RUSTFLAGS: -D warnings
```
FabianLars, line 42: i don't think it's worth to increase CI times for this

rhysd: but we actually didn't notice the warnings because we usually use debug build for development. And users will face the warnings because they build wry in release mode for building their application. If they build release binaries with `RUSTFLAGS="-D warnings"`, it will break their build.

Regarding to the increase, this step only takes [10 seconds](https://github.com/tauri-apps/wry/actions/runs/14089020300/job/39461355828) on CI and I feel it's worth.

rhysd:
I fixed the formatter issue and focue-pushed it.

.github/workflows/clippy-fmt.yml
lucasfernog: you'll only see the warnings if you try to build wry from source instead of the published crate

lucasfernog: let me guess.. `RUSTFLAGS="-D warnings"` comes from https://github.com/actions-rust-lang/setup-rust-toolchain ? I hate that 😂

PR #1534, opened 2025-03-28
feat: macOS/iOS: add option to disable or enable link previews when building a webview

Simon-Laux:
macOS/iOS: add option to disable or enable link previews when building a webview (the webkit api has it enabled by default)
  - `WebViewBuilderExtDarwin.allow_link_preview(allow_link_preview: bool)`

I personally only need this to disable link previews, so I would be fine with setting a different default for tauri (like turning it off by default). Or even always disable it and don't expose api to change it.

- [x] If this change requires a new version, then add a change file in `.changes` directory with the appropriate bump, see https://github.com/tauri-apps/wry/blob/dev/.changes/readme.md
- [x] Ensure that all your commits are signed https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
- [x] Ensure `cargo test` passes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1394,10 +1394,23 @@ impl<'a> WebViewBuilder<'a> {
 }
 
 #[cfg(any(target_os = "macos", target_os = "ios"))]
-#[derive(Clone, Default)]
+#[derive(Clone)]
 pub(crate) struct PlatformSpecificWebViewAttributes {
   data_store_identifier: Option<[u8; 16]>,
   traffic_light_inset: Option<dpi::Position>,
+  allow_link_preview: bool,
+}
+
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+impl Default for PlatformSpecificWebViewAttributes {
+  fn default() -> Self {
+    Self {
+      data_store_identifier: None,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1415,6 +1428,12 @@ pub trait WebViewBuilderExtDarwin {
   /// Warning: Do not use this if your chosen window library does not support traffic light insets.
   /// Warning: Only use this in **decorated** windows with a **hidden titlebar**!
   fn with_traffic_light_inset<P: Into<dpi::Position>>(self, position: P) -> Self;
+  /// Whether to show a link preview when long pressing on links. Available on macOS and iOS only.
+  ///
+  /// Default is true.
+  ///
+  /// See https://docs.rs/objc2-web-kit/latest/objc2_web_kit/struct.WKWebView.html#method.allowsLinkPreview
+  fn allow_link_preview(self, allow_link_preview: bool) -> Self;
```
pewsheen: Can we use `with_link_preview` to align with other builder method's naming style

pewsheen (commented):
Looks good with a nit

lucasfernog (approved):

PR #1538, opened 2025-04-05
fix(macos): Handle flipped coordinate system of a parent view

syrel:
Hi

On macOS, in a default (non-flipped) coordinate system, the origin is in the lower-left corner of the view and positive y-values extend upward. `wry` converts positions from its own coordinate system to the one defined on macOS. However, it   assumes that all coordinate systems on macOS are non-flipped, which is not always the case. This PR extends `window_position` with an [isFlipped](https://developer.apple.com/documentation/appkit/nsview/isflipped) check to also support flipped coordinate systems.

As a consequence, it fixes a positioning issue when adding a WebView to a `winit v0.30+`window as a child.

### DIFF
--- a/.changes/macos-handle-flipped-coordinates.md
+++ b/.changes/macos-handle-flipped-coordinates.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
+macOS: Handle flipped coordinates when adding a WebView as a child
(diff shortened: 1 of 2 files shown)


lucasfernog (approved):
nice catch!

PR #1547, opened 2025-04-17
fix: set WebKitGTK WebsiteDataManager's base_cache_directory

ymelois:
Set `WebsiteDataManagerBuilder::base_cache_directory` with the same path as `base_data_directory`.

This change allows the cache directory to be changed instead of using the default one [from WebKitGTK](https://webkitgtk.org/reference/webkit2gtk/stable/property.WebsiteDataManager.base-cache-directory.html).

### DIFF
--- a/.changes/webcontext-cache-directory.md
+++ b/.changes/webcontext-cache-directory.md
@@ -1,8 +1,7 @@
 ---
-"wry": minor
+"wry": patch
 ---
 
-Breaking change: `WebContext::new` now takes a `cache_directory` argument besides `data_directory`.
+Set `WebsiteDataManagerBuilder::base_cache_directory` with the same path as `base_data_directory`.
 
-This change allows users to specify a custom cache directory for the web context,
-instead of using the default one [from WebKitGTK](https://webkitgtk.org/reference/webkit2gtk/stable/property.WebsiteDataManager.base-cache-directory.html).
+This change allows the cache directory to be changed instead of using the default one [from WebKitGTK](https://webkitgtk.org/reference/webkit2gtk/stable/property.WebsiteDataManager.base-cache-directory.html).
--- a/.changes/webcontext-cache-directory.md
+++ b/.changes/webcontext-cache-directory.md
@@ -5,4 +5,4 @@
 Breaking change: `WebContext::new` now takes a `cache_directory` argument besides `data_directory`.
 
 This change allows users to specify a custom cache directory for the web context,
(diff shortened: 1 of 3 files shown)


FabianLars:
maybe we should just derive the cache dir from the data dir even if not 100% correct? That would match the behavior on Windows at least.

src/web_context.rs
```diff
@@ -33,12 +34,17 @@ pub struct WebContext {
 impl WebContext {
   /// Create a new [`WebContext`].
   ///
+  /// `cache_directory`:
+  /// * Whether the WebView window should have a custom cache path. This is useful in Windows
```
FabianLars: `Windows` in the data_directory comment is actually about the Windows OS so here it would be incorrect.

.changes/webcontext-cache-directory.md
```diff
@@ -0,0 +1,8 @@
+---
+"wry": minor
+---
+
+Breaking change: `WebContext::new` now takes a `cache_directory` argument besides `data_directory`.
+
+This change allows users to specify a custom cache directory for the web context,
+instead of using the default one [from WebKit2](https://webkitgtk.org/reference/webkit2gtk/stable/property.WebsiteDataManager.base-cache-directory.html).
```
FabianLars:
```suggestion
instead of using the default one [from WebKitGTK](https://webkitgtk.org/reference/webkit2gtk/stable/property.WebsiteDataManager.base-cache-directory.html).
```

src/webkitgtk/web_context.rs
```diff
@@ -40,6 +40,7 @@ impl WebContextImpl {
     let mut context_builder = WebContext::builder();
     if let Some(data_directory) = data_directory {
       let data_manager = WebsiteDataManager::builder()
+        .base_cache_directory(data_directory.to_string_lossy())
```
FabianLars, line 44: i can't remember what exactly the cache looks like on windows, do you think it makes sense to create a `cache` subdir like we do for the cookies below?

ymelois: by default webkitgtk put the cache in the same dir as the data, so I don't think it would be wrong to set it that way as well when providing a custom data path

FabianLars:
~~i hate to say it but your idea was better, i'll take care of it though in this PR. Sorry for wasting your time.~~ nevermind this is too annoying to handle on the tauri side. we'll revisit this thought later. webkitgtk doesn't cache much anyway.

FabianLars (approved):

PR #1548, opened 2025-04-18
refactor(windows): work around uri

Legend-Master:
The old work around naming and the manual uri detection is quite confusing

### DIFF
--- a/src/webview2/mod.rs
+++ b/src/webview2/mod.rs
@@ -1655,7 +1655,6 @@ unsafe fn set_theme(webview: &ICoreWebView2, theme: Theme) -> Result<()> {
 ///
 /// See https://github.com/MicrosoftEdge/WebView2Feedback/issues/73
 fn is_work_around_uri(uri: &str, http_or_https: &str, protocol: &str) -> bool {
-  //
   uri
     .strip_prefix(http_or_https)
     .and_then(|rest| rest.strip_prefix("://"))
--- a/src/webview2/mod.rs
+++ b/src/webview2/mod.rs
@@ -1649,8 +1649,13 @@ unsafe fn set_theme(webview: &ICoreWebView2, theme: Theme) -> Result<()> {
     .map_err(Into::into)
 }
 
+/// WebView2 supports non-standard protocols only on Windows 10+, so we have to use a workaround,
+/// conveting `{protocol}://localhost/abc` to `{http_or_https}://{protocol}.localhost/abc`,
+/// and this function tests if the URI starts with `{http_or_https}://{protocol}.`
+///
+/// See https://github.com/MicrosoftEdge/WebView2Feedback/issues/73
 fn is_work_around_uri(uri: &str, http_or_https: &str, protocol: &str) -> bool {
-  // Test if it starts with "{http_or_https}://{protocol}."
+  //
   uri
     .strip_prefix(http_or_https)
     .and_then(|rest| rest.strip_prefix("://"))
--- a/src/webview2/mod.rs
+++ b/src/webview2/mod.rs
@@ -1435,7 +1435,7 @@ impl InnerWebView {
     cookie.Expires(&mut expires)?;
 
     let expires = match expires {
-      -1.0 | _ if is_session.as_bool() => Some(cookie::Expiration::Session),
+      _ if expires == -1.0 || is_session.as_bool() => Some(cookie::Expiration::Session),
       datetime => cookie::time::OffsetDateTime::from_unix_timestamp(datetime as _)
         .ok()
         .map(cookie::Expiration::DateTime),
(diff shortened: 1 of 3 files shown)


src/webview2/mod.rs
```diff
@@ -1652,20 +1649,30 @@ unsafe fn set_theme(webview: &ICoreWebView2, theme: Theme) -> Result<()> {
...
-  let protocol_len = protocol.len();
-
-  // starts with `http` or `https``
-  &uri[..scheme_len] == scheme
-  // followed by `://`
-  && &uri[scheme_len..scheme_len + 3] == "://"
-  // followed by custom protocol name
-  && scheme_len + 3 + protocol_len < uri_len && &uri[scheme_len + 3.. scheme_len + 3 + protocol_len] == protocol
-  // and a dot
-  && scheme_len + 3 + protocol_len < uri_len && uri.as_bytes()[scheme_len + 3 + protocol_len] == b'.'
```
amrbashir, line 1668: the reason why I did this, was for performance, as it is called inside each custom protocol invocation, and I wanted to avoid allocation here

I am not opposed to improving this, just avoid allocation

other changes sound fine

amrbashir (changes requested):

src/webview2/mod.rs
Legend-Master: I see,  I'll change the implementation, did a small benchmark and the result is promising

```
    Finished `release` profile [optimized] target(s) in 0.03s
     Running `target\release\hello.exe`
[src\main.rs:30:5] "is_custom_protocol_uri" = "is_custom_protocol_uri"
[src\main.rs:30:5] now.elapsed() = 10.1609ms
[src\main.rs:31:5] is_custom_count = 499843
[src\main.rs:40:5] "is_work_around_uri" = "is_work_around_uri"
[src\main.rs:40:5] now.elapsed() = 8.9254ms
[src\main.rs:41:5] is_custom_count = 499843
```

Code:

```rust
use rand::{Rng, distr::Alphanumeric};
use std::time::Instant;

fn main() {
    let mut random_strings = vec![];
    for _ in 1..1_000_000 {
        let random: String = rand::rng()
            .sample_iter(&Alphanumeric)
            .take(10)
            .map(char::from)
            .collect();
        let string = if rand::random_bool(0.5) {
            format!("http://wry.{random}")
        } else {
            format!("http://{random}")
        };
        random_strings.push(string);
    }

    let mut is_custom_count = 0;
    let now = Instant::now();
    for string in &random_strings {
        if is_custom_protocol_uri(string, "http", "wry") {
            is_custom_count += 1;
        }
    }
    dbg!("is_custom_protocol_uri", now.elapsed());
    dbg!(is_custom_count);

    let mut is_custom_count = 0;
    let now = Instant::now();
    for string in &random_strings {
        if is_work_around_uri(string, "http", "wry") {
            is_custom_count += 1;
        }
    }
    dbg!("is_work_around_uri", now.elapsed());
    dbg!(is_custom_count);

    checks_if_custom_protocol_uri()
}

fn checks_if_custom_protocol_uri() {
    let scheme = "http";
    let uri = "http://wry.localhost/path/to/page";
    assert!(is_work_around_uri(uri, scheme, "wry"));
    assert!(!is_work_around_uri(uri, scheme, "asset"));
}

fn is_custom_protocol_uri(uri: &str, scheme: &'static str, protocol: &str) -> bool {
    let uri_len = uri.len();
    let scheme_len = scheme.len();
    let protocol_len = protocol.len();

    // starts with `http` or `https``
    &uri[..scheme_len] == scheme
    // followed by `://`
    && &uri[scheme_len..scheme_len + 3] == "://"
    // followed by custom protocol name
    && scheme_len + 3 + protocol_len < uri_len && &uri[scheme_len + 3.. scheme_len + 3 + protocol_len] == protocol
    // and a dot
    && uri.as_bytes()[scheme_len + 3 + protocol_len] == b'.'
}

fn is_work_around_uri(uri: &str, http_or_https: &str, protocol: &str) -> bool {
    // uri.starts_with(&work_around_uri_prefix(http_or_https, protocol))
    uri.strip_prefix(http_or_https)
        .and_then(|rest| rest.strip_prefix("://"))
        .and_then(|rest| rest.strip_prefix(protocol))
        .and_then(|rest| rest.strip_prefix("."))
        .is_some()
}
```

pewsheen: I think let `is_custom_protocol_uri` all compare with bytes should be faster because it doesn't need to decode UTF-8, and we are able to retrieve `scheme` and `protocol`'s byte length.

Just a nit that the implementation of `is_work_around_uri` looks much more readable to me and I like it, but the naming is confusing me like why URI need a workaround and what's the URI for?

Legend-Master: > I think let is_custom_protocol_uri all compare with bytes should be faster because it doesn't need to decode UTF-8, and we are able to retrieve scheme and protocol's byte length.

`protocol` and `scheme`(`http_or_https`) are stored in `String` already, and `uri` is converted from a `PWSTR`, so I don't think we can optimize much here, or do you mean we compare them through `as_btyes`? That didn't make a difference from testing

> but the naming is confusing me like why URI need a workaround and what's the URI for?

Yeah, I could add some docs

amrbashir (approved):

PR #1554, opened 2025-04-24
unwrap handled.

Agile86:
In offline mode, we encountered a crash due to an unwrap() call. Using unwrap() is bad practice, and the code was falling into it. This PR fixes the issue.

### DIFF
--- a/src/wkwebview/class/wry_web_view_delegate.rs
+++ b/src/wkwebview/class/wry_web_view_delegate.rs
@@ -58,7 +58,9 @@ define_class!(
             CStr::from_ptr(url_utf8).to_str(),
             CStr::from_ptr(js_utf8).to_str(),
           ) {
-            ipc_handler(Request::builder().uri(url).body(js.to_string()).unwrap());
+            if let Ok(r) = Request::builder().uri(url).body(js.to_string()) {
+              ipc_handler(r);
+            }
             return;
           }
         }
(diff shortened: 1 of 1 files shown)


src/wkwebview/class/wry_web_view_delegate.rs
```diff
@@ -58,7 +58,9 @@ define_class!(
             CStr::from_ptr(url_utf8).to_str(),
             CStr::from_ptr(js_utf8).to_str(),
           ) {
-            ipc_handler(Request::builder().uri(url).body(js.to_string()).unwrap());
+            if let Ok(r) = Request::builder().uri(url).body(js.to_string()) {
+              ipc_handler(r);
+            }
```
amrbashir: we should probably handle the error case and return a 500 response

amrbashir (changes requested):

src/wkwebview/class/wry_web_view_delegate.rs
Agile86: this func `did_receive` return void and there is a lot of place which ignoring return errors (if let Ok()), so I just apply generic pattern here.

Agile86:
Guys, do you have plans to accept this PR, or are you forcing me to rewrite the whole function to handle all errors?

FabianLars:
since it returns early there, maybe just another trace::warn call so devs actually know that it failed? otherwise it looks like a success.

Agile86:
> since it returns early there, maybe just another trace::warn call so devs actually know that it failed? otherwise it looks like a success.

done.

FabianLars (approved):

FabianLars:
Ah, and thanks for the PR :)

PR #1555, opened 2025-05-02
feat(ios): add support for setting the background color

FabianLars:
This is _required_ to get rid of white/black flashes on app launch. ref https://github.com/tauri-apps/tauri/issues/1564

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -370,6 +370,24 @@ impl InnerWebView {
         let frame = ns_view.frame();
         let webview: Retained<WryWebView> =
           objc2::msg_send![super(webview), initWithFrame: frame, configuration: &**config];
+        if let Some((red, green, blue, alpha)) = attributes.background_color {
+          // This is required first since the webview color is applied too late.
+          webview.setOpaque(false);
+
+          let color = objc2_ui_kit::UIColor::colorWithRed_green_blue_alpha(
+            (red / 255).into(),
+            (green / 255).into(),
+            (blue / 255).into(),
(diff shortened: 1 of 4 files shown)


src/wkwebview/mod.rs
```diff
@@ -370,6 +370,24 @@ impl InnerWebView {
         let frame = ns_view.frame();
         let webview: Retained<WryWebView> =
           objc2::msg_send![super(webview), initWithFrame: frame, configuration: &**config];
+        if let Some((red, green, blue, alpha)) = attributes.background_color {
+          // This is required first since the webview color is applied too late.
+          webview.setOpaque(false);
+
+          let color = objc2_ui_kit::UIColor::colorWithRed_green_blue_alpha(
+            (red / 255).into(),
```
pewsheen: I think we should convert the value into a floating point like `(red as f64 / 255.0)` otherwise the value < 255 will present as 0.

pewsheen (changes requested):

src/wkwebview/mod.rs
FabianLars: idk how i missed that, did i really only test colors that are close to 0 or 255 in all 3 colors? I mean, i'm colorblind but that's a bit wild 😂

pewsheen (dismissed):

src/wkwebview/mod.rs
pewsheen: Do we need a changelog?

FabianLars: yes

pewsheen (approved):

PR #1556, opened 2025-05-10
refactor: remove internal `and_then` chaining

Legend-Master:
I noticed this while debugging the compile time in tauri with `cargo-llvm-lines`

image

The `and_then` chaining was introduced in #1384 to record the duplcated custom protocol error

If we would return it in `build` anyways, let's just check it there instead

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -64,8 +64,6 @@ pub enum Error {
   CustomProtocolTaskInvalid,
   #[error("Failed to register URL scheme: {0}, could be due to invalid URL scheme or the scheme is already registered.")]
   UrlSchemeRegisterError(String),
-  #[error("Duplicate custom protocol registered on Linux: {0}")]
-  DuplicateCustomProtocol(String),
   #[error("Duplicate custom protocol registered on the same web context on Linux: {0}")]
   ContextDuplicateCustomProtocol(String),
   #[error(transparent)]
--- a/src/error.rs
+++ b/src/error.rs
@@ -64,7 +64,7 @@ pub enum Error {
   CustomProtocolTaskInvalid,
   #[error("Failed to register URL scheme: {0}, could be due to invalid URL scheme or the scheme is already registered.")]
   UrlSchemeRegisterError(String),
-  #[error("Duplicate custom protocol '{0}' registered on the builder")]
+  #[error("Duplicate custom protocol '{0}' registered on the WebViewBuilder")]
   DuplicateCustomProtocol(String),
   #[error("Duplicate custom protocol '{0}' registered on the same web context on Linux")]
   ContextDuplicateCustomProtocol(String),
--- a/src/error.rs
+++ b/src/error.rs
@@ -64,7 +64,9 @@ pub enum Error {
   CustomProtocolTaskInvalid,
   #[error("Failed to register URL scheme: {0}, could be due to invalid URL scheme or the scheme is already registered.")]
   UrlSchemeRegisterError(String),
-  #[error("Duplicate custom protocol registered on the same web context on Linux: {0}")]
+  #[error("Duplicate custom protocol '{0}' registered on the builder")]
+  DuplicateCustomProtocol(String),
+  #[error("Duplicate custom protocol '{0}' registered on the same web context on Linux")]
   ContextDuplicateCustomProtocol(String),
   #[error(transparent)]
   #[cfg(any(target_os = "macos", target_os = "ios"))]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1265,8 +1265,8 @@ impl<'a> WebViewBuilder<'a> {
       target_os = "netbsd",
       target_os = "openbsd",
     ))]
-    for protocol_name in self.attrs.custom_protocols.keys() {
-      if let Some(context) = &mut self.attrs.context {
+    if let Some(context) = &self.attrs.context {
+      for protocol_name in self.attrs.custom_protocols.keys() {
         if context.is_custom_protocol_registered(protocol_name) {
           return Err(crate::Error::ContextDuplicateCustomProtocol(
             protocol_name.to_string(),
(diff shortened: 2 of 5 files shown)


FabianLars:
Sometimes(?) simple is better :D

FabianLars:
I'm only on my phone rn so I may be imagining things but if we use hash map.insert with the protocol name as key we will silently over duplicates on the with_protocol method, no?

Legend-Master:
> but if we use hash map.insert with the protocol name as key we will silently over duplicates on the with_protocol method, no?

The newer handler will override the existing one in the builder, but I feel like that's what makes sense to me, since it's a builder that's not being built yet, why not to allow that?

FabianLars:
I'm okay with that, it's just different than it was with and_then 🤷 also idk if your argument really applies to a builder method you'll often call multiple times without wanting to overwrite the protocol. but yeah, either way works for me, just wanted to make sure it wasn't an accident.

FabianLars (dismissed):

amrbashir (changes requested):
Don't merge just yet, I will review it as soon as I can

src/error.rs
```diff
@@ -64,8 +64,6 @@ pub enum Error {
   CustomProtocolTaskInvalid,
   #[error("Failed to register URL scheme: {0}, could be due to invalid URL scheme or the scheme is already registered.")]
   UrlSchemeRegisterError(String),
-  #[error("Duplicate custom protocol registered on Linux: {0}")]
-  DuplicateCustomProtocol(String),
```
amrbashir: This error needs to be brought back, I'd rather emit an error than silently ignore one of the calls to `with_custom_protocol`

src/lib.rs
```diff
@@ -958,36 +928,18 @@ impl<'a> WebViewBuilder<'a> {
...
-        target_os = "netbsd",
-        target_os = "openbsd",
-      ))]
-      if let Some(context) = &mut b.attrs.context {
-        context.register_custom_protocol(name.clone())?;
-      }
-
-      if b.attrs.custom_protocols.iter().any(|(n, _)| n == &name) {
-        return Err(Error::DuplicateCustomProtocol(name));
-      }
```
amrbashir: we need to keep this check here, otherwise `.insert` will just replace the stored value and there is no way to know afterwards that we may had duplicate entries.

The way to avoid having to do `self.and_then` like before, is that we can store `err: Result<(), Error>` on the `WebviewBuilder` and call `err.and_then` here and do these checks, then propagate it in `.build()`


```rs

struct WebViewBuilder {
  err: Result<(), Error>
}

impl WebViewBuilder {
  fn with_protocol(self) -> Self {
	let mut err = Ok(());
	// 1. do checks here and update `err` variable if any
	// 2. insert into custom protocol, we can ignore duplicate since we already did the check 
    self.custom_protocols.insert();
    // 3. update internal err
    self.err.and_then(|_| err);
	self
  }
  
  fn build(self) {
    self.err?;
  }
}

```

src/web_context.rs
```diff
@@ -58,18 +58,9 @@ impl WebContext {
...
   }
 
-  #[allow(dead_code)]
-  pub(crate) fn register_custom_protocol(&mut self, name: String) -> Result<(), crate::Error> {
-    if self.custom_protocols.contains(&name) {
-      return Err(crate::Error::ContextDuplicateCustomProtocol(name));
-    }
-
-    Ok(())
-  }
```
amrbashir: this also needs to be reverted

src/webkitgtk/web_context.rs
```diff
@@ -143,6 +143,8 @@ impl WebContextExt for super::WebContext {
   where
     F: Fn(crate::WebViewId, Request<Vec<u8>>, RequestAsyncResponder) + 'static,
   {
+    self.custom_protocols.insert(name.to_owned());
```
amrbashir: I think this was a bug unnoticed, and should be instead in `WebContext::register_custom_protocol` for all platforms not just Linux

amrbashir (changes requested):

src/webkitgtk/web_context.rs
Legend-Master: Do other platforms share custom protocol handlers between webviews with the same `WebContext`? If not then why should we disallow this on those platforms?

src/error.rs
Legend-Master: I'm fine with that, removed this mainly because the error says 'Duplicate custom protocol registered on Linux' which I assumed to be the same as the one below, since there're technically no restrictions to register it multiple times in a builder

About if we should allow duplicates in the same builder though, I don't mind it either way

src/webkitgtk/web_context.rs
amrbashir: no, other platforms needs to re-register for each webview, but it feels weird being here, it should be in `WebContext::register_custom_protocol` behind an cfg for linux

src/error.rs
amrbashir: > I'm fine with that, removed this mainly because the error says 'Duplicate custom protocol registered on Linux' which I assumed to be the same as the one below, since there're technically no restrictions to register it multiple times in a builder

yeah that wording was just incorrect

src/lib.rs
```diff
@@ -958,36 +933,22 @@ impl<'a> WebViewBuilder<'a> {
...
-        Box::new(move |id, request, responder| {
-          let http_response = handler(id, request);
-          responder.respond(http_response);
-        }),
-      );
-
-      Ok(b)
-    })
+    if self.attrs.custom_protocols.contains_key(&name) {
+      self.error = Err(crate::Error::DuplicateCustomProtocol(name));
```
amrbashir: in the future there maybe other methods added that would require returning an err, so I think instead of replacing `self.error` we should chain `self.error.and_then`

amrbashir (approved):

Legend-Master:
The reason I put the linux web context check inside the final build is because I think it makes more sense to check it just before the build since it doesn't matter before that point

Anyways, that's fine either ways

amrbashir:
> The reason I put the linux web context check inside the final build is because I think it makes more sense to check it just before the build since it doesn't matter before that point

I moved it there so we can find the logic more easily, it doesn't make sense to delay the check until `.build` call.

PR #1561, opened 2025-06-03
fix(linux): deadlock in WebViewUriLoader

dgerhardt:
This fixes a deadlock, which could occur on Linux, when a WebView was destroyed before handling `LoadEvent::Finished`. As a consequence, no further requests could be performed.

With this fix, https://github.com/tauri-apps/tauri/issues/12589 is no longer consistently reproducible. But there still seem to be rare situations, in which a deadlock could occur, so I don't think the issue can be closed.

Note: This PR could be slightly improved by disconnecting the `connect_destroy` signal handler when `LoadEvent::Finished` is handled. But I'm not experienced in Rust development at all and have no idea how to move the `SignalHandlerId` into the closure to call `webview.disconnect(id)`.

## Additional context / long-term solution
The deadlock is caused by a workaround for a supposed bug in WebKitGTK. But the workaround introduces new bugs and negatively impacts request performance by queuing them up. Ideally, the underlying issue would be further investigated, so the workaround could be removed in the future. Currently, there doesn't seem to be an issue to track this problem and the bug has not been reported to the WebKitGTK project. As a start, it would be helpful to allow disabling the workaround. This way, it would be possible to verify if it is still needed for the latest WebKitGTK version.

### DIFF
--- a/.changes/fix-webkitgtk-webviewuriloader-deadlock.md
+++ b/.changes/fix-webkitgtk-webviewuriloader-deadlock.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
+On Linux, fix a deadlock, which could occur when destroying a WebView before loading has finished.
--- a/src/webkitgtk/web_context.rs
+++ b/src/webkitgtk/web_context.rs
@@ -5,7 +5,10 @@
 //! Unix platform extensions for [`WebContext`](super::WebContext).
 
 use crate::{Error, RequestAsyncResponder};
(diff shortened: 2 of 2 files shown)


dgerhardt:
@chippers @wusyong Could you have a look at this PR? It is related to a workaround introduced with https://github.com/tauri-apps/wry/pull/359, which you submitted/reviewed.

lucasfernog (approved):
nice catch, thanks!

Issue #1563, opened 2025-06-15
Android: add Intent handling


ambiso:
Currently tauri and frameworks that rely on wry/tao such as dioxus don't seem to support passing data to an android application through e.g. the SEND intent on android. For example I would like to share an URL from my browser with my app, but there's no way to receive the Intent from the native Rust code.

## Proposed solution
Receive intents in the WryActivity and call a function in the ndk_glue (in tao). A separate crate, that client code can depend on, should expose an API that allows receiving the intents from Rust.

## Alternatives

The obvious drawback of implementing this in wry/tao is feature bloat: receiving intents should maybe not really be a concern of wry.

### class extension

This feature possibly could e.g. also be implemented as part of the `WRY_WRYACTIVITY_CLASS_EXTENSION` envvar; however, since there's also a native component to this, the user would have to setup the build environment for another native library to compile for the target, so that the native code can be notified of intents.

In general it would be nice if there was an easy way to setup your app for receiving intents.

### separate activity

the user could create a separate activity altogether, that stores the intent data somewhere, then launches the main activity, which then retrieves the intent data from the storage location.
this would be possibly less flexible, since the original intent jobject would likely be lost.

## Additional context

<details>
<summary>
Demo video shows sharing an URL with an android application, and the app receiving and displaying the URL.</summary>

https://github.com/user-attachments/assets/76ce28d0-0df0-4c81-8d87-b3822597e350

</details>

PR #1565, opened 2025-06-15
android: add intent handling

ambiso:
Demo implementation of #1563

Overrides the `onNewIntent` function which is invoked when a running application is launched again with a new intent.
The activity's launch mode must be singleTop for `onNewIntent` to be called, but preferrably `singleInstance` or `singleTask`, since otherwise the ndk_context may be initialized twice.


FabianLars:
Added in another PR

ambiso:
Could you reference which PR implemented this?

FabianLars:
https://github.com/tauri-apps/wry/pull/1634

PR #1566, opened 2025-06-17
Update html5ever and kuchikiki

sftse:
Update `html5ever` and `kuchikiki` to the most recent version both are compatible with each other.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -470,15 +470,15 @@ dependencies = [
 
 [[package]]
 name = "cssparser"
-version = "0.27.2"
+version = "0.29.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


sftse:
@FabianLars Please retry, missed a spot.

FabianLars:
just fyi, i'll wait with this until we got feedback on the tauri pr - ~~until then, a changefile would be cool so we can release it without waiting for other changes~~ nvm, a versioning pr is already open so this wouldn't work

Legend-Master (approved):
I'm good with the change personally 🙃, thanks for the effort

FabianLars (approved):

sftse:
@FabianLars You might want to check whether the CI fmt check is working as intended. The formatting change here was on code I didn't otherwise touch, so the code on dev was not formatted.

PR #1568, opened 2025-06-21
Make `window.close` no-op

anatawa12:
Make `window.close` no-op for several reasons:
- its behavior is not good. currently it would leave broken window without notifying library user (tauri and application)
- It should possible to ignore `window.close`
- It's inconsistent with apple platform (and possibly android?)

Fixing tauri-apps/tauri#13331

### DIFF
--- a/.changes/window-close-no-op.md
+++ b/.changes/window-close-no-op.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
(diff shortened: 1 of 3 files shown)


lucasfernog:
i think we should add the window.close handler instead of making it no-op.

anatawa12:
yes it's best to ad dhandler but i'm proposing this as quick fix to the bug breaks the window

PR #1569, opened 2025-06-22
feat: add `WebView::set_cookie` and `WebView::delete_cookie`

### LINKED ISSUE #1511: [feature] Implement WebView::set_cookie
Some apps may require access to other website cookies (ex. user logins to a website and we use the authentification cookie to provide custom functionaliy)

Currently we can fetch cookies via `WebView::cookies` and `WebView::cookies_for_url` . Some apps may also require setting cookies in order to set an authentification token for a certain website. My use case for example is to make an app that handles multiple accounts and allows switching between them. The accounts are provided by a remote website and therefore I need access to a set cookie function in order to achieve this goal. Electron for example has a set cookie method and I think it would be nice to have one in tauri too.

Citing the following PR's:
https://github.com/tauri-apps/wry/pull/1394
https://github.com/tauri-apps/tauri/pull/12665

WSH032:
ref #1394
close #1511

### DIFF
--- a/examples/cookies.rs
+++ b/examples/cookies.rs
@@ -1,0 +1,63 @@
+// Copyright 2020-2023 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+use tao::{
+  event::{Event, WindowEvent},
+  event_loop::{ControlFlow, EventLoop},
+  window::WindowBuilder,
+};
+use wry::WebViewBuilder;
+
+fn main() -> wry::Result<()> {
+  let event_loop = EventLoop::new();
+  let window = WindowBuilder::new().build(&event_loop).unwrap();
+
+  let builder = WebViewBuilder::new().with_url("https://www.httpbin.org/cookies/set?foo=bar");
+
+  #[cfg(any(
+    target_os = "windows",
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "android"
+  ))]
+  let webview = builder.build(&window)?;
+  #[cfg(not(any(
+    target_os = "windows",
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "android"
+  )))]
+  let webview = {
+    use tao::platform::unix::WindowExtUnix;
+    use wry::WebViewBuilderExtUnix;
+    let vbox = window.default_vbox().unwrap();
+    builder.build_gtk(vbox)?
+  };
+
+  webview.set_cookie(
+    cookie::Cookie::build(("foo1", "bar1"))
+      .domain("www.httpbin.org")
+      .path("/")
+      .secure(true)
+      .http_only(true)
+      .max_age(cookie::time::Duration::seconds(10))
+      .inner(),
+  )?;
+
+  for cookie in webview.cookies()? {
+    println!("{}", cookie);
+  }
+
+  event_loop.run(move |event, _, control_flow| {
+    *control_flow = ControlFlow::Wait;
+
+    if let Event::WindowEvent {
+      event: WindowEvent::CloseRequested,
+      ..
+    } = event
+    {
+      *control_flow = ControlFlow::Exit;
+    }
+  });
+}
--- a/examples/cookies.rs
+++ b/examples/cookies.rs
@@ -45,8 +45,18 @@ fn main() -> wry::Result<()> {
       .inner(),
   )?;
 
+  let cookie_deleted = cookie::Cookie::build(("will_be_deleted", "will_be_deleted"));
+
+  webview.set_cookie(cookie_deleted.inner())?;
+  println!("Setting Cookies:");
   for cookie in webview.cookies()? {
-    println!("{}", cookie);
+    println!("\t{cookie}");
+  }
+
+  println!("After Deleting:");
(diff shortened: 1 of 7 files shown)


src/wkwebview/mod.rs
```diff
@@ -979,6 +981,70 @@ r#"Object.defineProperty(window, 'ipc', {
...
+    if let Some(http_only) = cookie.http_only() {
+      let http_only = if http_only {
+        nstring_true
+      } else {
+        nstring_false
+      };
+      // ref:
+      // - <https://stackoverflow.com/a/41697557>
+      // - <https://developer.apple.com/forums/thread/701770?answerId=706717022#706717022>
+      properties.insert(ns_string!("HttpOnly"), http_only);
```
WSH032, line 1032: I'm not sure if this is a private API, because Apple support staff said that the absence of this API is a bug?  
But I prefer to use it, since being able to set `http_only` cookies is a very important webview feature. Otherwise, why not just use JavaScript to set cookies?

WSH032 (commented):
I have already tested this on Windows and Linux. I need others to help test on macOS and iOS (since I don't have a Mac 😫).

I haven't implemented the Android part because I don't know Kotlin. I think this should be left for someone else to implement.

src/wkwebview/mod.rs
pewsheen: From my perspective, it's ok to use the property. The DTS didn't ask people not to use it, but instead provided a solution. I think it should be a positive sign.

AlloryDante:
Thank you so much for making this. Once this gets approved I can finally build my app. I need this as my app needs to modify auth cookies for certain websites and this is the only way I can implement it. Thank you @WSH032 and @pewsheen

src/wkwebview/mod.rs
```diff
@@ -979,6 +981,70 @@ r#"Object.defineProperty(window, 'ipc', {
...
+
+    let wkwebview_cookie = NSHTTPCookie::cookieWithProperties(&properties)
+      .expect("failed to create wkwebview cookie, report this as bug to `wry`");
+
+    // check if `HttpOnly` works as expected
+    debug_assert_eq!(
+      wkwebview_cookie.isHTTPOnly(),
+      cookie.http_only().unwrap_or(false),
+      "HTTPOnly mismatch, report this as bug to `wry`"
+    );
```
pewsheen: Can we remove this part since we should not let it happen.

src/wkwebview/mod.rs
```diff
@@ -979,6 +981,70 @@ r#"Object.defineProperty(window, 'ipc', {
...
+      let http_only = if http_only {
+        nstring_true
+      } else {
+        nstring_false
+      };
+      // ref:
+      // - <https://stackoverflow.com/a/41697557>
+      // - <https://developer.apple.com/forums/thread/701770?answerId=706717022#706717022>
+      properties.insert(ns_string!("HttpOnly"), http_only);
+    }
```
pewsheen, line 1033: Can we insert the property only when `http_only` is true? Since when it's false it does nothing to the cookie.

src/wkwebview/mod.rs
```diff
@@ -979,6 +981,70 @@ r#"Object.defineProperty(window, 'ipc', {
...
+
+    let properties: Retained<NSMutableDictionary<NSHTTPCookiePropertyKey, AnyObject>> =
+      NSMutableDictionary::from_slices(
+        &[
+          NSHTTPCookieName,
+          NSHTTPCookieValue,
+          NSHTTPCookiePath,
+          NSHTTPCookieDomain,
+        ],
+        &[&*name, &*value, &*path, &*domain],
```
pewsheen: I think it should be okay to use `&name`, ...  only?

examples/cookies.rs
```diff
@@ -0,0 +1,63 @@
...
+      .secure(true)
+      .http_only(true)
+      .max_age(cookie::time::Duration::seconds(10))
+      .inner(),
+  )?;
+
+  for cookie in webview.cookies()? {
+    println!("{}", cookie);
+  }
+
```
pewsheen, line 61: We can also add `delete_cookie` here and print again, ensure the deletion is working.

pewsheen (commented):
Thank you! I currently could only reveiw the macOS part. Overall it looks good to me, but we'll need someone to review webview2 and gtk parts.

Also, I think you might missing the same site policy in macOS? If you're not able to implement that part then I can help with that.

pewsheen (approved):

WSH032:
Oops, I just re-signed my commit, so now I need to force push.

pewsheen:
> Oops, I just re-signed my commit, so now I need to force push.

Oh okay. If it's hard to rebase on your side, you can overwrite my commit and force push.

src/wkwebview/mod.rs
WSH032: Personally, I prefer to follow the `Option<bool>` semantics: if you explicitly specify it, I will set it for you.  
I don't have a strong opinion on this, but I've used the same implementation on both Linux and Windows. If we want to change it, it's best to update both for consistency?

pewsheen: It's okay.

WSH032:
Thanks!

AlloryDante:
OMG, thank you so much for the contribution!

PR #1588, opened 2025-07-15
Feat: add `WebViewBuilder::with_limit_navigations_to_app_bound_domains`

### LINKED ISSUE #1587: Allow Setting App-Bound Domains for iOS
On iOS Service Workers are disabled in WKWebView by default.

This can be fixed by setting App-Bound Domains for iOS in `Info.plist` in addition to running
`webViewConfiguration.limitsNavigationsToAppBoundDomains = YES;` on the WKWebViewConfiguration object. Then all WKWebViews used in the app are able to use service workers [source](https://stackoverflow.com/questions/49673399/service-workers-unavailable-in-wkwebview-in-ios-11-3/64155509#64155509).

 It additionally further locks down the WKWebView to only allow requests to resources from domains which are explicitly whitelisted. That said, I mainly want the feature so I can use a Service Worker in my Tauri App.

I'd like a feature flag like as `limit-navigations-to-app-bound-domains` (as App-Bound Domains are only available on iOS 14+) which then will run `webViewConfiguration.limitsNavigationsToAppBoundDomains = YES;` for the iOS target. For all other targets setting that feature flag would be a no-op. 

Developers who choose to set this flag should also manually update their Info.plist files to manually set the specific App-Bound Domains for their App.
```
<plist>
<dict>
	<key>WKAppBoundDomains</key>
	<array>
		<string>localhost</string>
		<string>tauri.localhost</string>
	</array>
</dict>
</plist>
```

Happy to make a PR as well. Figured I should start with an issue first.

I considered using a Tauri plugin; unfortunately the reference to the WKWebviewConfiguration object is not made available to a Tauri plugin, at least as far as I could tell.

**Additional context**
App-Bound Domains Blog Post: https://webkit.org/blog/10882/app-bound-domains/

zphrs:
closes #1587.

In the above issue @FabianLars wrote:

> I don't think this needs a feature flag. We can check the iOS version in the function and make it do nothing on older versions.

I did this for my implementation: 
```rs
use crate::wkwebview::util::operating_system_version;

...

if pl_attrs.limit_navigations_to_app_bound_domains && operating_system_version().0 >= 14 {
  config.setLimitsNavigationsToAppBoundDomains(true);
}
```

### DIFF
--- a/.changes/app-bound-domains.md
+++ b/.changes/app-bound-domains.md
@@ -1,0 +1,6 @@
+---
+"wry": "minor"
+---
+
+Add `WebViewBuilder::with_limit_navigations_to_app_bound_domains` only on iOS.
+Function is a no-op if iOS version is less than iOS 14.
--- a/.changes/app-bound-domains.md
+++ b/.changes/app-bound-domains.md
@@ -1,5 +1,5 @@
 ---
-"wry": "minor"
+"wry": "patch"
 ---
 
 Add `WebViewBuilder::with_limit_navigations_to_app_bound_domains` only on iOS.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1360,6 +1360,8 @@ pub(crate) struct PlatformSpecificWebViewAttributes {
   allow_link_preview: bool,
   #[cfg(target_os = "ios")]
   input_accessory_view_builder: Option<Box<InputAccessoryViewBuilder>>,
+  #[cfg(target_os = "ios")]
+  limit_navigations_to_app_bound_domains: bool,
 }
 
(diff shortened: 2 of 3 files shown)


pewsheen (dismissed):
Thank you.

Btw, you'll need you upload your signing key to your GitHub account, otherwise the commit is unverified.

zphrs:
Done!

.changes/app-bound-domains.md
```diff
@@ -0,0 +1,6 @@
+---
+"wry": "minor"
```
pewsheen:
```suggestion
"wry": "patch"
```

pewsheen (approved):

PR #1594, opened 2025-08-06
fix(linux): set suggested filename for download started handler

FabianLars:
Opening as a draft mainly cause i didn't check how macos behaves yet. the intention here (apart from making the download handler actually work on linux) was to make it match the behavior on windows which shows the standard behavior you know from browsers.

it's ready for review, i just don't want it to be merged yet.

### DIFF
--- a/src/webkitgtk/web_context.rs
+++ b/src/webkitgtk/web_context.rs
@@ -329,7 +329,7 @@ impl WebContextExt for super::WebContext {
             let suggested_filename = if uri.starts_with("data:") {
               // for data: downloads webkitgtk will suggest to use the data as the filename
               // for example `"data:attachment/text," + encodeURI("some text")` will result in `text,some text`
-              "unknown"
+              "Unknown"
             } else {
               suggested_filename
             };
--- a/src/webkitgtk/web_context.rs
+++ b/src/webkitgtk/web_context.rs
@@ -326,25 +326,32 @@ impl WebContextExt for super::WebContext {
             let mut download_location =
               dirs::download_dir().unwrap_or_else(|| current_dir().unwrap_or_default());
 
-            let suggested_filename = if uri.starts_with("data:") {
-              // for data: downloads webkitgtk will suggest to use the data as the filename
-              // for example `"data:attachment/text," + encodeURI("some text")` will result in `text,some text`
-              "Unknown"
-            } else {
-              suggested_filename
-            };
+            let (mut suggested_filename, mut ext) = suggested_filename
+              .split_once('.')
+              .map(|(base, ext)| (base, format!(".{ext}")))
+              .unwrap_or((suggested_filename, "".to_string()));
 
-            download_location.push(suggested_filename);
+            // Weird bug but webkitgtk sets the extension to `.jpe` for JPEG files in these cases:
+            // <a href="data:image/jpeg;base64,..." download="" />
+            // <a href="/some-img.jpeg" download="some-name" />
+            //
+            // Other 4 character extensions like avif,webp seem to be set correctly.
+            if ext == ".jpe" {
+              ext = ".jpeg".to_string();
+            }
+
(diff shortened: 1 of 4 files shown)


src/webkitgtk/web_context.rs
```diff
@@ -312,27 +313,51 @@ impl WebContextExt for super::WebContext {
...
+          let uri = uri.to_string();
+
+          if let Some(download_started_handler) = download_started_handler.borrow_mut().as_mut() {
+            let mut download_location =
+              dirs::download_dir().unwrap_or_else(|| current_dir().unwrap_or_default());
+
+            let suggested_filename = if uri.starts_with("data:") {
+              // for data: downloads webkitgtk will suggest to use the data as the filename
+              // for example `"data:attachment/text," + encodeURI("some text")` will result in `text,some text`
+              "unknown"
```
pewsheen: Maybe we can at least parse its MIME type and give an ext in the future

pewsheen (dismissed):

src/webkitgtk/web_context.rs
FabianLars: hmm i wouldn't mind that though it would mean yet another dependency. For this PR the idea was to match Webview2 though and it doesn't do this either (it calls the file something like `Download` in the local language instead of `unknown` but i thought unnamed data urls are enough of an edge case to not care bout that)

pewsheen: It's okay not to add the function. I mention it because the suggested filename does have an extension attached when I download an inline `data:` image on Windows.

FabianLars: Oh really? Can you share the data: uri you tested?

pewsheen: https://css-tricks.com/examples/DataURIs/

I'm sure if I tested it correctly.

I right clicked the image and chose `Save image as`, and it uses `download.jpg` by default.

FabianLars: Thanks! Will check it out when I'm back.

FabianLars: hmm okay macos can do that too (the download via right click on the image doesn't work there though)

FabianLars: ahhh webkitgtk does so too, but without setting unknown and having the ext be kinda wonky. For example for that image it suggests `"9k=.jpe"` as the file name 😂 if i set the mime type to webp it's `"0k=.webp"` though...

i'll just use the suggested extension from webkitgtk and with a fix for jpe->jpeg

p.s. Firefox on Windows uses `Untitled` with no extension for that image btw x)

FabianLars:
i forgot the results of my macos tests but will open another PR for that if there was anything to change.

src/webkitgtk/web_context.rs
```diff
@@ -312,27 +313,58 @@ impl WebContextExt for super::WebContext {
...
+              .unwrap_or((suggested_filename, "".to_string()));
+
+            // Weird bug but webkitgtk sets the extension to `.jpe` for JPEG files in these cases:
+            // <a href="data:image/jpeg;base64,..." download="" />
+            // <a href="/some-img.jpeg" download="some-name" />
+            //
+            // Other 4 character extensions like avif,webp seem to be set correctly.
+            if ext == ".jpe" {
+              ext = ".jpeg".to_string();
+            }
```
pewsheen: I think we should keep `.jpe` because it looks like `.jpe` is a valid subset of the `.jpeg`... though it's being used rarely. 

This is kind of interesting so I googled the issue. The issue has existed for years that the system may use `.jpe` first than other aliases :\

https://bugs.kde.org/show_bug.cgi?id=439519
https://bugs.archlinux.org/task/54977

FabianLars: that honestly sucks 😂 but okay, i'll remove that check

pewsheen (approved):

PR #1599, opened 2025-08-13
fix(macos/ios): prepend the Downloads dir to the suggested download path

FabianLars:
turns out macos/ios had the same issue as linux except for the data: filename. sorry bout that, i thought it was different.


FabianLars:
guess i gotta test android as well but that has to wait a bit. On macOS there's also the issue of right click -> download not working but i didn't look into that yet.

lucasfernog (approved):

PR #1602, opened 2025-08-15
fix: Allow downloads by default on linux/macos to match windows behavior

### LINKED ISSUE #4633: [bug] Can not download in front end
### Describe the bug

in the browser, we can use <a> like this:
```js
 const eleLink = document.createElement("a");
                eleLink.href = url;
                eleLink.download = `${title} - ${artist}`;
                eleLink.style.display = "none";
                document.body.appendChild(eleLink);
                eleLink.click();
                document.body.removeChild(eleLink);
```
to download file in front; but in tauri,this function can not take effect.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
› OS: Mac OS 12.4.0 X64
  › Node.js: 16.14.0
  › npm: 8.3.1
  › pnpm: 6.32.3
  › yarn: 1.22.17
  › rustup: 1.24.3
  › rustc: 1.61.0
  › cargo: 1.61.0
  › Rust toolchain: stable-aarch64-apple-darwin
```

### Stack trace

_No response_

### Additional context

_No response_

### LINKED ISSUE #13734: Saving files from vanilla js code does nothing (but is working in modern browsers).
On modern webbrowsers, I can use the following code snippet ([source](https://codeberg.org/MaerchenfeeimGarten/DOORS/src/branch/main/index.html#L1067)) to save, for example, game progress or other data to a location choosen by the user:

```
function download_txt() {
        var hiddenElement = document.createElement('a')

         hiddenElement.href = 'data:attachment/text,' + encodeURI(JSON.stringify(geschafft_array)); //geschafft_array containes the game progress
         hiddenElement.target = '_blank';
         hiddenElement.download = 'savegame.doors';
         hiddenElement.click();
}
```

If I put the whole code of the game ([source code](https://codeberg.org/MaerchenfeeimGarten/DOORS/)) in the `src` folder of a vanilla-js-rust based tauri based project managed with cargo, this code does nothing and I did not get any error messages I could try to read. How could I resolve this issue? Maybe some permissions are not set or I need to replace the function with one that is compatible to the tauri framework?

FabianLars:
ref https://github.com/tauri-apps/tauri/pull/14006#issuecomment-3190490533

If we don't want to do this then we should disable it on Windows as right now it allows downloads if no handler is set.

closes https://github.com/tauri-apps/tauri/pull/14006
closes  https://github.com/tauri-apps/tauri/pull/13734
closes https://github.com/tauri-apps/tauri/pull/4633

### DIFF
--- a/.changes/allow-downloads-default.md
+++ b/.changes/allow-downloads-default.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
+Enabled all downloads by default to match WebView2 and browser behavior on all platforms. To disable this, provide a custom `download_started_handler`.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -585,6 +585,8 @@ pub struct WebViewAttributes<'a> {
   /// second is a mutable `PathBuf` reference that (possibly) represents where the file will be downloaded to. The latter
   /// parameter can be used to set the download location by assigning a new path to it, the assigned path _must_ be
   /// absolute. The closure returns a `bool` to allow or deny the download.
+  ///
+  /// [`Self::default()`] sets a handler allowing all downloads to match browser behavior.
   pub download_started_handler: Option<Box<dyn FnMut(String, &mut PathBuf) -> bool + 'static>>,
 
   /// A download completion handler to manage downloads that have finished.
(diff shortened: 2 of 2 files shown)


Legend-Master (approved):

Legend-Master:
From what people were asking, I think it makes sense to enable it by default

FabianLars:
well, people also ask us to remove the allowlist :P i thought disabling it would be more in line with how strict we are/were about well anything really 🤷

Legend-Master:
I think this is more or less in line with what we put in our default permissions (`core:default`) though

PR #1605, opened 2025-08-19
fix(linux): remove harmful request loading workaround

### LINKED ISSUE #12589: [bug] Opening a WebviewWindow after closing one breaks further Webview creation on Linux
### Describe the bug

If a `WebviewWindow` is created immediately after another one was closed, this breaks the creation of further `WebviewWindow`s on Linux. The new `WebviewWindow`s will remain blank.

The WebKit dev tools can be opened and show `about:blank` in the network console. Navigating to a website by setting `location.href` in the console works. No errors are logged.

### Reproduction

https://github.com/dgerhardt/tauri-multiwindow-webkitgtk-bug

### Expected behavior

The website set via `url` should be opened when `WebviewWindow`s are created.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Ubuntu 24.4.0 x86_64 (X64) (KDE on x11)
    ✔ webkit2gtk-4.1: 2.46.5
    ✔ rsvg2: 2.58.0
    ✔ rustc: 1.83.0 (90b35a623 2024-11-26)
    ✔ cargo: 1.83.0 (5ffbef321 2024-10-29)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (default)
    - node: 20.18.2
    - yarn: 1.22.19
    - npm: 10.8.2

[-] Packages
    - tauri 🦀: 2.2.5
    - tauri-build 🦀: 2.0.5
    - wry 🦀: 0.48.1
    - tao 🦀: 0.31.1
    - @tauri-apps/api : 2.2.0
    - @tauri-apps/cli : 2.2.7

[-] Plugins
    - tauri-plugin-opener 🦀: 2.2.5
    - @tauri-apps/plugin-opener : 2.2.5
    - tauri-plugin-log 🦀: 2.2.1
    - @tauri-apps/plugin-log : 2.2.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist/tauri-multiwindow-webkitgtk-bug/browser
    - devUrl: http://localhost:1420/
    - framework: Angular
    - bundler: Webpack
```

### Stack trace

```text

```

### Additional context

There are no issues on macOS and Windows. I'm using the JS API and have not tested if the same issue occurs when using the Rust API.

For my use case I'm opening a hidden `WebviewWindow` at app startup which is doing some initialization. When it is finished, it is closed and a new `WebviewWindow` is opened. This breaks the whole application on Linux because all new `WebviewWindow`s are blank now.

dgerhardt:
With the introduction of the multi webview unstable feature in #359, a workaround for request loading during webview initialization was added for an unknown concurrency bug in WebKitGTK. This PR now removes it because it has negative side effects and does not seem to be needed anymore:
* it introduces a deadlock situation (partially mitigated by #1561),
* it disables parallel loading when initializing multiple webviews, and
* the bug can no longer be reproduced using current WebKitGTK versions.

The workaround does not seem to have ever been intended to be a long term solution (FIXME comment). Unfortunately, there is no issue linked for the underlying issue and no reproduction was provided, which makes it hard to tell if the bug still exists with newer versions of WebKitGTK. But I have not been able to reproduce the issue on Ubuntu 24.04 using `libwebkit2gtk-4.1-0@2.48.3-0ubuntu0.24.04.1`. In the worst case scenario, this would cause a regression on an unstable feature while still fixing and improving webview loading in general on Linux.

This PR provides an alternate solution for #1562, which instead introduces a feature flag to opt out of the workaround.

Original description of the workaround:
> Prevents an unknown concurrency bug with loading multiple URIs at the same time on webkit2gtk.
>
> Using the queue prevents data race issues with loading uris for multiple [`WebView`]s in the
> same context at the same time. Occasionally, the one of the [`WebView`]s will be clobbered
> and it's content will be injected into a different [`WebView`].
>
> Example of `webview-c` clobbering `webview-b` while `webview-a` is okay:
> ```text
> webview-a triggers load-change::started
> URISchemeRequestCallback triggered with webview-a
> webview-a triggers load-change::committed
> webview-a triggers load-change::finished
> webview-b triggers load-change::started
> webview-c triggers load-change::started
> URISchemeRequestCallback triggered with webview-c
> URISchemeRequestCallback triggered with webview-c
> webview-c triggers load-change::committed
> webview-c triggers load-change::finished
> ```
>
> In that example, `webview-a` will load fine. `webview-b` will remain empty as the uri was
> never loaded. `webview-c` will contain the content of both `webview-b` and `webview-c`
> because it was triggered twice even through only started once. The content injected will not
> be sequential, and often is interjected in the middle of one of the other contents.
>
> FIXME: We think this may be an underlying concurrency bug in webkit2gtk as the usual ways of
> fixing threading issues are not working. Ideally, the locks are not needed if we can understand
> the true cause of the bug.

Closes https://github.com/tauri-apps/tauri/issues/12589

### DIFF
--- a/.changes/remove-webkitgtk-request-workaround.md
+++ b/.changes/remove-webkitgtk-request-workaround.md
@@ -1,0 +1,8 @@
+---
+wry: patch
+---
+
+On Linux, removed a workaround which forced inital requests for multiple webviews to be handled sequentially.
+The workaround was intended to fix a concurrency bug with loading multiple URIs at the same time on WebKitGTK.
+But it prevented parallelization and could cause a deadlock in certain situations.
+It is no longer needed with newer WebKitGTK versions.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1971,7 +1971,7 @@ impl WebView {
   }
 
   /// Returns the id of this webview.
-  pub fn id(&self) -> WebViewId {
+  pub fn id(&self) -> WebViewId<'_> {
     self.webview.id()
   }
 
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -669,7 +669,7 @@ impl InnerWebView {
     is_inspector_open
   }
 
-  pub fn id(&self) -> crate::WebViewId {
+  pub fn id(&self) -> crate::WebViewId<'_> {
     &self.id
   }
 
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -366,8 +366,7 @@ impl InnerWebView {
 
     // Navigation
     if let Some(url) = attributes.url {
-      web_context.queue_load_uri(w.webview.clone(), url, attributes.headers);
-      web_context.flush_queue_loader();
+      web_context.load_uri(w.webview.clone(), url, attributes.headers);
     } else if let Some(html) = attributes.html {
       w.webview.load_html(&html, None);
     }
--- a/src/webkitgtk/web_context.rs
+++ b/src/webkitgtk/web_context.rs
@@ -5,23 +5,15 @@
 //! Unix platform extensions for [`WebContext`](super::WebContext).
 
 use crate::{Error, RequestAsyncResponder};
-use gtk::{
-  glib::{self, MainContext, ObjectExt},
-  traits::WidgetExt,
-};
+use gtk::glib::{self, MainContext, ObjectExt};
 use http::{header::CONTENT_TYPE, HeaderName, HeaderValue, Request, Response as HttpResponse};
 use soup::{MessageHeaders, MessageHeadersType};
 use std::{
   borrow::Cow,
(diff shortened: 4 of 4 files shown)


lucasfernog (approved):
looks like it still works, let's see if something comes up.

@chippers please let us know if you still can reproduce the bug or if you have a reproduction example out there still :D

chippers:
> looks like it still works, let's see if something comes up.
> 
> @chippers please let us know if you still can reproduce the bug or if you have a reproduction example out there still :D

i think i have a video of it, because i remember making a video of my test case opening dozens of windows in a tiling manner on my 4k monitor to try and reproduce it (only sometimes), but it's been so long idk where it is or the test suite i made for it. if i find it ill post it but if it's no longer needed, that's good :)

PR #1595, opened 2025-08-09
feat: add openharmony support

richerfu:
This work can help wry to support wry. We already add some examples to test it. You can see details: https://github.com/richerfu/wry-example

And we also add some codes to support ohos for tao, related PR: https://github.com/tauri-apps/tao/pull/1114.

PR #1607, opened 2025-08-20
feat: support openharmony platform

richerfu:
https://github.com/tauri-apps/wry/pull/1595

### DIFF
--- a/src/ohos/mod.rs
+++ b/src/ohos/mod.rs
@@ -3,7 +3,7 @@ use std::borrow::Cow;
 use super::WebViewAttributes;
 use crate::{Error, Rect, RequestAsyncResponder, Result, RGBA};
 use cookie::Cookie;
-use http::{Request, Response};
+use http::{Request, Response, Uri};
 use openharmony_ability::{native_web::WebProxyBuilder, WebViewBuilder, WebViewStyle, Webview};
 use raw_window_handle::HasWindowHandle;
 
@@ -93,13 +93,20 @@ impl InnerWebView {
       .map_err(|e| Error::OpenHarmonyInitError(e.to_string()))?;
 
     let current_id = id.clone();
+    let ipc_webview = webview.clone();
     webview
       .on_controller_attach(move || {
         let _builder = WebProxyBuilder::new(current_id.clone(), "ipc".to_string())
           .add_method("postMessage", |_frame, params| {
             if let Some(ipc_handler) = &ipc_handler {
               let message = params.get(0).unwrap_or(&"".to_string()).to_owned();
-              ipc_handler(Request::new(message));
+              // when load html, url will be base64 and uri parse will failed, so we use about:blank avoid panic
+              let url = ipc_webview.url().unwrap_or("about:blank".to_string());
+              let uri = url
+                .parse::<Uri>()
+                .unwrap_or(Uri::from_static("about:blank"));
+
+              ipc_handler(Request::builder().uri(uri).body(message).unwrap());
             }
           })
           .build()
--- a/src/ohos/mod.rs
+++ b/src/ohos/mod.rs
@@ -48,7 +48,9 @@ impl InnerWebView {
       ..
     } = attributes;
 
-    let u = url.clone().unwrap_or_default();
+    let initial_url = url;
+    let initial_headers = headers;
+    let should_load_initial_url = html.is_none();
 
     let id = id
       .map(|id| id.to_string())
@@ -65,7 +67,6 @@ impl InnerWebView {
         visible: None,
         background_color,
       })
-      .url(u)
       .javascript_enabled(!javascript_disabled)
       .autoplay(autoplay)
       .initialization_scripts(vec![initialization_scripts
@@ -82,10 +83,6 @@ impl InnerWebView {
 
     if let Some(html) = html {
       webview_builder = webview_builder.html(html);
-    }
-
-    if let Some(headers) = headers {
-      webview_builder = webview_builder.headers(headers);
     }
 
     if let Some(user_agent) = user_agent {
@@ -131,6 +128,20 @@ impl InnerWebView {
           (callback)(&webview_id, req, RequestAsyncResponder { responder });
         })
         .unwrap();
+    }
+
+    if should_load_initial_url {
+      if let Some(url) = initial_url.as_deref() {
+        if let Some(headers) = initial_headers {
+          webview
+            .load_url_with_headers(url, headers)
+            .map_err(|e| Error::OpenHarmonyInitError(format!("Failed to load initial url with headers: {}", e)))?;
+        } else {
+          webview
+            .load_url(url)
+            .map_err(|e| Error::OpenHarmonyInitError(format!("Failed to load initial url: {}", e)))?;
+        }
+      }
     }
 
     Ok(Self { id, webview })
--- a/src/ohos/mod.rs
+++ b/src/ohos/mod.rs
@@ -50,7 +50,6 @@ impl InnerWebView {
 
     let initial_url = url;
     let initial_headers = headers;
-    let should_load_initial_url = html.is_none();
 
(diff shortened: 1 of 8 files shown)


lucasfernog:
this is working really well! 

see the [cargo-mobile2 integration](https://github.com/tauri-apps/cargo-mobile2/pull/478) and the [tauri branch](https://github.com/tauri-apps/tauri/tree/feat/open-harmony)

can we call https://docs.openharmony.cn/pages/v5.0/en/application-dev/reference/apis-arkweb/js-apis-webview.md#setwebdebuggingaccess when `webview_attributes.devtools` is true? I'm trying to debug a localhost usage, currently only remote URLs are working on tauri

richerfu:
> this is working really well!
> 
> see the [cargo-mobile2 integration](https://github.com/tauri-apps/cargo-mobile2/pull/478) and the [tauri branch](https://github.com/tauri-apps/tauri/tree/feat/open-harmony)
> 
> can we call https://docs.openharmony.cn/pages/v5.0/en/application-dev/reference/apis-arkweb/js-apis-webview.md#setwebdebuggingaccess when `webview_attributes.devtools` is true? I'm trying to debug a localhost usage, currently only remote URLs are working on tauri

Could you try those changes in your local code which is in `root/oh_modules/.ohpm/@ohos-rs+abilityxxx/xx` ? @lucasfernog 

If it works, i will publish a new version for `@ohos-rs/ability`

lucasfernog:
> > this is working really well!
> > see the [cargo-mobile2 integration](https://github.com/tauri-apps/cargo-mobile2/pull/478) and the [tauri branch](https://github.com/tauri-apps/tauri/tree/feat/open-harmony)
> > can we call https://docs.openharmony.cn/pages/v5.0/en/application-dev/reference/apis-arkweb/js-apis-webview.md#setwebdebuggingaccess when `webview_attributes.devtools` is true? I'm trying to debug a localhost usage, currently only remote URLs are working on tauri
> 
> Could you try those changes in your local code which is in `root/oh_modules/.ohpm/@ohos-rs+abilityxxx/xx` ? @lucasfernog
> 
> If it works, i will publish a new version for `@ohos-rs/ability`

https://github.com/harmony-contrib/openharmony-ability/pull/28#issuecomment-3206258591

lucasfernog:
are the javaScriptOnDocumentStart scripts executed out of order? i'm seeing some weird behavior, is it running in reverse order? 
for now i think i'll just combine all scripts into one.. but would be good to know exactly how the webview runs them

richerfu:
> javaScriptOnDocumentStart

I need to check it. May i tell you later..

lucasfernog:
also can't we really register the custom protocols at runtime? currently we need to provide the names beforehand on the derive macro right?

richerfu:
> also can't we really register the custom protocols at runtime? currently we need to provide the names beforehand on the derive macro right?
 
yeah,we need to register  the  custom protocols  in a very early stage. When we need to register them at runtime, i can try to find some ways  to  support  it

lucasfernog:
is the custom protocol triggered on fetch() calls? tauri uses that for the IPC

for instance this is what i get:
```js
await fetch('tauri://localhost')
VM174:1 Fetch API cannot load tauri://localhost/. URL scheme "tauri" is not supported.
```
even though the page loaded is `tauri://localhost`

lucasfernog:
> > also can't we really register the custom protocols at runtime? currently we need to provide the names beforehand on the derive macro right?
> 
> yeah,we need to register the custom protocols in a very early stage. When we need to register them at runtime, i can try to find some ways to support it

cool that's fine, thanks

src/ohos/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+      .map_err(|e| Error::OpenHarmonyInitError(e.to_string()))?;
+
+    let current_id = id.clone();
+    webview
+      .on_controller_attach(move || {
+        let _builder = WebProxyBuilder::new(current_id.clone(), "ipc".to_string())
+          .add_method("postMessage", |_, params| {
+            if let Some(ipc_handler) = &ipc_handler {
+              let message = params.get(0).unwrap_or(&"".to_string()).to_owned();
+              ipc_handler(Request::new(message));
```
lucasfernog: ideally we would fill the uri field of the request here, Tauri validates that

if the platform doesn't give us the URL of the frame that called the postMessage function, we just fill the current webview URL: https://github.com/tauri-apps/wry/blob/4b44ebb202b00226aae662a5289801b05ea6aaa4/src/webkitgtk/mod.rs#L644

src/ohos/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+          .build()
+          .expect("Failed to build web proxy");
+      })
+      .map_err(|e| {
+        Error::OpenHarmonyInitError(format!("Failed to add controller attach listener: {}", e))
+      })?;
+
+    for (protocol, callback) in custom_protocols {
+      webview
+        .custom_protocol(protocol, move |web, req, _flag| {
```
lucasfernog: async implementation needs https://github.com/harmony-contrib/openharmony-ability/pull/29
we need that to merge this, otherwise some tauri commands will be blocking and have this timeout set here

lucasfernog: for now Tauri is using the ipc_handler fallback, but ideally the custom protocol would also be triggered on window.fetch calls, see https://github.com/tauri-apps/wry/pull/1607/#issuecomment-3206967396 for the bug

richerfu: Please upgrade `openharmony-ability` and `openharmony-ability-derive`.

richerfu:
> is the custom protocol triggered on fetch() calls? tauri uses that for the IPC
> 
> for instance this is what i get:
> 
> ```js
> await fetch('tauri://localhost')
> VM174:1 Fetch API cannot load tauri://localhost/. URL scheme "tauri" is not supported.
> ```
> 
> even though the page loaded is `tauri://localhost`

Fixed.

lucasfernog:
you rock!!

src/ohos/mod.rs
richerfu: I want to know the behavior of gtk when we load webview with `with_html`. We will get html content or base64 content which is invalid URI and will crash with `.uri` in harmony. @lucasfernog

richerfu:
> are the javaScriptOnDocumentStart scripts executed out of order? i'm seeing some weird behavior, is it running in reverse order? for now i think i'll just combine all scripts into one.. but would be good to know exactly how the webview runs them

It run with `lexicographic order`.If we want to run them with array order, we need to use `runJavaScriptOnDocumentStart `, but it depends on api15 or newer. It's too new for us.

richerfu:
> also can't we really register the custom protocols at runtime? currently we need to provide the names beforehand on the derive macro right?

In what scenarios will we register at runtime? From the documentation, it is clear that the protocol can only be registered before the web is initialized. There is no any other method to register :(

src/ohos/mod.rs
richerfu: Fixed, see  detail https://github.com/harmony-contrib/openharmony-ability/issues/32

richerfu:
https://github.com/tauri-apps/tauri/issues/7287

richerfu:
Hi @lucasfernog dear friend, if we want to continue this work, what we should do. Hope your reply thanks :)

FabianLars:
Just so you're not surprised, Lucas is on vacation (or at least out of office) for another week or so.
I think work on this from our side will slow down though because our task at our workplace, which made us look into it in the first place, is done.

richerfu:
Copy that, hope you have a good time. :)

richerfu:
Rebased dev branch. cc @frankplus @lucasfernog @FabianLars

Legend-Master (approved):
Thanks! Awesome work!

LuSrackhall:
👍

PR #1613, opened 2025-09-06
fix(linux): improve download handler's default filename fallback logic

### LINKED ISSUE #14162: [bug] Vanilla js downloads get wrong filename
### Describe the bug

I am using rust as backend and Tauri version "2.8.5".
If I save a file with ([Source](https://codeberg.org/MaerchenfeeimGarten/DOORS/src/commit/9c87622e599512966de5dbce44d491bc47237451/index.html#L1105))
```
        function download_txt() {
            var hiddenElement = document.createElement('a');


            hiddenElement.href = 'data:attachment/text,' + encodeURI(JSON.stringify(geschafft_array));
            hiddenElement.target = '_blank';
            hiddenElement.download = 'savegame.doors';
            hiddenElement.click();
        }
```
it will be saved in the download folder `~/Downloads` (this is good) and will be named "Unknown.doors" althrough the file name is set to "savegame.doors" (this is the issue).
I am testing using the Appimage build on Ubuntu 24.04.2.
I am building in a Debian docker container.
I would expect the app to save the file with the filename "savegame.doors".
This issue is a follow up issue to https://github.com/tauri-apps/tauri/issues/13734

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
# cargo tauri info

[✔] Environment
    - OS: Debian 12.0.0 x86_64 (X64) (Unknown DE on Unknown Session)
    ✔ webkit2gtk-4.1: 2.48.3
    ✔ rsvg2: 2.54.7
    ✔ rustc: 1.89.0 (29483883e 2025-08-04)
    ✔ cargo: 1.89.0 (c24e10642 2025-06-23)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (overridden by environment variable RUSTUP_TOOLCHAIN)

[-] Packages
    - tauri 🦀: 2.8.5
    - tauri-build 🦀: 2.4.1
    - wry 🦀: 0.53.3
    - tao 🦀: 0.34.3
    - tauri-cli 🦀: 2.7.1 (outdated, latest: 2.8.4)

[-] Plugins
    - tauri-plugin-opener 🦀: 2.5.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../src
```

### Stack trace

```text

```

### Additional context

_No response_

FabianLars:
fixes https://github.com/tauri-apps/tauri/issues/14162

it's gotten so ugly lol

### DIFF
--- a/src/webkitgtk/web_context.rs
+++ b/src/webkitgtk/web_context.rs
@@ -327,10 +327,18 @@ impl WebContextExt for super::WebContext {
               .map(|(base, ext)| (base, format!(".{ext}")))
               .unwrap_or((suggested_filename, "".to_string()));
 
-            // for `data:` downloads, webkitgtk will suggest to use the raw data as the filename
-            // for example `"data:attachment/text,sometext"` will result in `text,sometext`
+            // For `data:` downloads, webkitgtk will suggest to use the raw data as the filename if the dev provided no name,
+            // for example `"data:attachment/text,sometext"` will result in `text,sometext` but longer data URLs will
+            // result in a cut-off filename, which makes it hard to predict reliably.
+            // TODO: If this keeps causing problems, just remove it and use whatever file name webkitgtk suggests.
             if uri.starts_with("data:") {
-              suggested_filename = "Unknown";
+              if let Some((_, uri_stripped)) = uri.split_once('/') {
+                if let Some((uri_stripped, _)) = uri_stripped.split_once(',') {
+                  if suggested_filename.starts_with(&format!("{uri_stripped},")) {
+                    suggested_filename = "Unknown";
+                  }
+                }
+              }
             }
 
             download_destination.push(format!("{suggested_filename}{ext}"));


pewsheen:
Maybe we can check if `suggested_filename` starts with `data:` only? Looks like user-agents are treating colon as an invalid character in `download` attribute and converting it to `_`.

FabianLars:
The problem is that the suggested filename doesn't start with that for example for the url of `data:attachment/text,sometext` webkitgtk will set the suggested filename to just `text,sometext`, it's completely void of the `data:` prefix or anything before `/` - i only tested this pr with that example, will double check that it was the same for more "normal" mimetypes eg text/plain or image/png but iirc it was

pewsheen:
> The problem is that the suggested filename doesn't start with that for example for the url of `data:attachment/text,sometext` webkitgtk will set the suggested filename to just `text,sometext`, it's completely void of the `data:` prefix or anything before `/` - i only tested this pr with that example, will double check that it was the same for more "normal" mimetypes eg text/plain or image/png but iirc it was

Uh...right

In that case, I think we can apply this PR first.

pewsheen (approved):

FabianLars:
fun right? :D 

tested it with other mime types and it works the same for them

PR #1617, opened 2025-09-29
Fix display sharing on macOS 14+

andre-digiexam:
Ported over https://github.com/tauri-apps/wry/pull/1196 to work on newest wry.

### DIFF
--- a/examples/simple.rs
+++ b/examples/simple.rs
@@ -14,7 +14,7 @@ fn main() -> wry::Result<()> {
   let window = WindowBuilder::new().build(&event_loop).unwrap();
 
   let builder = WebViewBuilder::new()
-    .with_url("http://tauri.app")
+    .with_url("https://webrtc.github.io/samples/src/content/getusermedia/getdisplaymedia/")
     .with_new_window_req_handler(|url, features| {
       println!("new window req: {url} {features:?}");
       wry::NewWindowResponse::Allow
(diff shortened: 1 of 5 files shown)


andre-digiexam:
Apologies, accidentally pushed to upstream. Feel free to make use of this code however :)

FabianLars:
Does it work well for you? Any issues? I can't remember what it was but i think there were reasons we put it on the side.

andre-digiexam:
> Does it work well for you? Any issues? I can't remember what it was but i think there were reasons we put it on the side.

It works well for us as we only support macOS 13+. I'm just concerned how it'll work when the private method is not available, as I think it was added in macOS 13, and I don't think you can conditionally include a method at runtime with `declare_class!`.

pewsheen:
> Does it work well for you? Any issues? I can't remember what it was but i think there were reasons we put it on the side.

Might related to this one (#1195 Fix getDispayMedia() getUserMedia() permission prompt on macOS)

FabianLars:
> and I don't think you can conditionally include a method at runtime with declare_class!.

That shouldn't be an issue with Delegate methods at least

> Might related to this one (https://github.com/tauri-apps/wry/issues/1195 Fix getDispayMedia() getUserMedia() permission prompt on macOS)

ah i think you're right

andre-digiexam:
@FabianLars Do you think it's worth it to create a new, proper PR with these changes? Or is there lacking nuance that was the reason the original wasn't merged?

FabianLars:
I don't know how to proceed either which is why the old PR was abandoned i guess. So https://github.com/tauri-apps/wry/pull/1196#issuecomment-2042247183 / https://github.com/tauri-apps/wry/issues/1195#issuecomment-2018956566 still needs a solution / some thinking.

Maybe this can also be combined with https://github.com/tauri-apps/wry/tree/feat/permission-request

PR #1619, opened 2025-10-02
Handle wkwebview web content process termination

JeffTsang:
For WKWebView, the webview's content process can get terminated, especially when the app is backgrounded. This results in a blank screen when you resume the app. To fix this, the webview can be reloaded in the webViewWebContentProcessDidTerminate method in WKNavigationDelegate.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -792,6 +792,13 @@ pub struct WebViewAttributes<'a> {
 
   /// Whether JavaScript should be disabled.
   pub javascript_disabled: bool,
+
+  /// Set a handler closure to respond to web content process termination.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Linux / Windows / Android**: Unsupported.
+  pub on_web_content_process_terminate_handler: Option<Box<dyn Fn()>>,
 }
 
 impl Default for WebViewAttributes<'_> {
@@ -834,6 +841,7 @@ impl Default for WebViewAttributes<'_> {
       }),
       background_throttling: None,
       javascript_disabled: false,
+      on_web_content_process_terminate_handler: None,
     }
   }
 }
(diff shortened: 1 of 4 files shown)


FabianLars:
Thanks for the PR! What's your opinion on https://github.com/ionic-team/capacitor/issues/6680 / https://github.com/ionic-team/capacitor/discussions/5488 ? This does seem to be a bit more controversial than you'd expect at first glance.

JeffTsang:
It seems that it might be best to allow developers to handle the process termination event themselves so they can tailor the response to their own apps. Perhaps the WryNavigationDelegate could have an on_web_content_process_terminate_handler similar to the on_page_load_handler.

FabianLars:
sounds reasonable, good idea

JeffTsang:
Great, I'll work on it later.

tdomhan:
Out of curiosity:
In logs related to webview crashs I have come across: "WebPageProxy::tryReloadAfterProcessTermination: process crashed and the client did not handle it, reloading the pageWebPageProxy::reload:". 
Would defining `webViewWebContentProcessDidTerminate` without setting a custom handler that calls reload on the webview change that behavior (as it's maybe considered handled)?

JeffTsang:
Reloading the webview usually works. However, it reloads the current url instead of the root url, so apps that cannot reload the current url will have problems. Also, some people report problems with event handlers or plugins not working properly after reloading.

JeffTsang:
@FabianLars I've added the handler.

JeffTsang:
On further testing, reloading the webview in webViewWebContentProcessDidTerminate doesn't seem to solve the issue of the white screen. So this PR is no longer necessary.

Another potential solution is to toggle the document body's display css property though javascript when the app will enter the foreground.

PR #1624, opened 2025-10-11
feat(ios/macos): Add handler for web content process termination

JeffTsang:
As it turns out, it's still important to handle the termination of the web content process.

### DIFF
--- a/.changes/web-content-process-termination.md
+++ b/.changes/web-content-process-termination.md
@@ -1,0 +1,5 @@
+---
+"wry": minor
+---
+
+Add handler for web content process termination.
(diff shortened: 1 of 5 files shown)


pewsheen (commented):
Thanks for the PR!

Looks like the attribute is restricted to macOS and iOS, I think they could be moved to platform platform-specific section in `lib.rs`:

```rust=
#[cfg(any(target_os = "macos", target_os = "ios"))]
pub(crate) struct PlatformSpecificWebViewAttributes {
  pub on_web_content_process_terminate_handler: Option<Box<dyn Fn()>>,
```

```rust=
#[cfg(any(target_os = "macos", target_os = "ios"))]
pub trait WebViewBuilderExtDarwin {
  fn with_on_web_content_process_terminate_handler(self, handler: impl Fn() + 'static) -> Self;
```


For curiosity, how do you test whether the method works? Right now I am killing all WebContent processes.

JeffTsang:
I've moved the handler to the platform specific attributes.

It's easiest to test in Simulator by putting the app into the background, killing the WebContent process, and bringing the app back into the foreground.

pewsheen (approved):

PR #1625, opened 2025-10-14
refactor(android): use OnBackPressedCallback

lucasfernog:
see https://developer.android.com/reference/android/app/Activity#onBackPressed()


FabianLars (approved):
not sure i tested this right since we already use the new callbakc in tauri but it seems to work right?

PR #1626, opened 2025-10-14
chore: remove rustc/rustdoc args for docsrs cfg to fix docs.rs builds

FabianLars:
seen in https://github.com/winnow-rs/winnow/issues/834#issuecomment-3403140625 and at least in my local docs.rs environment it works, also couldn't spot any differences in the generated docs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,6 @@ targets = [
   "x86_64-pc-windows-msvc",
   "x86_64-apple-darwin",
 ]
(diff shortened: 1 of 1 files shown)


pewsheen (approved):
The only difference would be that `Available on crate feature XXX only.` will not show on local build, but it should work on docs.rs I think.

Image

FabianLars:
That actually works locally (in a docs.rs local copy, not just `cargo doc`) as well
image

PR #1628, opened 2025-10-22
chore: manually release 0.35.5 to fix docsrs

FabianLars:
I hope renovate can handle that even with changefiles being present?


FabianLars:
only one way to find out

PR #1629, opened 2025-10-27
docs: simpler example on README with Nix

longregen:
The previous `shell.nix` version is quite opinionated. Users of `nix` would be expected to know how to build their own `mkShell`, based on this simple `nix-shell` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,24 +189,8 @@ sudo dnf install gtk3-devel webkit2gtk4.1-devel
 
 ###### Nix & NixOS
 
-```nix
-# shell.nix
-
-let
-   # Unstable Channel | Rolling Release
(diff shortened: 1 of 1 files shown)


FabianLars (approved):
I love that the back-and-forth on nix instructions now also concerns this repo and not just tauri anymore 😅

Anyway, thanks for the PR!

longregen:
> I love that the back-and-forth on nix instructions now also concerns this repo and not just tauri anymore 😅
> 
> Anyway, thanks for the PR!

And I apologize for the noise 😅 Thank you! 

If nixos users stumble upon that example, it won't be that super helpful to know more than which are the two packages required, and for non-nixos users, we save them time by reducing the number of characters in the readme.

PR #1633, opened 2025-11-17
feat(mobile): multi window support

lucasfernog:
follow-up for https://github.com/tauri-apps/tao/pull/1154

includes a refactor to only load Android library once

### DIFF
--- a/src/android/kotlin/WryActivity.kt
+++ b/src/android/kotlin/WryActivity.kt
@@ -59,7 +59,7 @@ abstract class WryActivity : AppCompatActivity() {
         if (handleBackNavigation) {
             val callback = object : OnBackPressedCallback(true) {
                 override fun handleOnBackPressed() {
-                    if (::this@WryActivity.mWebView.isInitialized) {
+                    if (this@WryActivity::mWebView.isInitialized) {
                         if (this@WryActivity.mWebView.canGoBack()) {
                             this@WryActivity.mWebView.goBack()
                         } else {
--- a/src/android/kotlin/WryActivity.kt
+++ b/src/android/kotlin/WryActivity.kt
@@ -127,6 +127,20 @@ abstract class WryActivity : AppCompatActivity() {
         Rust.onActivitySaveInstanceState()
     }
 
+    override fun onPause() {
+        super.onPause()
+        if (::mWebView.isInitialized) {
+            mWebView.onPause()
+        }
+    }
+
+    override fun onResume() {
+        super.onResume()
+        if (::mWebView.isInitialized) {
+            mWebView.onResume()
+        }
+    }
+
     override fun onDestroy() {
         super.onDestroy()
         Rust.onActivityDestroy(this)
--- a/src/android/kotlin/WryActivity.kt
+++ b/src/android/kotlin/WryActivity.kt
@@ -5,6 +5,7 @@
 package {{package}}
 
 import android.annotation.SuppressLint
+import android.content.Intent
 import android.os.Build
 import android.os.Bundle
 import android.webkit.WebView
@@ -20,6 +21,7 @@ object WryLifecycleObserver : DefaultLifecycleObserver {
     override fun onCreate(owner: LifecycleOwner) {
         super.onCreate(owner)
         Rust.create()
+        Rust.wryCreate()
     }
 
     override fun onStart(owner: LifecycleOwner) {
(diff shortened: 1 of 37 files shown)


src/wkwebview/mod.rs
```diff
@@ -474,9 +474,9 @@ impl InnerWebView {
       }
       #[cfg(target_os = "ios")]
       {
-        // set all autoresizingmasks
-        webview.setAutoresizingMask(UIViewAutoresizing::from_bits(31).unwrap());
-        // let () = msg_send![webview, setAutoresizingMask: 31];
+        webview.setAutoresizingMask(
```
velocitysystems, line 521: This sets the _autoresizing mask_ for the webview, which is the _legacy_ (pre-Auto Layout) way of handling view resizing in iOS.
- Doesn't automatically respect safe areas (notch, home indicator etc)
- Webview may extend under system UI elements
- Needs additional code to handle safe area insets

It may be worth considering moving to Auto Layout e.g.

```rust
webview.setTranslatesAutoresizingMaskIntoConstraints(false);

// Add constraints to fill parent while respecting safe area
NSLayoutConstraint::activateConstraints(&[
  webview.topAnchor().constraintEqualToAnchor(&parent.safeAreaLayoutGuide().topAnchor()),
  webview.bottomAnchor().constraintEqualToAnchor(&parent.safeAreaLayoutGuide().bottomAnchor()),
  webview.leadingAnchor().constraintEqualToAnchor(&parent.safeAreaLayoutGuide().leadingAnchor()),
  webview.trailingAnchor().constraintEqualToAnchor(&parent.safeAreaLayoutGuide().trailingAnchor()),
]);
```

src/android/main_pipe.rs
```diff
@@ -7,36 +7,129 @@ use crossbeam_channel::*;
...
+      .unwrap_or_default();
+    if equals {
+      return Some(*activity_id);
+    }
+  }
+  None
+}
+
+pub fn first_activity_id() -> Option<ActivityId> {
+  ACTIVITY_PROXY.lock().unwrap().keys().next().cloned()
```
velocitysystems, line 113: Possible race condition if activities are _destroyed_ while being accessed.

src/android/kotlin/WryActivity.kt
```diff
@@ -75,70 +111,44 @@ abstract class WryActivity : AppCompatActivity() {
...
-
-    override fun onPause() {
-        super.onPause()
-        pause()
-    }
-
-    override fun onStop() {
-        super.onStop()
-        stop()
+        id = savedInstanceState?.getInt(ACTIVITY_ID_KEY) ?: intent.extras?.getInt(ACTIVITY_ID_KEY) ?: hashCode()
```
velocitysystems, line 114: `hashCode()` fallback could _theoretically_ produce duplicate IDs, breaking message routing.

src/android/mod.rs
```diff
@@ -93,18 +94,49 @@ type EvalCallback = Box<dyn Fn(String) + Send + 'static>;
 pub static EVAL_ID_GENERATOR: Counter = Counter::new();
 pub static EVAL_CALLBACKS: OnceCell<Mutex<HashMap<i32, EvalCallback>>> = OnceCell::new();
 
+pub fn destroy_webview(activity_id: ActivityId, webview_id: &WebviewId) {
```
velocitysystems, line 91: While it doesn't pose a significant memory leak, do we have a similar function i.e. `destroy_activity(..)` where we clean up `ACTIVITY_PROXY`?

src/wkwebview/mod.rs
lucasfernog: that's something we should investigate in a separate PR - my first attempt resulted in a crash
and I believe this would be consider a breaking change too, most users offset the notch etc via HTML

src/android/main_pipe.rs
lucasfernog: ACTIVITY_PROXY is a mutex and first_activity_id is only used in cases where the actual activity instance doesn't matter

src/android/kotlin/WryActivity.kt
lucasfernog: that's why we are persisting the id in the saved instance state

src/android/mod.rs
lucasfernog: nice catch, [`8d88361`](https://github.com/tauri-apps/wry/pull/1633/commits/8d8836116aaaef271775e2958f3e38c4b5d18b04)

FabianLars (approved):

PR #1637, opened 2025-12-08
example: Add GPUI example

huacnlee:
Reopen #1383 to add GPUI example.

image

Now GPUI has release it crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,16 +19,62 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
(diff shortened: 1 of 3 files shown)


pewsheen:
Do we have policies for adding a dedicated library to examples? If we are not doing this, maybe it can be placed in gpui's example instead. @FabianLars 

Also, would it be better to move this into an independent cargo project under the example folder? Then we can remove the gpui from the dev-dependency since it's not really needed for the build.

FabianLars:
There's no policy in place. iirc we already do this somewhere (well, at least in the outsourced cargo-packager). I have to admit that this feels a bit out of place though looking at the current examples. eg i don't have any plans to add iced, egui, etc examples as well 🤔 

> Also, would it be better to move this into an independent cargo project under the example folder? Then we can remove the gpui from the dev-dependency since it's not really needed for the build.

Seeing the insane lockfile diff this sounds reasonable though i'm usually not a big fan of that. Maybe we should check the compile time/size implications before that.

PR #1409, opened 2024-11-04
fix(windows): set WebView2 `AllowExternalDrop` to false

amrbashir:
closes tauri-apps/tauri#11274

PR #1638, opened 2025-12-13
fix(windows): enable drag&drop for initially hidden windows (fix tauri#14643)

sht2017:
Fix for #1639, originally opened at tauri-apps/tauri#14643.

### DIFF
--- a/.changes/webview2-drag-drop-hidden-window.md
+++ b/.changes/webview2-drag-drop-hidden-window.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix WebView2 drag and drop for windows created hidden (visible=false) by re-registering drop targets after the window is shown/resized.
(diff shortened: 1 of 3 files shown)


Legend-Master:
Hey, thanks for contributing

I am still wondering what's really the cause of this problem, it seems to me that this only happens when you set `with_focused(false)` on the wry webview and `with_visible(false)` on tao, and even with that, I don't always reproduce it, and to me, it seems like if I remove the `SetAllowExternalDrop` call, I can no longer reproduce it

@amrbashir do you still remember how `SetAllowExternalDrop(false)` call works added in #1409? I have no problem getting the drag drop events when dragging files/links/folders from the recent files view in File Explorer with or without this call 🤔

amrbashir:
> @amrbashir do you still remember how `SetAllowExternalDrop(false)` call works added in #1409? I have no problem getting the drag drop events when dragging files/links/folders from the recent files view in File Explorer with or without this call 🤔

I don't really remember but it may have been an issue on Windows 11 only

PR #1646, opened 2025-12-28
refactor: remove redundant clones in WebView and download handling

Tunglies:
Cleanup some needless clone.

### DIFF
--- a/.changes/cleanup-redundant-clone.md
+++ b/.changes/cleanup-redundant-clone.md
@@ -1,0 +1,5 @@
+---
+"wry": patch:minor
+---
+
+Remove redundant clones in WebView and download handling. No user facing changes.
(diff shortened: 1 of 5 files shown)


.changes/cleanup-redundant-clone.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": patch:minor
```
Legend-Master: `patch:minor` doesn't really make sense, the `patch` means we do a semantic versioning `patch` bump (e.g. `1.0.0` -> `1.0.1`), and the `:xxx` part is used by the [.changes/config.json](https://github.com/tauri-apps/wry/blob/dev/.changes/config.json) config to categorize changes, and in this repo, we didn't config that so you can leave it empty

```suggestion
"wry": patch
```

pewsheen (approved):

PR #1649, opened 2025-12-31
enhance: Add an option to control browser-level autofill behavior on Windows

### LINKED ISSUE #1535: Expose `IsGeneralAutofillEnabled` WebView2 setting
WebView2 on Windows uses some weird autofill (called "Suggestions") that ignores `autocomplete="off"` attribute on input fields and you have to add obscure `aria-autocomplete="none"` and `list="autocompleteOff"`. I really don't want to add it to every single field and also in some cases (i.e. external components or websites) I can't control it.

It would be very helpful to just disable autofill globally with `IsGeneralAutofillEnabled` setting: https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2settings?view=webview2-winrt-1.0.3065.39#isgeneralautofillenabled

Expose `IsGeneralAutofillEnabled` so it will be possible to disable it in Tauri config.

aoxiangtianyu-go:
close tauri-apps/wry#1535

This PR adds an optional WebView attribute to control browser-level autofill behavior on Windows.

The motivation for exposing this option is that browser-level autofill behavior in WebView2
is controlled by the underlying runtime and cannot always be influenced through standard
HTML attributes alone. In some scenarios, applications need a way to explicitly opt out of
browser-managed autofill at the WebView level.

The default behavior is preserved unless the option is explicitly set.  
The change is limited to the WebView2 implementation and does not affect other platforms.

### Related issues
- https://github.com/tauri-apps/wry/issues/1535
- https://github.com/clash-verge-rev/clash-verge-rev/issues/5944

### Follow-up
I will submit a separate PR to `tauri` to expose this API at the application level.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -804,10 +804,10 @@ pub struct WebViewAttributes<'a> {
   ///
   /// - **Windows**: Supported. On Windows, WebView2's autofill feature (called
   ///   "Suggestions") may not honor `autocomplete="off"` attributes on input
-  ///   elements in some cases. When this option is `Some(false)`, that autofill
+  ///   elements in some cases. When this option is `false`, that autofill
   ///   behavior will be disabled.
   /// - **macOS / Linux / Android / iOS**: Unsupported and ignored.
-  pub general_autofill_enabled: Option<bool>,
+  pub general_autofill_enabled: bool,
 }
 
 impl Default for WebViewAttributes<'_> {
@@ -850,7 +850,7 @@ impl Default for WebViewAttributes<'_> {
       }),
       background_throttling: None,
       javascript_disabled: false,
-      general_autofill_enabled: None,
+      general_autofill_enabled: true,
     }
   }
 }
@@ -1433,11 +1433,11 @@ impl<'a> WebViewBuilder<'a> {
   ///
   /// - **Windows**: Supported. On Windows, WebView2's autofill feature (called
   ///   "Suggestions") may not honor `autocomplete="off"` attributes on input
-  ///   elements in some cases. When this option is `Some(false)`, that autofill
+  ///   elements in some cases. When this option is `false`, that autofill
   ///   behavior will be disabled.
   /// - **macOS / Linux / Android / iOS**: Unsupported and ignored.
   pub fn with_general_autofill_enabled(mut self, enabled: bool) -> Self {
-    self.attrs.general_autofill_enabled = Some(enabled);
+    self.attrs.general_autofill_enabled = enabled;
     self
   }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -795,10 +795,13 @@ pub struct WebViewAttributes<'a> {
 
   /// Controls the WebView's browser-level general autofill behavior.
   ///
+  /// **This option does not disable password or credit card autofill.**
+  ///
   /// When enabled, the WebView may automatically populate form fields using
   /// previously stored data such as addresses or contact information.
   ///
-  /// If set to `None`, the default platform behavior is preserved.
+  /// If not specified, this is `true` by default.
+  /// Setting this to `false` preserves the default platform behavior.
   ///
   /// ## Platform-specific
   ///
@@ -1424,10 +1427,13 @@ impl<'a> WebViewBuilder<'a> {
 
   /// Controls the WebView's browser-level general autofill behavior.
   ///
+  /// **This option does not disable password or credit card autofill.**
+  ///
   /// When enabled, the WebView may automatically populate form fields using
   /// previously stored data such as addresses or contact information.
   ///
-  /// If set to `None`, the default platform behavior is preserved.
+  /// If not specified, this is `true` by default.
+  /// Setting this to `false` preserves the default platform behavior.
   ///
   /// ## Platform-specific
   ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -792,6 +792,22 @@ pub struct WebViewAttributes<'a> {
 
   /// Whether JavaScript should be disabled.
   pub javascript_disabled: bool,
+
+  /// Controls the WebView's browser-level general autofill behavior.
+  ///
+  /// When enabled, the WebView may automatically populate form fields using
+  /// previously stored data such as addresses or contact information.
+  ///
+  /// If set to `None`, the default platform behavior is preserved.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows**: Supported. On Windows, WebView2's autofill feature (called
+  ///   "Suggestions") may not honor `autocomplete="off"` attributes on input
+  ///   elements in some cases. When this option is `Some(false)`, that autofill
+  ///   behavior will be disabled.
+  /// - **macOS / Linux / Android / iOS**: Unsupported and ignored.
+  pub general_autofill_enabled: Option<bool>,
 }
 
 impl Default for WebViewAttributes<'_> {
@@ -834,6 +850,7 @@ impl Default for WebViewAttributes<'_> {
       }),
       background_throttling: None,
       javascript_disabled: false,
+      general_autofill_enabled: None,
     }
   }
 }
@@ -1402,6 +1419,25 @@ impl<'a> WebViewBuilder<'a> {
   /// Whether JavaScript should be disabled.
   pub fn with_javascript_disabled(mut self) -> Self {
     self.attrs.javascript_disabled = true;
+    self
+  }
+
+  /// Controls the WebView's browser-level general autofill behavior.
+  ///
+  /// When enabled, the WebView may automatically populate form fields using
+  /// previously stored data such as addresses or contact information.
+  ///
+  /// If set to `None`, the default platform behavior is preserved.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows**: Supported. On Windows, WebView2's autofill feature (called
+  ///   "Suggestions") may not honor `autocomplete="off"` attributes on input
+  ///   elements in some cases. When this option is `Some(false)`, that autofill
+  ///   behavior will be disabled.
+  /// - **macOS / Linux / Android / iOS**: Unsupported and ignored.
+  pub fn with_general_autofill_enabled(mut self, enabled: bool) -> Self {
+    self.attrs.general_autofill_enabled = Some(enabled);
     self
   }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -801,7 +801,6 @@ pub struct WebViewAttributes<'a> {
   /// previously stored data such as addresses or contact information.
   ///
   /// If not specified, this is `true` by default.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -792,6 +792,22 @@ pub struct WebViewAttributes<'a> {
...
+  /// If set to `None`, the default platform behavior is preserved.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows**: Supported. On Windows, WebView2's autofill feature (called
+  ///   "Suggestions") may not honor `autocomplete="off"` attributes on input
+  ///   elements in some cases. When this option is `Some(false)`, that autofill
+  ///   behavior will be disabled.
+  /// - **macOS / Linux / Android / iOS**: Unsupported and ignored.
+  pub general_autofill_enabled: Option<bool>,
```
FabianLars: we could also use bool and default to true like in `focused`, or just rename it to `general_autofill_disabled` idk. Just not really a fan of Option<bool> in general but if you prefer we can keep it ig.

src/lib.rs
```diff
@@ -1405,6 +1422,25 @@ impl<'a> WebViewBuilder<'a> {
...
+  ///
+  /// If set to `None`, the default platform behavior is preserved.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows**: Supported. On Windows, WebView2's autofill feature (called
+  ///   "Suggestions") may not honor `autocomplete="off"` attributes on input
+  ///   elements in some cases. When this option is `Some(false)`, that autofill
+  ///   behavior will be disabled.
+  /// - **macOS / Linux / Android / iOS**: Unsupported and ignored.
```
FabianLars: There's no Option in this method so speaking of None and Some(false) is wrong.

FabianLars (commented):
thanks for the PR! just some nits :)

src/lib.rs
aoxiangtianyu-go: Yeah, you're right

src/lib.rs
aoxiangtianyu-go: OK,thanks

aoxiangtianyu-go:
I’ve addressed the review feedback and pushed the updates. Please let me know if there’s anything I should do. 
Thanks!

src/webview2/mod.rs
```diff
@@ -581,6 +581,12 @@ impl InnerWebView {
       }
     }
 
+    if let Ok(settings4) = settings.cast::<ICoreWebView2Settings4>() {
+      settings4.SetIsGeneralAutofillEnabled(
+      attributes.general_autofill_enabled,
```
Legend-Master: The format looks a bit off here, could you run `cargo fmt`?

src/lib.rs
```diff
@@ -792,6 +792,22 @@ pub struct WebViewAttributes<'a> {
 
   /// Whether JavaScript should be disabled.
   pub javascript_disabled: bool,
+
+  /// Controls the WebView's browser-level general autofill behavior.
+  ///
+  /// When enabled, the WebView may automatically populate form fields using
+  /// previously stored data such as addresses or contact information.
+  ///
+  /// If set to `None`, the default platform behavior is preserved.
```
Legend-Master: We need to update the `None` here to reflect the change made in 2eadced0610311c112921d701d69554d6b946197

src/lib.rs
```diff
@@ -1405,6 +1422,25 @@ impl<'a> WebViewBuilder<'a> {
     self
   }
 
+  /// Controls the WebView's browser-level general autofill behavior.
+  ///
+  /// When enabled, the WebView may automatically populate form fields using
+  /// previously stored data such as addresses or contact information.
+  ///
+  /// If set to `None`, the default platform behavior is preserved.
```
Legend-Master: Same as above, and it might be good to also add that this is `true` by default

src/lib.rs
```diff
@@ -1405,6 +1422,25 @@ impl<'a> WebViewBuilder<'a> {
     self
   }
 
+  /// Controls the WebView's browser-level general autofill behavior.
+  ///
+  /// When enabled, the WebView may automatically populate form fields using
```
Legend-Master, line 1431: We should also note that this doesn't disable the password auto fill

> https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings4?view=webview2-1.0.3650.58#put_isgeneralautofillenabled:~:text=This%20excludes%20password%20and%20credit%20card%20information.

aoxiangtianyu-go:
Thanks for the feedback! I'm currently on Chinese New Year holiday.
I will address the changes after the holiday.

aoxiangtianyu-go:
Thanks for the review! 
I've updated the docs to reflect the default value and clarified that password and credit card autofill are not affected.

src/lib.rs
```diff
@@ -1405,6 +1425,28 @@ impl<'a> WebViewBuilder<'a> {
...
 
+  /// Controls the WebView's browser-level general autofill behavior.
+  ///
+  /// **This option does not disable password or credit card autofill.**
+  ///
+  /// When enabled, the WebView may automatically populate form fields using
+  /// previously stored data such as addresses or contact information.
+  ///
+  /// If not specified, this is `true` by default.
+  /// Setting this to `false` preserves the default platform behavior.
```
Legend-Master:
```suggestion
```

This doesn't seem correct, let's remove it

src/lib.rs
```diff
@@ -792,6 +792,25 @@ pub struct WebViewAttributes<'a> {
...
+
+  /// Controls the WebView's browser-level general autofill behavior.
+  ///
+  /// **This option does not disable password or credit card autofill.**
+  ///
+  /// When enabled, the WebView may automatically populate form fields using
+  /// previously stored data such as addresses or contact information.
+  ///
+  /// If not specified, this is `true` by default.
+  /// Setting this to `false` preserves the default platform behavior.
```
Legend-Master:
```suggestion
```

This doesn't seem correct, let's remove it

aoxiangtianyu-go:
OK Thanks!

Legend-Master (approved):
Thanks

src/lib.rs
```diff
@@ -792,6 +792,24 @@ pub struct WebViewAttributes<'a> {
...
+  /// When enabled, the WebView may automatically populate form fields using
+  /// previously stored data such as addresses or contact information.
+  ///
+  /// If not specified, this is `true` by default.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows**: Supported. On Windows, WebView2's autofill feature (called
+  ///   "Suggestions") may not honor `autocomplete="off"` attributes on input
+  ///   elements in some cases. When this option is `false`, that autofill
```
FabianLars, line 809: oof i had a comment pending here for ages but didn't see that. Do you have some source or references for that issue? I don't think i heard about this before.

aoxiangtianyu-go: Hi, thanks for checking! The behavior I mentioned is documented in the Wry issue https://github.com/tauri-apps/wry/issues/1535, which explains that WebView2’s “Suggestions” (autofill) may ignore `autocomplete="off"` on input elements in some cases.  

In practice, i’ve observed this in https://github.com/clash-verge-rev/clash-verge-rev/issues/5944 , where using `autocomplete="new-password"` still triggered autofill.  

The Wry issue references the official WebView2 setting [`CoreWebView2Settings.IsGeneralAutofillEnabled`](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2settings?view=webview2-winrt-1.0.3065.39) as a way to control this behavior globally.

Legend-Master: > WebView2’s “Suggestions” (autofill) may ignore autocomplete="off" on input elements in some cases.

I can't really reproduce that though

> where using autocomplete="new-password" still triggered autofill.

That sounds correct no? You asked the browser to trigger autocomplete with new-password

I think maybe we could remove this part from the docs, I don't mind adding this in the wry webview settings though

aoxiangtianyu-go: > I can't really reproduce that though

I was able to successfully reproduce this issue in the clash-verge-rev/clash-ver#5944. However, when I created a brand new Tauri project to build a minimal reproduction case, I failed to reproduce it, and the root cause remains unclear.

> That sounds correct no? You asked the browser to trigger autocomplete with new-password

This is not the intended behavior. According to the MDN documentation for the `new-password` attribute value[`new-password`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#new-password): *"This may be used by the browser both to avoid accidentally filling in an existing password and to offer assistance in creating a secure password."*
Additionally, [MDN security implementation guide](https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion#disabling_autocompletion) explicitly states: *"If you are defining a user management page where a user can specify a new password for another person, and therefore, you want to **prevent** autofilling of password fields, you can use autocomplete="new-password"."*
These documents clearly indicate that the intended behavior of `new-password` is to avoid unwanted autofill of existing passwords, rather than triggering autofill.

> I think maybe we could remove this part from the docs, I don't mind adding this in the wry webview settings though

I will fully follow the decision on whether to remove this part from the docs or add it to the wry webview settings.

pewsheen: I was previously handling an issue with `autocomplete=off`, where Chrome would ignore it if the input was named `password` or had the type `password`. But I'm unsure if this behavior exists in WebView2, as it's also Chromium-based. 

> These documents clearly indicate that the intended behavior of new-password is to avoid unwanted autofill of existing passwords, rather than triggering autofill.

Though this is not directly related to the issue, just curious.
I'm not quite understanding what you want to achieve with `autocomplete="new-password"`? I think giving it `autocomplete="new-password"` attribute is telling the browser to autofill with a new password from a password manager?
If your goal is to turn off the autocomplete, and if it is the logic from WebView2 that it will ignore `autocomplete="off"` in some specific fields, maybe setting `general_autofill_enabled` to `false` is the solution instead of setting `autocomplete="new-password"`?

aoxiangtianyu-go: > If your goal is to turn off the autocomplete, and if it is the logic from WebView2 that it will ignore `autocomplete="off"` in some specific fields, maybe setting `general_autofill_enabled` to `false` is the solution instead of setting `autocomplete="new-password"`?

Exactly. Currently, Tauri and Wry do not expose `ICoreWebView2Settings4::SetIsGeneralAutofillEnabled`, so using `autocomplete="new-password"` in the frontend is merely a temporary **workaround** we're forced to use.

However, as we discussed, these frontend hacks are unreliable because WebView2/Chromium often overrides them. This is precisely why I initiated this PR: to provide a robust, native way to disable autofill at the WebView level during construction, rather than relying on inconsistent HTML attributes.

PR #1652, opened 2026-01-04
refactor: improve performance in wkwebview by optimizing string handling and memory usage

Tunglies:


Tunglies:
We already have some likely usage for hardcode &str with `ns_string!`
```rust
// file: src/wkwebview/mod
let proxies: Retained<NSArray<NSObject>> = NSArray::arrayWithObject(&*proxy_config);
        data_store.setValue_forKey(Some(&proxies), ns_string!("proxyConfigurations"));
```

Shall we rewrite the remains `NSString::from_str(&'static str)` to `ns_string!` ? etc.

```rust
pub const IPC_MESSAGE_HANDLER_NAME: &str = "ipc";

// rewrite
let ipc = NSString::from_str(IPC_MESSAGE_HANDLER_NAME);
// to
let ipc = ns_string!(IPC_MESSAGE_HANDLER_NAME);
```

This would migrate runtime cost to compile time cost.

CC @Legend-Master

FabianLars:
> Shall we rewrite the remains NSString::from_str(&'static str) to ns_string! ? etc.

well, why not? 🤷 Actually half-serious question, can you think of a reason not to do it?

Tunglies:
I thought it was I have not carefully reviewd the source code during the migration, since no any feedback. We good though.

Tunglies:
Should be able to review now :3

Legend-Master (approved):
Thanks

Legend-Master:
The doc test is failing because of https://github.com/rust-lang/rust/issues/151028, unrelated to the PR

PR #1654, opened 2026-01-09
feat: add permission handler API across webview backends

### LINKED ISSUE #81: Support for web APIs that require permissions
Currently (at least on Linux), doing any action that requires permission (such as geolocation) fails as the permission request is immediately denied.

There should be a way to specify what permissions the app should have so they are automatically allowed.

- [x] Yes (at least for Linux and maybe windows)
- [ ] No

**Additional context**
For linux: https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#WebKitWebView-permission-request

F0RLE:
## Description

This PR adds a unified **Permission Handler API** for intercepting webview permission requests across WebView2, WKWebView, WebKitGTK, and Android.

The API lets applications decide how to handle permission requests at runtime by returning `Allow`, `Deny`, or `Default` from a new `WebViewBuilder::with_permission_handler` callback.

`Default` falls back to the platform or browser default behavior instead of forcing an explicit prompt. This keeps the API portable across backends where prompt/default handling is exposed differently.

### Key Changes

- Added `WebViewBuilder::with_permission_handler`.
- Added `PermissionKind` for webview permission categories such as camera, microphone, geolocation, notifications, display capture, MIDI, local fonts, file system access, autoplay, and more.
- Added `PermissionResponse` with `Allow`, `Deny`, and `Default`.
- Added a `permission_handler` example using [permission.site](https://permission.site).

### Platform Implementations

- **Windows (WebView2)**
  - Handles WebView2 `PermissionRequested` events.
  - Maps WebView2 permission kinds to `PermissionKind`.
  - Applies `Allow` and `Deny`; `Default` leaves WebView2 default handling in place.

- **macOS / iOS (WKWebView)**
  - Handles `requestMediaCapturePermission`.
  - Maps camera and microphone capture requests to `PermissionKind`.
  - Maps `Default` to WebKit's prompt/default permission flow.
  - Handles combined `CameraAndMicrophone` requests by checking camera and microphone separately, with deny taking precedence.

- **Linux (WebKitGTK)**
  - Handles the `permission-request` signal.
  - Supports geolocation, notifications, pointer lock, camera, microphone, and display capture where available.
  - Uses `is_for_display_device()` with the `v2_42` feature and falls back to display-capture detection for older WebKitGTK versions.
  - Leaves the request unhandled for `Default`, so WebKitGTK can apply its default behavior.

- **Android**
  - Handles `WebChromeClient.onPermissionRequest` and geolocation permission prompts through JNI.
  - Supports camera, microphone, protected media, MIDI, and geolocation requests.
  - Scopes permission handlers per WebView so multi-webview apps route requests to the correct handler.
  - `Deny` rejects the Android request; `Allow` and `Default` continue the normal Android runtime permission flow.
  - Android runtime permissions may still trigger native OS prompts before access is granted.

### Example

```bash
cargo run --example permission_handler
```

### DIFF
--- a/examples/permission_handler.rs
+++ b/examples/permission_handler.rs
@@ -26,7 +26,7 @@ fn main() -> wry::Result<()> {
     .with_url("https://permission.site/")
     .with_permission_handler(|kind| {
       let response = match kind {
-        PermissionKind::Geolocation => PermissionResponse::Prompt,
+        PermissionKind::Geolocation => PermissionResponse::Default,
         _ => PermissionResponse::Allow,
       };
       println!("[permission] {kind} → {response}");
--- a/examples/permission_handler.rs
+++ b/examples/permission_handler.rs
@@ -22,7 +22,7 @@ fn main() -> wry::Result<()> {
     .build(&event_loop)
     .unwrap();
 
-  let _webview = WebViewBuilder::new()
+  let builder = WebViewBuilder::new()
     .with_url("https://permission.site/")
     .with_permission_handler(|kind| {
       let response = match kind {
@@ -31,8 +31,27 @@ fn main() -> wry::Result<()> {
       };
       println!("[permission] {kind} → {response}");
       response
-    })
-    .build(&window)?;
+    });
+
+  #[cfg(any(
+    target_os = "windows",
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "android"
+  ))]
+  let _webview = builder.build(&window)?;
+  #[cfg(not(any(
+    target_os = "windows",
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "android"
+  )))]
+  let _webview = {
+    use tao::platform::unix::WindowExtUnix;
+    use wry::WebViewBuilderExtUnix;
+    let vbox = window.default_vbox().unwrap();
+    builder.build_gtk(vbox)?
+  };
 
   event_loop.run(move |event, _, control_flow| {
     *control_flow = ControlFlow::Wait;
--- a/examples/permission_handler.rs
+++ b/examples/permission_handler.rs
@@ -1,0 +1,47 @@
+// Copyright 2020-2023 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+//! Example demonstrating the permission handler API.
+//!
+//! Run:  cargo run --example permission_handler
+//! Then click the buttons and watch the terminal output.
+
+fn main() -> wry::Result<()> {
+  use tao::{
+    event::{Event, WindowEvent},
+    event_loop::{ControlFlow, EventLoop},
+    window::WindowBuilder,
+  };
+  use wry::{PermissionKind, PermissionResponse, WebViewBuilder};
+
+  let event_loop = EventLoop::new();
+  let window = WindowBuilder::new()
+    .with_title("Permission Handler Example")
+    .with_inner_size(tao::dpi::LogicalSize::new(800, 600))
+    .build(&event_loop)
+    .unwrap();
+
+  let _webview = WebViewBuilder::new()
+    .with_url("https://permission.site/")
+    .with_permission_handler(|kind| {
+      let response = match kind {
+        PermissionKind::Geolocation => PermissionResponse::Prompt,
+        _ => PermissionResponse::Allow,
+      };
+      println!("[permission] {kind} → {response}");
+      response
+    })
+    .build(&window)?;
+
+  event_loop.run(move |event, _, control_flow| {
+    *control_flow = ControlFlow::Wait;
+    if let Event::WindowEvent {
+      event: WindowEvent::CloseRequested,
+      ..
+    } = event
+    {
+      *control_flow = ControlFlow::Exit;
+    }
+  });
+}
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -502,6 +502,9 @@ pub unsafe fn onPageLoaded(mut env: JNIEnv, _: JClass, webview_id: JString, url:
   }
 }
 
+/// Returns true to deny the Android request.
+///
+/// Returns false to let Kotlin continue with Android's normal runtime permission flow.
 #[allow(non_snake_case)]
 pub unsafe fn onPermissionRequestNative(
   mut env: JNIEnv,
@@ -509,8 +512,6 @@ pub unsafe fn onPermissionRequestNative(
   webview_id: JString,
   resources: jni::objects::JObjectArray,
 ) -> jboolean {
-  // Return true to deny the Android request. false lets Kotlin continue with
-  // Android's normal runtime permission flow.
   let mut denied = false;
   let Ok(webview_id) = env.get_string(&webview_id) else {
     return false.into();
@@ -547,6 +548,9 @@ pub unsafe fn onPermissionRequestNative(
   denied.into()
 }
 
+/// Returns true to deny geolocation.
+///
+/// Returns false to let Kotlin continue with Android's normal runtime permission flow.
 #[allow(non_snake_case)]
 pub unsafe fn onGeolocationPermissionRequestNative(
   mut env: JNIEnv,
@@ -554,8 +558,6 @@ pub unsafe fn onGeolocationPermissionRequestNative(
   webview_id: JString,
   _origin: JString,
 ) -> jboolean {
-  // Return true to deny geolocation. false lets Kotlin continue with Android's
-  // normal runtime permission flow.
   let Ok(webview_id) = env.get_string(&webview_id) else {
     return false.into();
   };
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -509,6 +509,8 @@ pub unsafe fn onPermissionRequestNative(
   webview_id: JString,
   resources: jni::objects::JObjectArray,
 ) -> jboolean {
+  // Return true to deny the Android request. false lets Kotlin continue with
+  // Android's normal runtime permission flow.
   let mut denied = false;
   let Ok(webview_id) = env.get_string(&webview_id) else {
     return false.into();
@@ -535,9 +537,7 @@ pub unsafe fn onPermissionRequestNative(
 
           match (handler.handler)(kind) {
             PermissionResponse::Deny => denied = true,
-            PermissionResponse::Allow
-            | PermissionResponse::Default
-            | PermissionResponse::Prompt => {}
+            PermissionResponse::Allow | PermissionResponse::Default => {}
           }
         }
       }
@@ -554,6 +554,8 @@ pub unsafe fn onGeolocationPermissionRequestNative(
   webview_id: JString,
   _origin: JString,
 ) -> jboolean {
+  // Return true to deny geolocation. false lets Kotlin continue with Android's
+  // normal runtime permission flow.
   let Ok(webview_id) = env.get_string(&webview_id) else {
     return false.into();
   };
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -93,7 +93,7 @@ macro_rules! android_binding {
       RustWebChromeClient,
       onPermissionRequestNative,
       [JString, jni::objects::JObjectArray],
-      jint
+      jboolean
     );
     android_fn!(
       $domain,
@@ -101,7 +101,7 @@ macro_rules! android_binding {
       RustWebChromeClient,
       onGeolocationPermissionRequestNative,
       [JString, JString],
-      jint
+      jboolean
     );
   }};
 }
@@ -508,17 +508,15 @@ pub unsafe fn onPermissionRequestNative(
   _: JClass,
   webview_id: JString,
   resources: jni::objects::JObjectArray,
-) -> jint {
-  let mut allowed = false;
+) -> jboolean {
   let mut denied = false;
-  let mut prompt = false;
   let Ok(webview_id) = env.get_string(&webview_id) else {
-    return 2;
+    return false.into();
   };
   let webview_id = webview_id.to_str().ok().unwrap_or_default();
   let permission_handlers = PERMISSION_HANDLER.lock().unwrap();
   let Some(handler) = permission_handlers.get(webview_id) else {
-    return 2;
+    return false.into();
   };
 
   if let Ok(size) = env.get_array_length(&resources) {
@@ -536,26 +534,17 @@ pub unsafe fn onPermissionRequestNative(
           };
 
           match (handler.handler)(kind) {
-            PermissionResponse::Allow => allowed = true,
             PermissionResponse::Deny => denied = true,
-            PermissionResponse::Prompt => prompt = true,
-            PermissionResponse::Default => {}
+            PermissionResponse::Allow
+            | PermissionResponse::Default
+            | PermissionResponse::Prompt => {}
           }
         }
       }
     }
   }
 
-  // Consolidated decision logic
-  if denied {
-    1 // Deny
-  } else if allowed {
-    0 // Allow
-  } else if prompt {
-    3 // Prompt
-  } else {
-    2 // Default
-  }
+  denied.into()
 }
 
 #[allow(non_snake_case)]
@@ -564,20 +553,19 @@ pub unsafe fn onGeolocationPermissionRequestNative(
   _: JClass,
   webview_id: JString,
   _origin: JString,
-) -> jint {
+) -> jboolean {
   let Ok(webview_id) = env.get_string(&webview_id) else {
-    return 2;
+    return false.into();
   };
   let webview_id = webview_id.to_str().ok().unwrap_or_default();
   let permission_handlers = PERMISSION_HANDLER.lock().unwrap();
   let Some(handler) = permission_handlers.get(webview_id) else {
-    return 2;
+    return false.into();
   };
 
-  match (handler.handler)(PermissionKind::Geolocation) {
-    PermissionResponse::Allow => 0,
-    PermissionResponse::Deny => 1,
-    PermissionResponse::Default => 2,
-    PermissionResponse::Prompt => 3,
-  }
+  matches!(
+    (handler.handler)(PermissionKind::Geolocation),
+    PermissionResponse::Deny
+  )
+  .into()
 }
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -95,6 +95,14 @@ macro_rules! android_binding {
       [JString, jni::objects::JObjectArray],
       jint
     );
+    android_fn!(
+      $domain,
+      $package,
+      RustWebChromeClient,
+      onGeolocationPermissionRequestNative,
+      [JString, JString],
+      jint
+    );
   }};
 }
 
@@ -543,5 +551,29 @@ pub unsafe fn onPermissionRequestNative(
     3 // Prompt
   } else {
     2 // Default
+  }
+}
+
+#[allow(non_snake_case)]
+pub unsafe fn onGeolocationPermissionRequestNative(
+  mut env: JNIEnv,
+  _: JClass,
+  webview_id: JString,
+  _origin: JString,
+) -> jint {
+  let Ok(webview_id) = env.get_string(&webview_id) else {
+    return 2;
+  };
+  let webview_id = webview_id.to_str().ok().unwrap_or_default();
+  let permission_handlers = PERMISSION_HANDLER.lock().unwrap();
+  let Some(handler) = permission_handlers.get(webview_id) else {
+    return 2;
+  };
+
+  match (handler.handler)(PermissionKind::Geolocation) {
+    PermissionResponse::Allow => 0,
+    PermissionResponse::Deny => 1,
+    PermissionResponse::Default => 2,
+    PermissionResponse::Prompt => 3,
   }
 }
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -92,7 +92,7 @@ macro_rules! android_binding {
       $package,
       RustWebChromeClient,
       onPermissionRequestNative,
-      [jni::objects::JObjectArray],
+      [JString, jni::objects::JObjectArray],
       jint
     );
   }};
@@ -490,14 +490,24 @@ pub unsafe fn onPageLoaded(mut env: JNIEnv, _: JClass, webview_id: JString, url:
   }
 }
 
+#[allow(non_snake_case)]
 pub unsafe fn onPermissionRequestNative(
   mut env: JNIEnv,
   _: JClass,
+  webview_id: JString,
   resources: jni::objects::JObjectArray,
 ) -> jint {
   let mut allowed = false;
   let mut denied = false;
   let mut prompt = false;
+  let Ok(webview_id) = env.get_string(&webview_id) else {
+    return 2;
+  };
+  let webview_id = webview_id.to_str().ok().unwrap_or_default();
+  let permission_handlers = PERMISSION_HANDLER.lock().unwrap();
+  let Some(handler) = permission_handlers.get(webview_id) else {
+    return 2;
+  };
 
   if let Ok(size) = env.get_array_length(&resources) {
     for i in 0..size {
@@ -513,13 +523,11 @@ pub unsafe fn onPermissionRequestNative(
             _ => PermissionKind::Other,
           };
 
-          if let Some(handler) = &*PERMISSION_HANDLER.lock().unwrap() {
-            match (handler.handler)(kind) {
-              PermissionResponse::Allow => allowed = true,
-              PermissionResponse::Deny => denied = true,
-              PermissionResponse::Prompt => prompt = true,
-              PermissionResponse::Default => {}
-            }
+          match (handler.handler)(kind) {
+            PermissionResponse::Allow => allowed = true,
+            PermissionResponse::Deny => denied = true,
+            PermissionResponse::Prompt => prompt = true,
+            PermissionResponse::Default => {}
           }
         }
       }
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -93,7 +93,7 @@ macro_rules! android_binding {
       RustWebChromeClient,
       onPermissionRequestNative,
       [JString, jni::objects::JObjectArray],
-      jboolean
+      jint
     );
     android_fn!(
       $domain,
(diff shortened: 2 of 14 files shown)


src/wkwebview/class/wry_web_view_ui_delegate.rs
```diff
@@ -132,11 +133,30 @@ define_class!(
...
+      let decision = if let Some(handler) = &self.ivars().permission_handler {
+        match handler(permission_kind) {
+          PermissionResponse::Allow => WKPermissionDecision::Grant,
+          PermissionResponse::Deny => WKPermissionDecision::Deny,
+          PermissionResponse::Default => WKPermissionDecision::Grant, // Default to grant for backwards compatibility
+        }
+      } else {
+        // No handler set, default to grant (backwards compatible behavior)
+        WKPermissionDecision::Grant
+      };
```
pewsheen, line 163: I think we should align the default action with the OS, giving `WKPermissionDecision::Prompt`. Granting by default looks dangerous to me.

src/wkwebview/class/wry_web_view_ui_delegate.rs
```diff
@@ -132,11 +133,30 @@ define_class!(
...
+      capture_type: WKMediaCaptureType,
       decision_handler: &Block<dyn Fn(WKPermissionDecision)>,
     ) {
-      //https://developer.apple.com/documentation/webkit/wkpermissiondecision?language=objc
-      (*decision_handler).call((WKPermissionDecision::Grant,));
+      // Determine permission kind based on capture type
+      let permission_kind = match capture_type {
+        WKMediaCaptureType::Camera => PermissionKind::Camera,
+        WKMediaCaptureType::Microphone => PermissionKind::Microphone,
+        WKMediaCaptureType::CameraAndMicrophone => PermissionKind::Microphone, // Treat as microphone for now
```
pewsheen: Collapsing `CameraAndMicrophone` to `Microphone` also looks dangerous to me.

Maybe we can add some platform-specific type and document it well, or split it into something like `vec![Microphone, Camera]` and granted if and only if Microphone and Camera are both granted.

src/lib.rs
```diff
@@ -1260,6 +1323,37 @@ impl<'a> WebViewBuilder<'a> {
...
 
+  /// Set a handler to intercept permission requests from the webview.
+  ///
+  /// The handler receives the [`PermissionKind`] and should return
+  /// the desired [`PermissionResponse`].
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Windows**: Fully supported via WebView2's PermissionRequested event.
+  /// - **macOS / iOS / Linux / Android**: Not yet implemented, handler is ignored.
```
pewsheen: Looks like you have implemented most of them!

pewsheen (commented):
I currently only have a MacBook, so I'll share some thoughts about the macOS implementation here. For WebView2, maybe @Legend-Master can give some advice?

Personally, I tend not to loosen permission controls, to avoid them being used for malicious purposes.

src/lib.rs
```diff
@@ -535,6 +535,41 @@ pub struct NewWindowFeatures {
   pub opener: NewWindowOpener,
 }
 
+/// Permission types that can be requested by the webview.
+///
+/// See [`WebViewBuilder::with_permission_handler`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PermissionKind {
```
Legend-Master: Let's mark this as `#[non_exhaustive]` since we might support more in the future

Legend-Master (commented):
Looks good for the webview2 implementation at least at a glance, awesome work!

Could you also add a change file?

> https://github.com/tauri-apps/wry/blob/dev/.changes/readme.md

src/wkwebview/class/wry_web_view_ui_delegate.rs
Legend-Master: It seems to match the old behavior here? I'm not really aware of the historical context here though

pewsheen: Hmm, you're right. Let's leave it granted then. But I think we might  need a discussion for this, maybe in a new issue.

src/lib.rs
F0RLE: Done

src/wkwebview/class/wry_web_view_ui_delegate.rs
F0RLE: I've added a PermissionResponse::Prompt variant. This keeps the default behavior as Grant (for backward compatibility) but allows apps to explicitly opt-in to the system prompt if they want. I think this covers both use cases.

src/lib.rs
Legend-Master: Yeah, we should update this to match the docs on `permission_handler`

src/webview2/mod.rs
```diff
@@ -511,6 +511,53 @@ impl InnerWebView {
       }
     }
 
+    // Permission handler
+    if let Some(permission_handler) = attributes.permission_handler.take() {
+      let permission_handler = Rc::new(permission_handler);
+      unsafe {
+        webview.add_PermissionRequested(
```
Legend-Master, line 532: Maybe we could note that this will only fire for requests, that means if the user has allowed/denied the request before or you have used this API to do that, it will never fire again, I'm not exactly sure about other platforms though

src/webview2/mod.rs
```diff
@@ -511,6 +511,53 @@ impl InnerWebView {
       }
     }
 
+    // Permission handler
+    if let Some(permission_handler) = attributes.permission_handler.take() {
+      let permission_handler = Rc::new(permission_handler);
```
Legend-Master: Seem like this is not needed?

```suggestion
```

Legend-Master (dismissed):
The webview2 code looks good to me and works well when tested on my computer, just a few small nitpicks, you might need to wait for other members to give some feedbacks on other platforms though

Also, you'll need to sign your commits before we can merge them

src/wkwebview/class/wry_web_view_ui_delegate.rs
F0RLE: @pewsheen Confirming that the Camera/Microphone split has been implemented in macOS (WKWebView) so that PermissionKind::Camera and PermissionKind::Microphone are checked individually when WKMediaCaptureType::CameraAndMicrophone is requested. Also, the default response now favors system behavior.

src/lib.rs
F0RLE: @Legend-Master Confirming that the redundant Rc wrapper in src/webview2/mod.rs has been removed. I've also updated the doc comments for permission_handler to clarify that it only triggers for new requests and added Android support to the platform-specific list. A .changes file has been added to document these updates.

src/webview2/mod.rs
F0RLE: Thanks for the suggestion! I've updated the documentation in `src/lib.rs` (and in the Tauri core PR as well) to clarify that this handler only triggers for new requests and won't override persistent user settings. I also updated the backends and refined mappings based on other feedback.

src/wkwebview/class/wry_web_view_ui_delegate.rs
```diff
@@ -132,11 +133,51 @@ define_class!(
...
-      (*decision_handler).call((WKPermissionDecision::Grant,));
+      // Call user's permission handler if set
+      let decision = if let Some(handler) = &self.ivars().permission_handler {
+        match capture_type {
+          WKMediaCaptureType::Camera => match handler(PermissionKind::Camera) {
+            PermissionResponse::Allow => WKPermissionDecision::Grant,
+            PermissionResponse::Deny => WKPermissionDecision::Deny,
+            PermissionResponse::Default => WKPermissionDecision::Prompt,
+            PermissionResponse::Prompt => WKPermissionDecision::Prompt,
+          },
```
pewsheen:
```rust
match handler(PermissionKind::Camera) {
            PermissionResponse::Allow => WKPermissionDecision::Grant,
            PermissionResponse::Deny => WKPermissionDecision::Deny,
            PermissionResponse::Default => WKPermissionDecision::Prompt,
            PermissionResponse::Prompt => WKPermissionDecision::Prompt,
          },
```

This looks duplicated in branches, let's extract it to the local `fn` and call it in `Camera, Microphone and Other`

pewsheen (commented):
macOS implementation looks good to me. Only a nit that I think we can improve.

src/wkwebview/class/wry_web_view_ui_delegate.rs
F0RLE: Done! Extracted the mapping logic into a local `translate_response` closure in the latest commit.

src/wkwebview/class/wry_web_view_ui_delegate.rs
```diff
@@ -132,11 +133,43 @@ define_class!(
...
+
+            match (mic_res, cam_res) {
+              (PermissionResponse::Allow, PermissionResponse::Allow) => WKPermissionDecision::Grant,
+              (PermissionResponse::Deny, _) | (_, PermissionResponse::Deny) => {
+                WKPermissionDecision::Deny
+              }
+              (PermissionResponse::Prompt, _) | (_, PermissionResponse::Prompt) => {
+                WKPermissionDecision::Prompt
+              }
+              _ => WKPermissionDecision::Deny,
```
pewsheen: Oops, sorry. I just noticed that the default arm should be `WKPermissionDecision::Prompt`?

like (Allow, Default), (Default, Allow), (Default, Default) should go for Prompt

maybe we just need 
```rust
            match (mic_res, cam_res) {
              (PermissionResponse::Allow, PermissionResponse::Allow) => WKPermissionDecision::Grant,
              (PermissionResponse::Deny, _) | (_, PermissionResponse::Deny) => {
                WKPermissionDecision::Deny
              }
              _ => WKPermissionDecision::Prompt,
            }
```

examples/permission_handler.rs
```diff
@@ -0,0 +1,174 @@
...
+      .with_title("Permission Handler Example")
+      .with_inner_size(tao::dpi::LogicalSize::new(700, 500))
+      .build(&event_loop)
+      .unwrap();
+
+    // Register a custom protocol so the page runs in a trusted origin,
+    // which enables navigator.mediaDevices and other permission-gated APIs.
+    let html_bytes = HTML.as_bytes().to_vec();
+    let builder = WebViewBuilder::new()
+      .with_asynchronous_custom_protocol("perm".into(), move |_id, _req, responder| {
```
Legend-Master: Can we use a site like https://permission.site/ instead here?

src/lib.rs
```diff
@@ -535,6 +535,116 @@ pub struct NewWindowFeatures {
...
+  /// Clipboard read permission.
+  ClipboardRead,
+  /// Display capture permission (for getDisplayMedia).
+  DisplayCapture,
+  /// Midi access permission.
+  Midi,
+  /// Nfc access permission.
+  ///
+  /// ## Platform-specific
+  /// - **Windows / macOS / Linux / Android / iOS**: Not yet supported by platform backends.
```
Legend-Master: Not supported on any platform?

Legend-Master (commented):
And you'll need to sign your commits

src/wkwebview/class/wry_web_view_ui_delegate.rs
F0RLE: Done. Simplified the match logic to default to Prompt for combined requests as suggested. Also squashed all changes into a single signed commit.

examples/permission_handler.rs
F0RLE: Updated the example to use permission.site and simplified the implementation.

src/lib.rs
F0RLE: Applied. Updated NFC and Bluetooth support notices to 'All platforms: Not yet supported' for clarity.

F0RLE:
> The webview2 code looks good to me and works well when tested on my computer, just a few small nitpicks, you might need to wait for other members to give some feedbacks on other platforms though
> 
> Also, you'll need to sign your commits before we can merge

Done. The commit is now digitally signed with an SSH key.

src/lib.rs
```diff
@@ -535,6 +535,116 @@ pub struct NewWindowFeatures {
...
+  /// Nfc access permission.
+  ///
+  /// ## Platform-specific
+  /// - **All platforms**: Not yet supported.
+  Nfc,
+  /// Bluetooth access permission.
+  ///
+  /// ## Platform-specific
+  ///   - **All platforms**: Not yet supported by platform backends.
+  Bluetooth,
```
Legend-Master: We should remove these or comment them out if they're not supported at all

src/lib.rs
```diff
@@ -796,6 +906,37 @@ pub struct WebViewAttributes<'a> {
...
+  /// > This handler only triggers for new permission requests. If the user has already
+  /// > allowed or denied a permission persistently within the webview, the browser
+  /// > will use the saved preference instead of calling this handler.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Windows**: Fully supported via WebView2's PermissionRequested event.
+  /// - **macOS / iOS**: Fully supported via WKUIDelegate's requestMediaCapturePermission.
+  /// - **Linux**: Fully supported via WebKitGTK's permission-request signal.
+  /// - **Android**: Experimental support implemented via JNI.
```
Legend-Master: Could you explain a bit about what does 'experimental support' mean here?

pewsheen (dismissed):
`wkwebview` looks good to me 👍 . Let's see if there's something that needs to be discussed in other platform.

src/lib.rs
```diff
@@ -1044,7 +1174,7 @@ impl<'a> WebViewBuilder<'a> {
   /// # Reading assets on mobile
   ///
   /// - Android: For loading content from the `assets` folder (which is copied to the Andorid apk) please
-  ///   use the function [`with_asset_loader`] from [`WebViewBuilderExtAndroid`] instead.
+  ///   use the function `with_asset_loader` from `WebViewBuilderExtAndroid` instead.
```
Legend-Master: Please revert unrelated changes

Legend-Master (commented):
Looking good for the webview2 part and general APIs, I'll take a look at the webkitgtk implementation later when I got time. The Android implementation will need someone who's more familiar with the mobile part to review though

Thanks for the patience and awesome work!

Also please don't force push to a single commit on every change, it makes it hard to follow what's changed between each reviews, we will squash merge at the end

Legend-Master (dismissed):
I have gone ahead and did some adjustments, hope you don't mind

The Windows and Linux implementations are looking good to me, waiting for a reviewer who's more familiar with the mobile implementations to take a final look before merging

F0RLE:
Thanks for the review and for the adjustments as well! They look good. Let me know if there's anything else for the Android or webkitgtk implementation.

src/webkitgtk/mod.rs
```diff
@@ -573,6 +577,51 @@ impl InnerWebView {
...
+    if let Some(permission_handler) = attributes.permission_handler.take() {
+      webview.connect_permission_request(move |_webview, request| {
+        // Determine permission kind
+        let permission_kind =
+          if let Some(media_request) = request.downcast_ref::<UserMediaPermissionRequest>() {
+            if media_request.is_for_audio_device() {
+              PermissionKind::Microphone
+            } else if media_request.is_for_video_device() {
+              PermissionKind::Camera
+            } else {
```
FabianLars: can the same request ask for mic and camera permissions at the same time like on macos? if so this would always only ask for audio permissions

src/webkitgtk/mod.rs
```diff
@@ -573,6 +577,51 @@ impl InnerWebView {
...
+              // On WebKitGTK < 2.42, there's no is_for_display_device()
```
FabianLars: is there something for >2.42 then? can't find it right now but maybe it's not in the docs because it requries building webkitgtk from source iirc?

don't have access to a linux system rn

src/wkwebview/class/wry_web_view_ui_delegate.rs
```diff
@@ -132,11 +133,40 @@ define_class!(
...
+            let cam_res = handler(PermissionKind::Camera);
+
+            match (mic_res, cam_res) {
+              (PermissionResponse::Allow, PermissionResponse::Allow) => WKPermissionDecision::Grant,
+              (PermissionResponse::Deny, _) | (_, PermissionResponse::Deny) => {
+                WKPermissionDecision::Deny
+              }
+              _ => WKPermissionDecision::Prompt,
+            }
+          }
```
FabianLars, line 158: related to my linux comment, i wonder if it makes sense to abstract over this shared permission type or whether we should just have PermissionKind::CameraAndMicrophone as well cause it exists on 3 out of 5 platforms 🤔

F0RLE (commented):
Thanks for the review @FabianLars! Addressed your comments in the latest commit:

1. **Combined audio+video on webkitgtk** — fixed. Now each permission (mic/camera) is checked individually via `is_for_audio_device()` / `is_for_video_device()`. If either is denied, the whole request is denied.

2. **WebKitGTK >2.42** — `is_for_display_device()` is gated behind the `v2_42` feature flag. For older versions we fall back to `!is_audio && !is_video` as a heuristic for display capture.

3. **CameraAndMicrophone** — on both webkitgtk and macOS/iOS the combined request is now handled by checking `PermissionKind::Microphone` and `PermissionKind::Camera` separately, so the handler can grant one and deny the other independently. I think keeping them separate is cleaner than adding a combined variant.

src/wkwebview/class/wry_web_view_ui_delegate.rs
F0RLE: Good point! I checked the platform docs more carefully:

- **WebKitGTK**: A single `UserMediaPermissionRequest` can have both `is_for_audio_device` and `is_for_video_device` return `true` simultaneously (e.g. `getUserMedia({ audio: true, video: true })`). The current code handles this by checking each individually with deny wins.

- **macOS/iOS**: `WKMediaCaptureType.CameraAndMicrophone` is a single enum variant — one request, one decision. We currently split it into two handler calls which is a bit awkward.

I think keeping `PermissionKind` separate is still the right call because:
1. Most pages request camera or mic individually, not combined
2. The handler can grant one and deny the other
3. For combined requests, "deny wins" is a sensible default

If someone needs explicit control over the combined case, we could add a `CameraAndMicrophone` variant later, but I think the current approach covers the common use cases well enough.

F0RLE:
Updated this PR against latest `dev` and resolved the Android conflicts.

The Android permission handling is still scoped per webview, and the checks are green again. I also updated the linked Tauri PR to use the new Wry commit.

@FabianLars could you take another look at the Android part when you have time?

Legend-Master:
Sorry for the force push, I couldn't figure out how to fix the previous bugged merge without it

src/android/kotlin/RustWebChromeClient.kt
```diff
@@ -21,53 +21,13 @@ import android.provider.MediaStore
...
-
-  private val activity: WryActivity
-  private var permissionLauncher: ActivityResultLauncher<Array<String>>
-  private var activityLauncher: ActivityResultLauncher<Intent>
-  private var permissionListener: PermissionListener? = null
-  private var activityListener: ActivityResultListener? = null
-
-  init {
-    activity = appActivity
-    val permissionCallback =
```
Legend-Master, line 51: Looks like you have moved this to `WryActivity`, could you explain the reason for the move?

F0RLE: Yeah, this is mostly because of Android lifecycle rules.

The permission launcher has to be registered from the activity early enough. `RustWebChromeClient` is created later when the webview is created, so registering it there can be too late.

`WryActivity` now only handles the Android runtime permission prompt, while `RustWebChromeClient` still keeps the webview-specific logic and passes the correct `webViewId` to the native handler.

src/android/kotlin/RustWebChromeClient.kt
```diff
@@ -92,32 +52,74 @@ class RustWebChromeClient(appActivity: WryActivity) : WebChromeClient() {
...
-      permissionList.add(Manifest.permission.MODIFY_AUDIO_SETTINGS)
-      permissionList.add(Manifest.permission.RECORD_AUDIO)
+
+    val grantedResources = ArrayList<String>()
+    val defaultResources = ArrayList<String>()
+
+    for (resource in requestedResources) {
+      when (onPermissionRequestNative(webViewId, arrayOf(resource))) {
+        0 -> grantedResources.add(resource)
+        1 -> {
```
Legend-Master: Looks like we can only prompt or deny on Android right? Then maybe we can simplify the options here, just a bool crossing the JNI, indicating if it's denied or should go through the normal prompt

F0RLE: Yeah, that makes sense.

Android still needs to go through the runtime permission flow for camera/mic/location, so treating `Allow`, `Default`, and `Prompt` as "continue with the normal Android prompt" is probably clearer here.

I'll simplify the JNI return value to just indicate whether the request should be denied or continue normally.

F0RLE: Done in cced03cb. Android now only sends a boolean over JNI: true means deny, false means continue with the normal Android permission flow. `Allow`, `Default`, and `Prompt` all fall through to the normal prompt path.

src/android/kotlin/RustWebChromeClient.kt
```diff
@@ -92,32 +52,59 @@ class RustWebChromeClient(appActivity: WryActivity) : WebChromeClient() {
...
+    if (requestedResources.isEmpty()) {
+      request.deny()
+      return
     }
-    if (listOf(*request.resources).contains("android.webkit.resource.AUDIO_CAPTURE")) {
-      permissionList.add(Manifest.permission.MODIFY_AUDIO_SETTINGS)
-      permissionList.add(Manifest.permission.RECORD_AUDIO)
+
+    for (resource in requestedResources) {
+      if (onPermissionRequestNative(webViewId, arrayOf(resource))) {
```
Legend-Master: This is a bit weird that we are iterating through an array of `requestedResources` and then wrapping, sending each resource in an array to the rust side

src/android/kotlin/RustWebChromeClient.kt
```diff
@@ -92,32 +52,59 @@ class RustWebChromeClient(appActivity: WryActivity) : WebChromeClient() {
...
+      permissionList.add(Manifest.permission.CAMERA)
     }
+    if (resources.contains(PermissionRequest.RESOURCE_AUDIO_CAPTURE)) {
+      permissionList.add(Manifest.permission.MODIFY_AUDIO_SETTINGS)
+      permissionList.add(Manifest.permission.RECORD_AUDIO)
+    }
+    return permissionList
   }
 
+  private external fun onPermissionRequestNative(webviewId: String, resources: Array<String>): Boolean
```
Legend-Master: let's document what the return value means, also the rust side

src/permissions.rs
```diff
@@ -0,0 +1,166 @@
...
+  /// Deny the permission.
+  Deny,
+  /// Use default behavior (show system prompt).
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Windows / macOS / Android**: The default behavior is [`Self::Prompt`]
+  /// - **Linux**: The default behavior is [`Self::Deny`]
+  #[default]
+  Default,
```
Legend-Master, line 150: Looking at our current implementations, maybe we can remove this `Default` variant added because of the macOS defaults

> https://github.com/tauri-apps/wry/pull/1654#discussion_r2754950042

(The default is `WKPermissionDecision::Grant`, but if you attached a `permission_handler`, you're responsible to pick one)

Legend-Master: Or we can remove `Prompt` if you prefer, since on

- Windows, you only get `CoreWebView2PermissionState::Default`
  > Specifies that the default browser behavior is used, which normally prompt users for decision.
  > https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2permissionstate?view=webview2-dotnet-1.0.3912.50
- Linux you get `Deny`
- macOS you get `Prompt`
- Android you get `Prompt`

src/android/kotlin/RustWebChromeClient.kt
F0RLE: Done. Kotlin now passes the full resource array to JNI once, and Rust checks all requested resources in that single call.

src/android/kotlin/RustWebChromeClient.kt
F0RLE: Done. I added comments on both sides: true means deny the request, false means continue with the normal Android permission flow.

src/permissions.rs
F0RLE: Done. I removed Prompt and kept Default. Default now represents falling back to the platform/browser default behavior, which maps better across WebView2, macOS, Linux, and Android than promising an explicit prompt everywhere.

src/android/kotlin/RustWebChromeClient.kt
```diff
@@ -240,35 +227,61 @@ class RustWebChromeClient(appActivity: WryActivity) : WebChromeClient() {
...
       }
-      permissionLauncher.launch(geoPermissions)
     } else {
       // permission is already granted
       callback.invoke(origin, true, false)
       Logger.debug("onGeolocationPermissionsShowPrompt: has required permission")
     }
   }
 
+  private fun safePermissionRequestResources(resources: Array<String>): Array<String> {
```
Legend-Master: Since we have `PermissionKind::Other`, do we want to filter out unrecognized permission requests here?

src/android/binding.rs
```diff
@@ -485,3 +501,73 @@ pub unsafe fn onPageLoaded(mut env: JNIEnv, _: JClass, webview_id: JString, url:
...
+
+#[allow(non_snake_case)]
+pub unsafe fn onGeolocationPermissionRequestNative(
+  mut env: JNIEnv,
+  _: JClass,
+  webview_id: JString,
+  _origin: JString,
+) -> jboolean {
+  // Return true to deny geolocation. false lets Kotlin continue with Android's
+  // normal runtime permission flow.
```
Legend-Master: Let's move this to above the function as doc comments

src/android/binding.rs
```diff
@@ -485,3 +501,73 @@ pub unsafe fn onPageLoaded(mut env: JNIEnv, _: JClass, webview_id: JString, url:
...
+
+#[allow(non_snake_case)]
+pub unsafe fn onPermissionRequestNative(
+  mut env: JNIEnv,
+  _: JClass,
+  webview_id: JString,
+  resources: jni::objects::JObjectArray,
+) -> jboolean {
+  // Return true to deny the Android request. false lets Kotlin continue with
+  // Android's normal runtime permission flow.
```
Legend-Master: Let's move this to above the function as doc comments

src/android/kotlin/RustWebChromeClient.kt
F0RLE: Done. Kotlin now passes the full requested resource array to Rust, so unknown Android resources can reach the handler as PermissionKind::Other. The Android grant path still only grants resources Wry explicitly supports; if nothing is grantable, it denies.

src/android/binding.rs
F0RLE: Done. I moved the return-value note above the JNI function as a doc comment.

src/android/binding.rs
F0RLE: Done. I moved the return-value note above the JNI function as a doc comment.

src/android/kotlin/RustWebChromeClient.kt
Legend-Master: Looks like dc2766c made so even If we allow `PermissionKind::Other`, it would still be denied?

F0RLE: You're right, that made Other visible to the handler but still impossible to actually allow. I changed the Android JNI path to return a small decision code instead of a boolean: deny wins if any resource is denied, explicit allow grants the full requested resource list, and default keeps the safe filtered Android flow. So PermissionKind::Other can now be allowed explicitly, while Default still does not grant unknown resources.

Legend-Master: Pushed e7c84dc to add partial allow support, allow goes through, default only allows known permissions

Legend-Master (approved):
Thanks again for the incredible patience and awesome work!

PR #1655, opened 2026-01-16
feat: Derive PartialEq in Rect

marc2332:
Derive PartialEq in Rect :)

### DIFF
--- a/.changes/partialeq-rect.md
+++ b/.changes/partialeq-rect.md
@@ -1,5 +1,5 @@
 ---
-'wry': minor
+'wry': patch
 ---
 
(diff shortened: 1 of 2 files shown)


.changes/partialeq-rect.md
```diff
@@ -0,0 +1,5 @@
+---
+'wry': minor
```
FabianLars:
```suggestion
'wry': patch
```

Should be non-breaking, right? Since we're 0.x we can do a patch then.

marc2332: Yeah, fair

FabianLars (approved):

Issue #583, opened 2022-03-04
Deadlock creating a new window on the IPC handler on Windows


lucasfernog:
**Describe the bug**
On Tauri, we allow users to create a window from a IPC handler. This results in a deadlock on Windows on the `create_controller` function (the second closure is never called). This might be an issue on the windows crate instead.

**Steps To Reproduce**
I've changed the multi_window example to reproduce this issue (note that I'm not using the event loop proxy to send a message because Tauri needs a response from the event loop (the window id).

```rust
// Copyright 2019-2021 Tauri Programme within The Commons Conservancy
// SPDX-License-Identifier: Apache-2.0
// SPDX-License-Identifier: MIT

fn main() -> wry::Result<()> {
  use std::collections::HashMap;
  use std::ops::Deref;
  use wry::{
    application::{
      event::{Event, WindowEvent},
      event_loop::{ControlFlow, EventLoop, EventLoopProxy, EventLoopWindowTarget},
      window::{Window, WindowBuilder, WindowId},
    },
    webview::{WebView, WebViewBuilder},
  };

  #[derive(Clone)]
  enum UserEvents {
    CloseWindow(WindowId),
    NewWindow(),
  }

  fn create_new_window(
    title: String,
    event_loop: EventLoopWindowTarget<UserEvents>,
    proxy: EventLoopProxy<UserEvents>,
  ) -> (WindowId, WebView) {
    let window = WindowBuilder::new()
      .with_title(title)
      .build(&event_loop)
      .unwrap();
    let window_id = window.id();
    let event_loop_ = event_loop.clone();
    let handler = move |window: &Window, req: String| match req.as_str() {
      "new-window" => {
        //let _ = proxy.send_event(UserEvents::NewWindow());
        create_new_window("something".into(), event_loop_.clone(), proxy.clone());
      }
      "close" => {
        let _ = proxy.send_event(UserEvents::CloseWindow(window.id()));
      }
      _ if req.starts_with("change-title") => {
        let title = req.replace("change-title:", "");
        window.set_title(title.as_str());
      }
      _ => {}
    };

    let webview = WebViewBuilder::new(window)
      .unwrap()
      .with_html(
        r#"
          <button onclick="window.ipc.postMessage('new-window')">Open a new window</button>
          <button onclick="window.ipc.postMessage('close')">Close current window</button>
          <input oninput="window.ipc.postMessage(`change-title:${this.value}`)" />
      "#,
      )
      .unwrap()
[...]

PR #1657, opened 2026-01-21
fix!: remove Send, Sync bounds from closure to fix tauri issue 14801

sftse:
The Send and Sync bounds on the closure are not used and prevent the most straightforward fix to [tauri issue 14801](https://github.com/tauri-apps/tauri/issues/14801) that I can come up with.

I'm basing my PR on heuristics, and have no idea if the removal of these bounds is going to affect some changes down the line.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -700,8 +700,7 @@ pub struct WebViewAttributes<'a> {
   /// - **Windows**: The closure is executed on a separate thread to prevent a deadlock.
   ///
   /// [window.open]: https://developer.mozilla.org/en-US/docs/Web/API/Window/open
-  pub new_window_req_handler:
-    Option<Box<dyn Fn(String, NewWindowFeatures) -> NewWindowResponse + Send + Sync>>,
+  pub new_window_req_handler: Option<Box<dyn Fn(String, NewWindowFeatures) -> NewWindowResponse>>,
 
   /// Enables clipboard access for the page rendered on **Linux** and **Windows**.
   ///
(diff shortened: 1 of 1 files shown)


Legend-Master:
@lucasfernog `new_window_req_handler`'s comment says the callback is executed in a separate thread to prevent deadlock on Windows, could you tell where exactly does it happen? The current unsafe Send Sync is not really correct, os webviews returned from the callback can't be `Send + Sync`

FabianLars:
i assume it's the same thing that requires WebviewWindows to be build on separate things, no?

Legend-Master:
> i assume it's the same thing that requires WebviewWindows to be build on separate things, no?

That's what I assume as well but I never really understood that one either... The description/comment isn't very clear about it so

FabianLars:
yeah i don't know/remember either. i think i was never able to repro this using just wry tho

Legend-Master:
> https://github.com/tauri-apps/wry/issues/583#issuecomment-1713822901
> https://github.com/tauri-apps/wry/issues/583#issuecomment-3186158240
> https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/threading-model#reentrancy

I think I understand now, we can not create new webviews inside any of the webview2 callbacks, so we are calling `new_window_req_handler` in another thread, letting the callback to complete, then on the `new_window_req_handler` thread, we send a message to the main thread requesting a new window, and when it completes, it sends the handle back to the `new_window_req_handler` thread to complete the deferred event handler

I think this is something we should document about in the code

FabianLars:
is this something we can abstract away? i guess that's tricky cause we can't block on the mainthread 🤔

Legend-Master:
I'm not sure if this will work or is it a good idea:

- We call defer inside the callback (`args.GetDeferral`, like we're currently doing), then message the current event loop to schedule a function
- callback completes
- On the next message (or whenever we get the message), run the function we scheduled inside the callback

PR #1660, opened 2026-01-31
fix(windows): run `new_window_req_handler` on main thread

Legend-Master:
Implement the solution in https://github.com/tauri-apps/wry/pull/1657#issuecomment-3823336202, closes #1657

Run `new_window_req_handler` on main thread and remove the `Send + Sync` restriction on it

Also fixed:

- `NewWindowResponse::Create` not working if no custom protocol's registered
- `tauri-runtime-wry`'s usage of `RefCell` on another thread pointed out by @sftse in https://github.com/tauri-apps/tauri/pull/14805

### DIFF
--- a/src/webview2/mod.rs
+++ b/src/webview2/mod.rs
@@ -689,7 +689,7 @@ impl InnerWebView {
     let new_window_req_handler = attributes
       .new_window_req_handler
       .take()
-      .map(std::sync::Arc::new);
+      .map(std::rc::Rc::new);
     let env_ = env.clone();
     // New window handler
     webview.add_NewWindowRequested(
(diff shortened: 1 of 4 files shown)


sftse:
Thanks for taking this on.

Because the field is `pub`, the change is breaking.

```rust
pub struct Foo {
    pub f: Box<dyn Fn() + Send + Sync>, // remove Send + Sync breaks the code
}

fn main() {
    let f = move || {};
    let foo = Foo { f: Box::new(f) };
    std::thread::scope(|s| {
        s.spawn(|| &foo);
    });
}
```

src/webview2/mod.rs
```diff
@@ -1137,6 +1135,13 @@ impl InnerWebView {
     env.CreateWebResourceResponse(None, status_code as i32, &status, &error)
   }
 
+  /// Send `function` to run on `hwnd`'s thread
+  ///
+  /// ## SAFETY:
+  ///
+  /// This function doesn't force a `Send` to make it easier to use,
+  /// the caller must call this function on the same thread as `hwnd`
+  /// or ensure the function is safe to `Send`
```
sftse: I don't know enough about the code to judge this, but I just briefly sanity-checked whether the function passed is ever Send.
All the call sites pass a function that is not Send, so it might be sensible to drop the second precondition even if it were correct. Judging a more complex precondition is more difficult than a simple one. It also seems like a precondition that would warrant a comment referencing external documentation on `PostMessageW`, since this is the only unsafe function called in this block.

Legend-Master: Not sure what's the better words here 🤔

sftse: Would this be incorrect or incomplete? The closures passed are already not Send.
```suggestion
  /// Send `function` to run on `hwnd`'s thread
  ///
  /// ## SAFETY:
  ///
  /// The caller must call this function on the same thread as `hwnd`.
```

Legend-Master: It's quite tricky here, our custom protocol handler is `!Send` but marked with `unsafe impl Send` to send to another thread for async tasks, and then send back and called through `dispatch_handler`, so technically, it is only accessed on the `hwnd`'s thread but it is not called on `hwnd`'s thread

FabianLars (approved):

Issue #463, opened 2022-01-04
wry system_tray and system_tray_no_menu panic exit on windows 10


davehorner:
**Describe the bug**
Running the system_tray examples, the system tray shows up; when clicked, a window flashes and the application exits with console errors.  Other examples are working fine.

**Steps To Reproduce**
> cargo run --example system_tray
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target\debug\examples\system_tray.exe`
Wry has started!
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\webview\webview2\mod.rs:235:16
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
webview2: panic in callback function. Aborting because it's UB to unwind across FFI boundaries.
error: process didn't exit successfully: `target\debug\examples\system_tray.exe` (exit code: 255)

> cargo run --example system_tray_no_menu
    Finished dev [unoptimized + debuginfo] target(s) in 3.95s
     Running `target\debug\examples\system_tray_no_menu.exe`
Wry has started!
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\webview\webview2\mod.rs:235:16
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
webview2: panic in callback function. Aborting because it's UB to unwind across FFI boundaries.
error: process didn't exit successfully: `target\debug\examples\system_tray_no_menu.exe` (exit code: 255)

**Expected behavior**
both programs exit with a panic.  expect it to work and keep running.

**Platform and Versions (please complete the following information):**
OS: windows 10
Rustc:  rustc --version
rustc 1.56.1 (59eed8a2a 2021-11-01)

**Would you assign yourself to resolve this bug?**
- [ ] Yes
- [ X] No - do not know how.

**Additional context**
The other examples like multi_window work fine so I know that the webview2 component is working

amrbashir:
you rust is outdated, can you try upgrading to 1.57 ? we had some reports about bugs with 1.56 but got fixed in 1.57

davehorner:
>rustc --version
rustc 1.57.0 (f1edd0429 2021-11-29)

the behavior is the same.  the window may stay open a little longer.  The no_menu displayed a window that resized but only painted black in the newly surfaced window portions.

amrbashir:
It is working fine on my end, can you try getting a backtrace ?

davehorner:
```Wry has started!
[...]

PR #286, opened 2021-06-04
Add WebViewAttributes

wusyong:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Documentation
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes. Issue #___
- [x] No

**The PR fulfills these requirements:**

- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

PR #1662, opened 2026-02-05
feat(macos): implement background_color for WKWebView

JSKitty:
## Summary

Implements `background_color` / `with_background_color` / `set_background_color` on **macOS** — the only remaining platform where this was documented as "Not implemented".

- Disables the default white WKWebView background via the `drawsBackground` KVC key on `WKWebViewConfiguration` (pre-init), the same key already used by the `transparent` feature
- Sets `underPageBackgroundColor` (public API, macOS 12+) on the `WKWebView` instance for overscroll/loading areas
- Updates all three doc comments to reflect macOS support

## Context

The `backgroundColor` config in Tauri's `tauri.conf.json` correctly sets the **NSWindow** background on macOS, and the **webview** background on Windows (WebView2), Linux (WebKitGTK), iOS, and Android. However, wry's macOS codepath was a no-op — the WKWebView layer on top of the window retained its default white background, causing a visible white flash before HTML/CSS paints.

This has been a long-standing issue affecting dark-themed apps that use a hidden-then-shown window pattern to prevent flash on startup. The fix was deferred in the original PR #682 because no maintainer had macOS hardware to test on. The [Vector Privacy](https://github.com/VectorPrivacy/Vector) team would like to change that — we're committed to assisting fully with the review and merging of this PR, for the benefit of Vector as well as all other Tauri v2 apps on macOS that deserve a standardised way of preventing "the unholy flashbang".

## Real-World Tests (on a [real Tauri v2 app](https://github.com/VectorPrivacy/Vector))

**Before (With Flashbang!)**:
<video src="https://github.com/user-attachments/assets/77f9e15f-30d1-4ee8-8150-c6deac8a4e50" autoplay loop muted></video>

**After (No Flashbang!)**:
<video src="https://github.com/user-attachments/assets/e6b28580-6de8-4c23-b559-3580a91dcf8c" autoplay loop muted></video>

## Approach: `drawsBackground` KVC key

There is **no public Apple API** to disable WKWebView's default white background on macOS (`NSView` doesn't have a writable `opaque` property like iOS's `UIView`). The `drawsBackground` KVC key on `WKWebViewConfiguration` is the standard approach — and is already used by wry's own [`transparent` feature](https://github.com/tauri-apps/wry/blob/dev/src/wkwebview/mod.rs#L354-L359).

We investigated alternatives:
- `setUnderPageBackgroundColor` alone — does not prevent the initial white flash (only affects overscroll areas)
- `setOpaque(false)` — not available on macOS `NSView` (iOS-only setter)
- JS-side `setTimeout` delays before `.show()` — device-dependent, unreliable across hardware

### Feature-gating

If the maintainers prefer extra caution, this could be gated behind the existing `transparent` or `macos-private-api` feature flag — though we'd argue it should be ungated since `background_color` is a standard webview configuration, and the same KVC key is already used by the `transparent` feature on macOS.

## Research: safety of the `drawsBackground` KVC key

We conducted extensive research into the App Store safety of this key. Findings:

**Zero documented App Store rejections.** After searching Apple Developer Forums ([thread/121139](https://developer.apple.com/forums/thread/121139), [thread/65553](https://developer.apple.com/forums/thread/65553), [thread/87474](https://developer.apple.com/forums/thread/87474)), GitHub issues across wry, Tauri, WebKit, react-native-webview, Sparkle, and the WebKit bug tracker ([bug 200528](https://bugs.webkit.org/show_bug.cgi?id=200528)) — not a single report documents an actual Mac App Store rejection caused by `drawsBackground`.

**KVC string keys vs private selectors.** Apple's App Store scanner works by extracting Objective-C selectors from the binary's `__objc_selrefs` section and matching them against a private API database. When `drawsBackground` is accessed via `setValue:forKey:` (a public KVC API), the binary only contains the public selector `setValue:forKey:` and a string literal — **not** the private selector `_setDrawsBackground:`. Documented App Store rejections ([Electron #20027](https://github.com/electron/electron/issues/20027), [react-native-webview #2527](https://github.com/react-native-webview/react-native-webview/issues/2527), [cordova-plugin-ionic-webview #286](https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/286)) all flagged direct private selectors or class names in `__objc_selrefs`/`__objc_classrefs`, never KVC string keys.

**~9 years of stability.** The `drawsBackground` property has been present in WebKit since at least macOS 10.12 Sierra (2016) through macOS 15 Sequoia (2025), declared in `WKWebViewPrivate.h` and `WKWebViewConfigurationPrivate.h`. It has never been removed or broken across releases.

**Apple is aware of the gap.** Feedback [FB7539179](https://github.com/feedback-assistant/reports/issues/81) (filed January 2020) explicitly requested Apple make transparent WKWebView backgrounds a public API on macOS, noting developers are forced to use a "semi-private property." Apple has not made it public, but has also not broken or flagged it.

**No public alternative exists.** Apple added `underPageBackgroundColor` (macOS 12+) as a public API, but it only controls the color visible when scrolling past page bounds — it does not replace `drawsBackground` for preventing the white background from rendering.

**wry's existing warnings are precautionary.** The documentation at `src/lib.rs` lines 309-312 warns to "avoid this in release build if your app needs to publish to App Store" for the `transparent` feature. This warning cites no actual rejection event and is precautionary in nature ([tauri-docs #463](https://github.com/tauri-apps/tauri-docs/issues/463) likewise contains no evidence of rejections).

## Test plan

- [x] Tested on macOS (Apple Silicon) with [a Tauri v2 app](https://github.com/VectorPrivacy/Vector) using `backgroundColor: "#030303"` + `visible: false` + show-on-content-ready pattern
- [x] Verified **zero white flash** on startup in release builds — no JS-side delays needed
- [x] Verified overscroll areas show the correct dark color via `underPageBackgroundColor`
- [x] Confirmed the fix works with both init-time (`with_background_color`) and runtime (`set_background_color`) paths
- [ ] Windows/Linux builds should be unaffected (no changes to those codepaths)

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -352,19 +352,9 @@ impl InnerWebView {
       }
 
       #[cfg(feature = "transparent")]
-      if attributes.transparent {
+      if attributes.transparent || attributes.background_color.is_some() {
         let no = NSNumber::numberWithBool(false);
         // Equivalent Obj-C:
-        config.setValue_forKey(Some(&no), ns_string!("drawsBackground"));
-      }
-
-      // On macOS, disable the default white WKWebView background when a custom
-      // background color is specified. Applied to the config before initWithFrame
-      // so the WKWebView never renders a white frame.
-      // Uses the same `drawsBackground` KVC key as the `transparent` feature above.
-      #[cfg(target_os = "macos")]
-      if attributes.background_color.is_some() {
-        let no = NSNumber::numberWithBool(false);
         config.setValue_forKey(Some(&no), ns_string!("drawsBackground"));
       }
 
@@ -929,7 +919,7 @@ r#"Object.defineProperty(window, 'ipc', {
       self.webview.setBackgroundColor(Some(&color));
     }
 
-    #[cfg(target_os = "macos")]
+    #[cfg(all(target_os = "macos", feature = "transparent"))]
     unsafe {
       let (red, green, blue, alpha) = _background_color;
 
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -358,6 +358,16 @@ impl InnerWebView {
         config.setValue_forKey(Some(&no), ns_string!("drawsBackground"));
       }
 
+      // On macOS, disable the default white WKWebView background when a custom
+      // background color is specified. Applied to the config before initWithFrame
+      // so the WKWebView never renders a white frame.
+      // Uses the same `drawsBackground` KVC key as the `transparent` feature above.
+      #[cfg(target_os = "macos")]
+      if attributes.background_color.is_some() {
+        let no = NSNumber::numberWithBool(false);
+        config.setValue_forKey(Some(&no), ns_string!("drawsBackground"));
+      }
+
       #[cfg(feature = "fullscreen")]
       // Equivalent Obj-C:
       _preference.setValue_forKey(Some(&_yes), ns_string!("fullScreenEnabled"));
@@ -399,6 +409,20 @@ impl InnerWebView {
         };
         let webview: Retained<WryWebView> =
           objc2::msg_send![super(webview), initWithFrame: frame, configuration: &**config];
+
+        // Set the under-page background color for overscroll areas (public API, macOS 12+).
+        // drawsBackground is already disabled on the config above, so the window background
+        // shows through. This handles the color visible when scrolling past page bounds.
+        if let Some((red, green, blue, alpha)) = attributes.background_color {
+          let color = objc2_app_kit::NSColor::colorWithSRGBRed_green_blue_alpha(
+            red as f64 / 255.0,
+            green as f64 / 255.0,
+            blue as f64 / 255.0,
+            alpha as f64 / 255.0,
+          );
+          webview.setUnderPageBackgroundColor(Some(&color));
+        }
+
         webview
       };
       #[cfg(target_os = "ios")]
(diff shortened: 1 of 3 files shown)


src/wkwebview/mod.rs
```diff
@@ -358,6 +358,16 @@ impl InnerWebView {
         config.setValue_forKey(Some(&no), ns_string!("drawsBackground"));
       }
 
+      // On macOS, disable the default white WKWebView background when a custom
+      // background color is specified. Applied to the config before initWithFrame
+      // so the WKWebView never renders a white frame.
+      // Uses the same `drawsBackground` KVC key as the `transparent` feature above.
+      #[cfg(target_os = "macos")]
+      if attributes.background_color.is_some() {
```
FabianLars: you can merge this into the transparent `if` above. since it _is_ a private api it should be behind a feature flag. having one use of it behind a flag and the other not does not make much sense. Of course we could remove the flag since nobody ever reported issues with it, but we all know that the reports will come in as soon as we do 😂

JSKitty: I had a feeling you were gonna say similar, fine! 😛

pewsheen (commented):
Thanks for the PR. My only concern here is that Tauri has minimal support for macOS 10.15+.
I'm not sure if it will crash when calling `setUnderPageBackgroundColor` under macOS 12.

Maybe we should set a version check like how datastore did and mention in document that only macOS12+ will work:
https://github.com/tauri-apps/wry/blob/9fa99b14fd8fb8aa739a17c6bb6e074c858da0c8/src/wkwebview/mod.rs#L224

JSKitty:
That sounds reasonable to me. I can adjust accordingly?

pewsheen:
> That sounds reasonable to me. I can adjust accordingly?

Sure!

JSKitty:
Gated, re-tested `Wry` in my Production App on macOS 26.2 (Tahoe) with the expected functionality, I do not have an *older* MacBook (Intel) to test the false branch however.

pewsheen:
~~I'll see if I can run macOS 11 VM on my machine.~~

I can't run macOS 11 either lol. 
@FabianLars, do you have the machine to test this one? Otherwise I think the code is good to merge.

@JSKitty we'll need you to add a changelog in `.changes/` (https://github.com/tauri-apps/wry/blob/dev/.changes/readme.md)

FabianLars:
I think I still have a VM with it somewhere, will check after breakfast

FabianLars:
i meant breakfast tomorrow 🙃

FabianLars:
Okay so my macOS 11 VM has gone missing and my macOS 10.15 VM is broken beyond repair, so i used https://developer.apple.com/documentation/webkit/wkwebview/isinspectable?language=objc to simulate a similar situation on my macOS 12 VM. At least in that test, the kind of version guard we have is enough, which is enough for me to approve this.

FabianLars (approved):

pewsheen (approved):

PR #1671, opened 2026-02-17
refactor: replace `kuchikiki` with `dom_query`

thomaseizinger:
The `kuchikiki` dependency has not been actively maintained in a while and pulls in a lot of old dependencies. We can replace fairly easily with the more maintained `dom_query` crate.

See https://github.com/tauri-apps/tauri/issues/14958 for details.


thomaseizinger:
I noticed that this code doesn't have any tests. Should we extract this into a non-Android gated module somewhere and write some tests that can run in CI?

FabianLars:
thanks for the pr! much appreciated :) `cargo update derive_more --precise 2.0.1` or something like that should fix CI error.

> I noticed that this code doesn't have any tests. Should we extract this into a non-Android gated module somewhere and write some tests that can run in CI?

Sounds reasonable but it wouldn't be a blocker for me

thomaseizinger:
> thanks for the pr! much appreciated :) `cargo update derive_more --precise 2.0.1` or something like that should fix CI error.

Done!

thomaseizinger:
> > I noticed that this code doesn't have any tests. Should we extract this into a non-Android gated module somewhere and write some tests that can run in CI?
> 
> Sounds reasonable but it wouldn't be a blocker for me

I don't want to break anything so I'll see to write a test for this.

Legend-Master:
~~I have not got the time to take a closer look yet, just from a quick browse, it seems like there's  https://crates.io/crates/scraper which is much more popular, do we have any reason choosing `dom_query` over it?~~

Never mind, its main branch moved to edition 2024 last month which requires a MSRV bump

FabianLars:
> Never mind, its main branch moved to edition 2024 last month which requires a MSRV bump

wouldn't be a blocker imo. we're getting closer to edition 2024 by the day

thomaseizinger:
I'd prefer merging something sooner and not blocking on an MSRV upgrade. Once the tests are in place, it should be easy to migrate though.

Cargo.toml
```diff
@@ -195,14 +195,13 @@ objc2-app-kit = { version = "0.3.0", default-features = false, features = [
...
 crossbeam-channel = "0.5"
-html5ever = "0.29"
-kuchiki = { package = "kuchikiki", version = "=0.8.8-speedreader" }
 sha2 = "0.10"
 base64 = "0.22"
 jni = "0.21"
 ndk = "0.9"
 tao-macros = "0.1"
 libc = "0.2"
+dom_query = "0.25.1"
```
Legend-Master: I think we can use `default-features = false` here since we don't need the markdown feature from it

Cargo.toml
```diff
@@ -212,6 +211,9 @@ winit = "0.30"
 getrandom = "0.3"
 http-range = "0.1"
 percent-encoding = "2.3"
+base64 = "0.22"
+sha2 = "0.10"
+dom_query = "0.25.1"
```
Legend-Master: Do we need this?

Legend-Master (commented):
Seems like the failed miri test is from https://github.com/servo/stylo/issues/299

Cargo.toml
thomaseizinger: I don't think CI is running tests on Android, which is why I kept it outside the `android` module.

Legend-Master: Ah, I see now, will revert back to the previous solution

Legend-Master (dismissed):
Could you also add change file?

thomaseizinger:
> Could you also add change file?

Done!

Legend-Master (approved):

Legend-Master:
Thanks for the work and patience

Issue #1358, opened 2024-09-09
Add screenshot capability


zakhenry:
**Is your feature request related to a problem? Please describe.**
I want to be able to capture screenshots of the webview, specifically from the rust side both for integration testing and as a part of error reporting when the webview is in an untrusted state.

**Describe the solution you'd like**
On the WebviewWindow, I'd like a method `async capture_screenshot() -> Bytes` or similar.

**Describe alternatives you've considered**
It might be possible to have the capture performed by the javascript, though this might be untrustworthy if the webapp is in a broken state.

**Additional context**
webkit ref: https://developer.apple.com/documentation/webkit/wkwebview/2873260-takesnapshot
chromiumoxide ref: https://docs.rs/chromiumoxide/latest/chromiumoxide/page/struct.Page.html#method.screenshot

PR #1674, opened 2026-02-23
feat: add Rust-side webview screenshot API

josh-richardson:
Related to #1358. Prior art: #266, #1258. I'd really like to be able to take screenshots of my webviews in my Wry app.

Adds a Rust-side screenshot API for `WebView`: `WebView::screenshot(handler)` which returns visible-region screenshots as PNG bytes. Implemented for and tested on:
- Windows 11 Pro 25H2 26200.7840
- Arch Linux (6.6.119-1-MANJARO, Gnome 49.2, Wayland (had to run with `WEBKIT_DISABLE_DMABUF_RENDERER=1`))
- macOS Sequoia 15.6
- Unsupported and stubbed on iOS and Android.

I've also added  `examples/screenshot_smoke.rs` (inline HTML, waits for page load + 1s, writes `screenshot.png`). I suspect this should probably be removed before this is merged, I just wanted an easy way of testing cross platform.

Scope is intentionally limited to visible-region screenshots only (no full-document capture, afaict APIs to do this cross platform are not mature / well standardized yet). 

On Linux I'm using webkit2gtk snapshot API and encodes PNGs via GTK/GDK Pixbuf, so it does not introduce a cairo dependency. On macOS, it uses WKWebView’s native snapshot API rather than OS/screen capture APIs, which avoids the screen-recording permission requirement that affected the older approach.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2033,6 +2033,21 @@ impl WebView {
     self.webview.print()
   }
 
+  /// Capture a PNG screenshot of the currently visible webview contents.
+  ///
+  /// The screenshot is returned asynchronously via `handler`.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / macOS / Windows**: Implemented (visible region only).
+  /// - **Android / iOS**: Unsupported.
+  pub fn screenshot<F>(&self, handler: F) -> Result<()>
+  where
+    F: Fn(Result<Vec<u8>>) + 'static + Send,
+  {
+    self.webview.screenshot(handler)
+  }
+
   /// Get a list of cookies for specific url.
   pub fn cookies_for_url(&self, url: &str) -> Result<Vec<cookie::Cookie<'static>>> {
     self.webview.cookies_for_url(url)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2040,7 +2040,7 @@ impl WebView {
   /// ## Platform-specific
   ///
   /// - **Linux / macOS / Windows**: Implemented (visible region only).
-  /// - **Android / iOS**: Unsupported.
+  /// - **Android / iOS**: Not supported; `handler` will never be called.
   pub fn screenshot<F>(&self, handler: F) -> Result<()>
   where
     F: Fn(Result<Vec<u8>>) + 'static + Send,
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -72,7 +72,7 @@ use crate::wkwebview::util::operating_system_version;
 use objc2_web_kit::WKWebView;
 
 use objc2_web_kit::{
-  WKAudiovisualMediaTypes, WKInactiveSchedulingPolicy, WKSnapshotConfiguration, WKURLSchemeHandler,
+  WKAudiovisualMediaTypes, WKInactiveSchedulingPolicy, WKURLSchemeHandler,
   WKUserContentController, WKUserScript, WKUserScriptInjectionTime, WKWebViewConfiguration,
   WKWebsiteDataStore,
 };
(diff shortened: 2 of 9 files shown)


src/wkwebview/mod.rs
```diff
@@ -836,6 +840,55 @@ r#"Object.defineProperty(window, 'ipc', {
...
 
+  pub fn screenshot<F>(&self, handler: F) -> crate::Result<()>
+  where
+    F: Fn(Result<Vec<u8>>) + 'static + Send,
+  {
+    // Safety: objc runtime calls are unsafe
+    #[cfg(target_os = "macos")]
+    unsafe {
+      let config = WKSnapshotConfiguration::new(self.mtm);
+      config.setAfterScreenUpdates(true);
```
FabianLars: this is the default https://developer.apple.com/documentation/webkit/wksnapshotconfiguration/afterscreenupdates?language=objc so i think we can remove this and give takeSnapshotWithConfiguration `None` for the config?

src/wkwebview/mod.rs
```diff
@@ -836,6 +840,55 @@ r#"Object.defineProperty(window, 'ipc', {
...
+
+      let callback = block2::RcBlock::new(move |image: *mut NSImage, _err: *mut NSError| {
```
FabianLars: should we not check for an error first?

src/wkwebview/mod.rs
```diff
@@ -836,6 +840,55 @@ r#"Object.defineProperty(window, 'ipc', {
...
+        };
+
+        let props: Retained<NSDictionary<NSBitmapImageRepPropertyKey, AnyObject>> =
+          NSDictionary::new();
+        let png = bitmap.representationUsingType_properties(NSBitmapImageFileType::PNG, &props);
+        match png {
+          Some(data) => handler(Ok(data.to_vec())),
+          None => handler(Err(Error::NilScreenshot)),
+        }
+      });
```
FabianLars, line 884: this is _very_ different from all other implementations i saw on the internet (see the other 2 prs for how it usually looks like). Do you have some source for this perhaps? Also, did you actually test this on macOS?
Looks like people usually have issues with scaling with retina somethingsomething

src/wkwebview/mod.rs
```diff
@@ -836,6 +840,55 @@ r#"Object.defineProperty(window, 'ipc', {
...
+      self
+        .webview
+        .takeSnapshotWithConfiguration_completionHandler(Some(&config), &callback);
+    }
+
+    #[cfg(target_os = "ios")]
+    {
+      // Unsupported
+      let _ = handler;
+    }
```
FabianLars, line 895: we should add this before merging but i think i can take this on if you can't do mobile :)

src/lib.rs
```diff
@@ -2073,6 +2033,21 @@ impl WebView {
...
 
+  /// Capture a PNG screenshot of the currently visible webview contents.
+  ///
+  /// The screenshot is returned asynchronously via `handler`.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / macOS / Windows**: Implemented (visible region only).
+  /// - **Android / iOS**: Not supported; `handler` will never be called.
+  pub fn screenshot<F>(&self, handler: F) -> Result<()>
```
FabianLars, line 2044: would be so awesome if we could abstract away the handler but i guess that's more something for crate consumers like tauri

FabianLars (commented):
just some comments on macos, windows and linux look good at first glance and seem to match the other PRs and https://github.com/Fractal-Tess/tauri-plugin-snapshot (which i helped with back then so take that with a grain of salt)

josh-richardson:
Thanks for the review. I'll address these tomorrow when I wake up.

src/wkwebview/mod.rs
josh-richardson: This ought to be resolved

src/wkwebview/mod.rs
josh-richardson: Implemented

src/wkwebview/mod.rs
josh-richardson: I did test this on macOS Sequoia 15.6, but it's entirely possible that I missed bugs if there are issues with scaling at high DPIs. Implementation is now closer to the other open PRs.

src/wkwebview/mod.rs
josh-richardson: I can probably try my hand but I can't promise anything good - will update if I have issues.

GavinPacini:
Colleague of @josh-richardson's here. Want to add that I've tested the latest commit on both:
 - a Macbook Pro running macOS 26.3 and 
 - a hackintosh running macOS 15.6.1 (connected to via DP to 1440p LG monitor)
 
Screenshot is 1:1 with the window I see on my screen, so can confirm working on both Retina and non-Retina displays.

PR #1676, opened 2026-02-25
Fix display capture permission

Eiddy5:


Eiddy5:
ok

FabianLars:
it's not this simple, see https://github.com/tauri-apps/wry/pull/1196

PR #1677, opened 2026-02-27
fix(macOS): panic when loading multiple dylibs (fix #1675)

### LINKED ISSUE #1675: macOS crash when loading wry from multiple dylibs in the same process
If you load two dylibs linked to `wry` into the same process on macOS and create webviews, we'll panic:

```
thread '<unnamed>' (15953063) panicked at /Users/russruss/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wry-0.54.2/src/wkwebview/class/wry_web_view.rs:37:1:
could not create new class WryWebView. Perhaps a class with that name already exists?
```

Expected behavior would be to not panic :-)

Seems quite likely the root cause is https://github.com/madsmtm/objc2/issues/713 and https://github.com/rust-windowing/raw-window-metal/issues/29, which has been fixed on a currently unreleased version of `objc2`.

This was working way back in `0.45.0`, but was broken in `0.46.0` and remains broken in `0.54.2`, the latest version as of this report.  So, it seems that the migration to objc2 introduced this issue.

OS: macOS (tested 26.3 but likely not dependent on version)
Rustc: tested `1.93.1` but likely not dependent on rustc version

The pattern of loading multiple dylibs is quite common in "plug-in" software (particularly audio plug-ins), so it would be nice if this use case were supported!

Seems like to fix this we'd have to:

 - Release an objc2 version with the fix
 - Move to "auto-generated" class names for all `define_class`es

russellmcc:
closes #1675 

This fixes the panic by using the new "auto-generating" name feature of `define_class!` `objc2` - the name of the objective C class for the delegates will include the wry version number so will not be unsound to have multiple in the process.

I also added a unique name to the custom protocol handler, as this was causing issues even after the fix above.

I tested this with a local build of a few audio plug-ins using wry loaded into the same process

### DIFF
--- a/.changes/mac-dylib.md
+++ b/.changes/mac-dylib.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix bug where wry would crash when loaded by multiple dylibs in the same process on macOS.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,7 +192,7 @@ version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1d59b4c170e16f0405a2e95aff44432a0d41aa97675f3d52623effe95792a037"
 dependencies = [
- "objc2 0.6.0",
+ "objc2 0.6.4",
 ]
 
 [[package]]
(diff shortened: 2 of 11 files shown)


pewsheen:
Thank you! Do you have a minimum reproducible demo or repo so we can have a quick test for it.

russellmcc:
> Thank you! Do you have a minimum reproducible demo or repo so we can have a quick test for it.

I can put something together!

russellmcc:
Here's a repro of both issues:

 - The crash
 - Protocol incorrectly sharing between unrelated dylibs

https://github.com/russellmcc/wry-dylib-repro

pewsheen (approved):
Looks good! Your repro is clean and very helpful!

PR #1680, opened 2026-03-03
fix(windows): infinite loop on handler error

Legend-Master:
Regression from #1394

We need to send back the error instead of returning the error from the handler, not sure if there's a better way to write this though

Note: this is not a fix for the re-entrant hang https://github.com/tauri-apps/wry/issues/583#issuecomment-3186158240

### DIFF
--- a/.changes/infinite-loop-on-handler-error.md
+++ b/.changes/infinite-loop-on-handler-error.md
@@ -1,0 +1,5 @@
+---
+'wry': 'patch:bug'
+---
+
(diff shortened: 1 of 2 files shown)


FabianLars:
crazy that this wasn't noticed in 2 years

FabianLars (approved):
(untested)

Legend-Master (commented):
Guess we just assumed the hang was some sort of dead lock we had because of the re-entrant thing

PR #1682, opened 2026-03-06
docs: add gtk opengl triangle example

alvindimas05:
Since wgpu doesn't support or conflict on GTK, I added GTK OpenGL with glow example implementation as an alternative.


FabianLars (approved):
didn't check the code but the example works on my end :) thanks!

PR #1685, opened 2026-03-10
docs: `WebViewAttributes` stability

Legend-Master:
Mainly copied in the one we had for tauri config

Looking at the patch version we just released, #1649 is technically breaking since we added a field to `WebViewAttributes` with all `pub` fields

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -549,11 +549,13 @@ pub type WebViewId<'a> = &'a str;
 /// and create the struct with the [Struct Update Syntax] (`..Default::default()`), which may need a
 /// `#[allow(clippy::needless_update)]` attribute if you are declaring all fields.
 ///
-/// ```no_run
+/// ```
+/// # use wry::WebViewAttributes;
+///
 /// let attributes = WebViewAttributes {
 ///   visible: false,
 ///   ..Default::default()
-/// }
+/// };
 ///
 /// let WebViewAttributes {
 ///   visible,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -540,6 +540,29 @@ pub struct NewWindowFeatures {
 /// An id for a webview
 pub type WebViewId<'a> = &'a str;
 
+/// # Stability
+///
+/// We might add new fields in patch versions, so prefer [`WebViewBuilder`] when possible
+///
+/// If you do decide to use this, to avoid breakage,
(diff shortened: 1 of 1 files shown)


Legend-Master:
Seems like `drag_drop_handler` is marked non `pub` when `drag-drop` feature is disabled, I wonder if we should still keep this feature flag, like all the implementations are not gated by the feature flag, only the entry point (`WebViewAttributes::drag_drop_handler` and `WebViewBuilder::with_drag_drop_handler`) are

We should either remove the flag or properly gate the implementations

pewsheen:
> Seems like `drag_drop_handler` is marked non `pub` when `drag-drop` feature is disabled, I wonder if we should still keep this feature flag, like all the implementations are not gated by the feature flag, only the entry point (`WebViewAttributes::drag_drop_handler` and `WebViewBuilder::with_drag_drop_handler`) are
> 
> We should either remove the flag or properly gate the implementations

Sounds reasonable to me. @amrbashir Do you remember why drag-and-drop needs to be guarded?

amrbashir:
I think it was done so users won't see the field if they didn't enable the feature flag, but it shouldn't matter either way

FabianLars:
marking the struct as non_exhaustive would be too annoying to use right?

wondering if it makes sense to have stability docs like this in a 0.x library instead of adding a semver ci check that helps prevent us breaking it in a patch.

and if we were to release wry v1 i'd assume a builder pattern or whatever instead of an unstable struct would be used.

Legend-Master:
> marking the struct as non_exhaustive would be too annoying to use right?

You can't construct a `non_exhaustive` struct directly?

I think even we're on 0.x, a breaking change still requires everything depending on it to upgrade and without a notice like this, almost any new features will be 'breaking' while nothing's actually breaking

Alternatively, remove it from the public APIs, the builder should fully cover its use cases except for maybe getting data out the builder

FabianLars:
> You can't construct a non_exhaustive struct directly?

lol

> I think even we're on 0.x, a breaking change still requires everything depending on it to upgrade and without a notice like this, almost any new features will be 'breaking' while nothing's actually breaking

yeah but if devs have to lock the patch versions cause they're breaking why do we even use semver?

> Alternatively, remove it from the public APIs, the builder should fully cover its use cases except for maybe getting data out the builder

I would not mind that at all! If not even tauri uses the struct, who does?

PR #1692, opened 2026-03-15
fix(webkitgtk): normalize background color values to 0.0–1.0

### LINKED ISSUE #1690: set_background_color doesn't work on WebKitGTK — raw u8 values passed to gdk::RGBA::new() which expects 0.0–1.0
`backgroundColor` is ignored on Linux (WebKitGTK) — the webview is always white no matter what color you set.
 
The problem is in `src/webkitgtk/mod.rs`: the u8 color values (0–255) get cast to f64 and passed straight into `gdk::RGBA::new()`, which expects 0.0–1.0. GDK clamps anything >1.0 to 1.0, so any non-zero color ends up as white.
 
Happens in two places — line 294 (init) and line 765 (runtime setter):
 
```rust
webview.set_background_color(&gtk::gdk::RGBA::new(
  red as _, green as _, blue as _, alpha as _,  // u8 → f64, not normalized
));
```
 
Should be:
 
```rust
webview.set_background_color(&gtk::gdk::RGBA::new(
  red as f64 / 255.0,
  green as f64 / 255.0,
  blue as f64 / 255.0,
  alpha as f64 / 255.0,
));
```
 
I can put up a PR if you want.
 
**Environment:** wry 0.54.2 (also checked dev branch), webkit2gtk 2.50.5, Fedora 43 / GNOME Wayland

Sbenazar:
## Summary

`gdk::RGBA::new()` expects color values in the 0.0–1.0 range, but the `u8` components (0–255) from `background_color` were cast directly to `f64` without normalizing. GDK clamps anything >1.0 to 1.0, so any non-zero background color ended up as white on Linux.

This PR divides each component by 255.0, fixing both the init path and the runtime setter.

Fixes #1690

### DIFF
--- a/.changes/fix-webkitgtk-background-color.md
+++ b/.changes/fix-webkitgtk-background-color.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+fix(webkitgtk): normalize background color values to 0.0–1.0
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -292,7 +292,10 @@ impl InnerWebView {
       // background color
       if let Some((red, green, blue, alpha)) = attributes.background_color {
         webview.set_background_color(&gtk::gdk::RGBA::new(
-          red as _, green as _, blue as _, alpha as _,
+          red as f64 / 255.0,
+          green as f64 / 255.0,
+          blue as f64 / 255.0,
+          alpha as f64 / 255.0,
         ));
       }
     }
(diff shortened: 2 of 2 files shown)


FabianLars:
Thanks for the PR! Can you perhaps [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits? We have branch protections in-place requiring that.

Also, please add a short changefile like https://github.com/Legend-Master/wry/blob/e9445e4421252c70b8e2f76559747492fd130e89/.changes/remove-drag-drop-feature.md (using patch instead of minor)

pewsheen (approved):
Thank you!

PR #1703, opened 2026-03-26
fix(linux): use explicit Ext trait imports

FabianLars:
fixes https://github.com/tauri-apps/tauri/issues/15149

makes no sense to me but here we are

### DIFF
--- a/.changes/rust-194.md
+++ b/.changes/rust-194.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
+Fixed a compile error about missing Ext traits when building wry 0.24 / tauri v1 apps with Rust 1.94.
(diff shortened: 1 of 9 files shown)


.changes/rust-194.md
```diff
@@ -0,0 +1,5 @@
+---
+wry: patch
+---
+
+Fixed a compile error about missing Ext traits when building wry 0.24 / tauri v1 apps with Rust 1.94.
```
FabianLars, line 5: 1.93 worked on my machine (it's what i used before noticing i wasn't on 1.94) - but i can update the changefile to mentioned 1.93 even if it may not be problematic for everyone.

bg-softcom:
Nevermind, 1.93 worked for me too, I must have misremembered, sorry

Thanks for the fix!

Issue #1379, opened 2024-10-03
shouldInterceptRequest is not called on main thread (android) causing UB + race conditions


jkelleyrtp:
**Describe the bug**

The mechanism driving handleRequest on android has an `unsafe impl Sync` bound that allows it to be called from `shouldInterceptRequest`.

However, this is wrong, since `shouldInterceptRequest` is not called from the main thread:

> Note: This method is called on a thread other than the UI thread so clients should exercise caution when accessing private data or the view system.

https://developer.android.com/reference/android/webkit/WebViewClient#shouldInterceptRequest(android.webkit.WebView,%20android.webkit.WebResourceRequest)

This is causing dioxus to segfault, locks to fail, RC to double-drop, etc.

This function should instead use a channel to block until the response is ready from the main thread.

It's likely there are more functions being called from not-the-main-thread.

PR #1715, opened 2026-04-13
fix(android): Speed up async dev assets loading

### LINKED ISSUE #14097: [bug] vite file loading on android very slow
### Describe the bug

I have a vite backend which listens on port 5173, my hosts wifi ip is `192.168.1.93`.

If I run `cargo tauri android dev`, it opens the app on my android device. Then it takes about 3 minutes to load all vite files. Most of the time the connections are stalled for up to 50s. All files are loaded from `tauri.localhost`.

In the documentation there's not much about `tauri.localhost`.

I changed the url of the webview to `http://192.168.1.93:5173` via chrome://inspect and the site loaded in about 3 seconds, so I realized sth. is really wrong with this `tauri.localhost` domain.

My next try was to directly set `http://192.168.1.93:5173` in tauri.conf.json at `app.windows.url`. Somehow it did not change anything. So I tried a different url like `https://github.com`, that worked. Next try with a wrong ip: `http://192.168.1.92:5173` -> Host not reachable.

So I realized my hosts ip is swapped out to the extremly slow `tauri.localhost` domain. After some tingering I found a workaround which works for my setup to have the tauri app up in seconds:

```
cargo tauri android dev --config '{"app":{"windows":[{"url":"http://192.168.1.93:5173"}]}}' --host 0.0.0.0
```

the host argument is needed, so my url is not changed back to `tauri.localhost`.

I guess this is a bug, at least it is not documented very well what `tauri.localhost` is. I tried to find the code passages to fix it myself, but I couldn't find it in the tauri source code.

### Reproduction

_No response_

### Expected behavior

1. It works quickly out of the box and needs no workaround.
2. The documentation explains what `tauri.localhost` is.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Fedora 42.0.0 x86_64 (X64) (gnome on wayland)
    ✔ webkit2gtk-4.1: 2.48.5
    ✔ rsvg2: 2.60.0
    ✔ rustc: 1.85.0 (4d91de4e4 2025-02-17)
    ✔ cargo: 1.85.0 (d73d2caf9 2024-12-31)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: 1.85.0-x86_64-unknown-linux-gnu (overridden by environment variable RUSTUP_TOOLCHAIN)
    - node: 22.11.0
    - yarn: 4.8.1
    - npm: 10.9.0

[-] Packages
    - tauri 🦀: 2.8.4
    - tauri-build 🦀: 2.4.0
    - wry 🦀: 0.53.2
    - tao 🦀: 0.34.2
    - tauri-cli 🦀: 2.8.3
    - @tauri-apps/api : not installed!
    - @tauri-apps/cli : 2.8.3

[-] Plugins
    - tauri-plugin-log 🦀: 2.6.0
    - @tauri-apps/plugin-log : not installed!
    - tauri-plugin-fs 🦀: 2.4.2
    - @tauri-apps/plugin-fs : not installed!

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:5173/
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

_No response_

DeTeam:
* Attempt to fix https://github.com/tauri-apps/tauri/issues/14097
* Related to https://github.com/tauri-apps/tauri/pull/15220

Looks like on Android besides loading assets sync in the tauri land we also have a single mutex lock point when accessing custom protocol handler. This PR allows releasing the lock sooner by cloning the underlying handler.

### DIFF
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -99,7 +99,13 @@ fn handle_request(
   let webview_id = env.get_string(&webview_id)?;
   let webview_id = webview_id.to_str().ok().unwrap_or_default();
 
-  if let Some(handler) = REQUEST_HANDLER.lock().unwrap().get(webview_id) {
+  let handler = REQUEST_HANDLER
+    .lock()
+    .unwrap()
+    .get(webview_id)
+    .map(|handler| handler.handler.clone());
+
+  if let Some(handler) = handler {
     #[cfg(feature = "tracing")]
     let span =
       tracing::info_span!(parent: None, "wry::custom_protocol::handle", uri = tracing::field::Empty).entered();
@@ -162,7 +168,7 @@ fn handle_request(
     let response = {
       #[cfg(feature = "tracing")]
       let _span = tracing::info_span!("wry::custom_protocol::call_handler").entered();
-      (handler.handler)(
+      handler(
         webview_id,
         final_request,
         is_document_start_script_enabled != 0,
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -110,7 +110,8 @@ fn handle_request(
 
   #[cfg(feature = "tracing")]
   let span =
-    tracing::info_span!(parent: None, "wry::custom_protocol::handle", uri = tracing::field::Empty);
+    tracing::info_span!(parent: None, "wry::custom_protocol::handle", uri = tracing::field::Empty)
+      .entered();
 
   let mut request_builder = Request::builder();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -644,7 +644,7 @@ struct WebViewAttributes<'a> {
   ///   locate your files in those directories. For more information, see [Loading in-app content](https://developer.android.com/guide/webapps/load-local-content) page.
   /// - iOS: To get the path of your assets, you can call [`CFBundle::resources_path`](https://docs.rs/core-foundation/latest/core_foundation/bundle/struct.CFBundle.html#method.resources_path). So url like `wry://assets/index.html` could get the html file in assets directory.
   pub custom_protocols:
-    HashMap<String, Box<dyn Fn(WebViewId, Request<Vec<u8>>, RequestAsyncResponder)>>,
+    HashMap<String, Box<dyn Fn(WebViewId, Request<Vec<u8>>, RequestAsyncResponder) + Send + Sync>>,
 
   /// The IPC handler to receive the message from Javascript on webview
   /// using `window.ipc.postMessage("insert_message_here")` to host Rust code.
@@ -1037,7 +1037,7 @@ impl<'a> WebViewBuilder<'a> {
   #[cfg(feature = "protocol")]
   pub fn with_custom_protocol<F>(mut self, name: String, handler: F) -> Self
   where
-    F: Fn(WebViewId, Request<Vec<u8>>) -> Response<Cow<'static, [u8]>> + 'static,
+    F: Fn(WebViewId, Request<Vec<u8>>) -> Response<Cow<'static, [u8]>> + Send + Sync + 'static,
   {
     #[cfg(any(
       target_os = "linux",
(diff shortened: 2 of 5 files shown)


src/android/binding.rs
```diff
@@ -99,7 +99,13 @@ fn handle_request(
...
   let webview_id = webview_id.to_str().ok().unwrap_or_default();
 
-  if let Some(handler) = REQUEST_HANDLER.lock().unwrap().get(webview_id) {
+  let handler = REQUEST_HANDLER
+    .lock()
+    .unwrap()
+    .get(webview_id)
+    .map(|handler| handler.handler.clone());
+
+  if let Some(handler) = handler {
```
Legend-Master: We don't yet support real async custom protocol handlers on Android, so this wouldn't really help, see https://github.com/tauri-apps/wry/blob/44e26ef27428f3b9f8d00f62a6ffda887ea8a982/src/android/mod.rs#L269-L281

We are blocking on this either ways right now

DeTeam: Right. Previously there was a mutex lock, now the blocking is on recv/send, so it should make things at least a bit better? ^_^

Legend-Master: No, not really...

DeTeam: Is this understanding correct:`shouldInterceptRequest` in `RustWebViewClient.kt` runs on a separate threadpool then there's a bunch of threads that will lock on recv and allow multiple handlers `custom_protocol_handler` to do work at the same time (if runs on a saparete, e.g. tokio, runtime), meaning that when we block in `rx.recv_timeout` we wait for a bunch of responses at once (in this PR) VS waiting on the `REQUEST_HANDLER.lock()` and doing one request at a time?

Legend-Master: Actually... I thought the `handle_request` is called on main thread because we didn't restrict the custom protocol handlers to be `Sync` but it's not, reference: #1379. So we do support async handlers, but the `Sync` bound is wrong here...

Anyways, the lock is the one blocking here then, sorry for the confusion!

src/lib.rs
```diff
@@ -644,7 +644,7 @@ struct WebViewAttributes<'a> {
   ///   locate your files in those directories. For more information, see [Loading in-app content](https://developer.android.com/guide/webapps/load-local-content) page.
   /// - iOS: To get the path of your assets, you can call [`CFBundle::resources_path`](https://docs.rs/core-foundation/latest/core_foundation/bundle/struct.CFBundle.html#method.resources_path). So url like `wry://assets/index.html` could get the html file in assets directory.
   pub custom_protocols:
```
Legend-Master, line 643: @lucasfernog We do require `Send + Sync` in tauri, and this change does reflect what we're actually doing, any thoughts?

> https://docs.rs/tauri/latest/tauri/struct.Builder.html#method.register_uri_scheme_protocol

Alternatively, we could send the request to main thread and handle things there

Legend-Master (approved):
Looks good from my part, still want some opinions from other members though

DeTeam:
This change together with https://github.com/tauri-apps/tauri/pull/15220 can nicely speed things up both modern platforms 😊  just a little nudge/reminder so this PR doesn't get lost, if that's ok @lucasfernog, whenever you have time!

DeTeam:
@lucasfernog @Legend-Master sorry for another reminder 😊  
are you folks still busy with some other release or something else and don't have time for my two ios and android PRs?

FabianLars:
Lucas is still on vacation - but we're also busy elsewhere, yes :P

DeTeam:
@lucasfernog are you back?
I saw the other mobile DX PR was merged, maybe we can include this one too in the next releases? ^_^

DeTeam:
@lucasfernog @FabianLars @Legend-Master sorry for pinging again, but it would be cool to get some closure on this PR, it's been open for 2 months already. Would you folks have capacity soon to decide how to move forward? 😊

FabianLars:
2 months are rookie numbers here 😂 i asked him in DMs - if no response by end of week i think we can go ahead with this as is.

lucasfernog (approved):

PR #1719, opened 2026-04-30
perf(macOS): avoid copying large protocol bodies

Tunglies:
On macOS, Wry now avoids an extra copy for owned custom protocol
response bodies of 128KB or larger by transferring the Vec<u8> directly into NSData.

### DIFF
--- a/.changes/macos-protocol-body-nocopy.md
+++ b/.changes/macos-protocol-body-nocopy.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+On macOS, avoid an extra copy for owned custom protocol response bodies of 128KB or larger by transferring the body buffer into `NSData`.
--- a/.changes/macos-protocol-body-nocopy.md
+++ b/.changes/macos-protocol-body-nocopy.md
@@ -2,4 +2,4 @@
 "wry": patch
 ---
 
-On macOS, avoid an extra copy for owned custom protocol response bodies of 128KB or larger by transferring the body buffer into `NSData`.
+On macOS, avoid an extra copy for owned custom protocol response bodies by transferring the body buffer into `NSData`.
--- a/src/wkwebview/class/url_scheme_handler.rs
+++ b/src/wkwebview/class/url_scheme_handler.rs
@@ -26,8 +26,6 @@ use objc2_web_kit::{WKURLSchemeHandler, WKURLSchemeTask};
 
 use crate::{wkwebview::WEBVIEW_STATE, RequestAsyncResponder, WryWebView};
 
-const NO_COPY_DATA_THRESHOLD: usize = 128 * 1024;
-
 pub fn create(name: &str) -> &AnyClass {
   unsafe {
     // Include the address of WEBVIEW_STATE in the class name so that each dylib in the process
--- a/src/wkwebview/class/url_scheme_handler.rs
+++ b/src/wkwebview/class/url_scheme_handler.rs
@@ -267,8 +267,8 @@ extern "C" fn start_task(
                 .map_err(|_e| crate::Error::CustomProtocolTaskInvalid)?;
 
                 let data = match sent_response.into_body() {
-                    Cow::Owned(content) => NSData::from_vec(content),
-                    Cow::Borrowed(content) => NSData::with_bytes(content),
+                  Cow::Owned(content) => NSData::from_vec(content),
+                  Cow::Borrowed(content) => NSData::with_bytes(content),
                 };
 
                 // Check validity again
--- a/src/wkwebview/class/url_scheme_handler.rs
+++ b/src/wkwebview/class/url_scheme_handler.rs
@@ -4,7 +4,7 @@
 
 use std::{
   borrow::Cow,
-  ffi::{c_char, c_void, CStr},
+  ffi::{c_char, CStr},
   panic::AssertUnwindSafe,
   ptr::NonNull,
 };
@@ -268,20 +268,14 @@ extern "C" fn start_task(
                 .map_err(|_e| crate::Error::CustomProtocolTaskInvalid)?;
 
                 let data = if content_len < NO_COPY_DATA_THRESHOLD {
-                  let data = NSData::alloc();
                   // Keep small responses on the original copy path; no-copy deallocation costs more.
-                  NSData::initWithBytes_length(data, content.as_ptr() as *mut c_void, content.len())
+                  NSData::with_bytes(content)
                 } else {
                   match sent_response.into_body() {
                     Cow::Owned(content) => NSData::from_vec(content),
                     Cow::Borrowed(content) => {
-                      let data = NSData::alloc();
                       // Copy borrowed responses because NSData cannot take ownership.
-                      NSData::initWithBytes_length(
-                        data,
-                        content.as_ptr() as *mut c_void,
-                        content.len(),
-                      )
+                      NSData::with_bytes(content)
                     }
                   }
                 };
--- a/src/wkwebview/class/url_scheme_handler.rs
+++ b/src/wkwebview/class/url_scheme_handler.rs
@@ -266,21 +266,9 @@ extern "C" fn start_task(
                 }))
                 .map_err(|_e| crate::Error::CustomProtocolTaskInvalid)?;
 
-                let content = sent_response.into_body();
-                // It yields benefits by eliminating buffer copying starting from 128 KB.
-                // However, sizes below 64 bytes cause regressions(or noise). 128 KB as
-                // it aligns better with engineering standards and is more human-readable.
-                let data = if content_len < NO_COPY_DATA_THRESHOLD {
-                  // Keep small responses on the original copy path; no-copy deallocation costs more.
-                  NSData::with_bytes(content.as_ref())
-                } else {
-                  match content {
+                let data = match sent_response.into_body() {
                     Cow::Owned(content) => NSData::from_vec(content),
-                    Cow::Borrowed(content) => {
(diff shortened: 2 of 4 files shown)


Legend-Master (changes requested):
Don't think I understand this, [`NSData::from_vec`](https://docs.rs/objc2-foundation/latest/objc2_foundation/struct.NSData.html#method.from_vec) uses `initWithBytes_length` internally as well, how is it avoiding the copy here? And why are we only doing this for the bigger chunks?

That being said, I don't mind if we migrate to `NSData::with_bytes` here for readability though

Tunglies:
[NSData::from_vec](https://github.com/madsmtm/objc2/blob/d70c2a50086cc050c47726dc3e37e7ec4ddf30df/framework-crates/objc2-foundation/src/data.rs#L30-L32) did not use `initWithBytes_length` but `with_vec` internally. And [with_vec](https://github.com/madsmtm/objc2/blob/d70c2a50086cc050c47726dc3e37e7ec4ddf30df/framework-crates/objc2-foundation/src/data.rs#L310-L336)  uses [initWithBytesNoCopy:length:deallocator:](https://github.com/madsmtm/objc2/blob/d70c2a50086cc050c47726dc3e37e7ec4ddf30df/framework-crates/objc2-foundation/src/data.rs#L329-L336) to hand the Vec buffer to NSData without copying.

The chunk size is a workaround based on local benchmarks; it yields benefits by eliminating buffer copying starting from 88 bytes. However, sizes below 88 bytes cause regressions. 128 bytes as it aligns better with engineering standards and is more human-readable.

Legend-Master:
Sorry my bad, looked at the wrong line 🤦‍♂️, thanks for checking

src/wkwebview/class/url_scheme_handler.rs
```diff
@@ -215,6 +217,7 @@ extern "C" fn start_task(
                 check_task_is_valid(&webview, task_key, task_uuid.clone())?;
 
                 let content = sent_response.body();
```
Legend-Master:
```suggestion
```

Remove this and rely on the `sent_response.into_body()` below

src/wkwebview/class/url_scheme_handler.rs
```diff
@@ -264,15 +267,18 @@ extern "C" fn start_task(
...
-                // Send data
-                let data = NSData::alloc();
-                // MIGRATE NOTE: we copied the content to the NSData because content will be freed
-                // when out of scope but NSData will also free the content when it's done and cause doube free.
-                let data = NSData::initWithBytes_length(
-                  data,
-                  content.as_ptr() as *mut c_void,
-                  content.len(),
-                );
+                let data = if content_len < NO_COPY_DATA_THRESHOLD {
```
Legend-Master: Let's add some comments about the reasons, ideally share some benchmark result details

Legend-Master (dismissed):
Looks good to me, just some nitpicks

Tunglies:
Some custom benchmark, for only IPC performance part. 
  
| Payload | dev ms/call | perf ms/call | Delta | dev throughput | perf throughput |
  |---:|---:|---:|---:|---:|---:|
  | 128KB | 0.458 ms | 0.437 ms | -4.56% | 272.8 MiB/s | 285.8 MiB/s |
  | 256KB | 0.568 ms | 0.515 ms | -9.30% | 440.1 MiB/s | 485.3 MiB/s |
  | 512KB | 0.768 ms | 0.671 ms | -12.68% | 652.0 MiB/s | 745.9 MiB/s |
  | 1024KB | 1.124 ms | 0.918 ms | -18.35% | 892.1 MiB/s | 1091.1 MiB/s |
  | 2048KB | 1.729 ms | 1.400 ms | -19.04% | 1158.6 MiB/s | 1432.3 MiB/s |
  | 3072KB | 2.371 ms | 1.800 ms | -24.10% | 1273.3 MiB/s | 1672.7 MiB/s |

Tunglies:
> Sorry my bad, looked at the wrong line 🤦‍♂️, thanks for checking

No, you good. I get some mess at time. :)

Legend-Master:
> Some custom benchmark, for only IPC performance part.

I'm entirely sure I get the table headers, like is `dev` before and `perf` after the change? Do we have a data point for below `NO_COPY_DATA_THRESHOLD` like the regression you mentioned? Also was this ran in release build, asking because simply copying a 128KB blob took way longer than I thought

Tunglies:
Adjusted my benchmark closer to IPC performance and avoid zero data optimize for benchmark, it tells 64Kb could get around 3.33%-4% improvement which conflict with the previous lower than 88KB leads regression, but now latest no-copy size regression(noise) happens on lower than 64KB.  But 64KB performance have very slightly improved. Let's keep threshold to 128 for engineering standards.

---

dev: wry 44e26ef27428f3b9f8d00f62a6ffda887ea8a982
perf: current PR, THRESHOLD=1bit
Ran in release build.

Greater than 64KB do have significant improvement.

| Payload | Copy median(dev) | No-copy median(perf) | Median delta | Verdict |
  |---:|---:|---:|---:|---|
  | 1 KiB | 0.353 ms | 0.349 ms | -0.97% | no-copy faster |
  | 2 KiB | 0.352 ms | 0.353 ms | 0.42% | no-copy slower |
  | 4 KiB | 0.364 ms | 0.363 ms | -0.20% | neutral |
  | 8 KiB | 0.364 ms | 0.367 ms | 0.80% | no-copy slower |
  | 16 KiB | 0.362 ms | 0.363 ms | 0.20% | neutral |
  | 32 KiB | 0.366 ms | 0.362 ms | -1.13% | no-copy faster |
  | 64 KiB | 0.397 ms | 0.382 ms | -3.93% | no-copy faster |
  | 128 KiB | 0.451 ms | 0.431 ms | -4.55% | no-copy faster |
  | 256 KiB | 0.553 ms | 0.520 ms | -6.01% | no-copy faster |
  | 512 KiB | 0.715 ms | 0.656 ms | -8.20% | no-copy faster |
  | 1024 KiB | 1.047 ms | 0.906 ms | -13.43% | no-copy faster |
  | 2048 KiB | 1.563 ms | 1.375 ms | -12.00% | no-copy faster |
  | 3072 KiB | 2.214 ms | 1.833 ms | -17.20% | no-copy faster |

Lower than 64KB  most are noise or regression at sometimes, not stable to test.
64KB makes very small difference, cant tells the difference in 0.01x ms.

  | Payload | Copy median(dev) | No-copy median(perf) | Median delta | Verdict |
  |---:|---:|---:|---:|---|
  | 28 KiB | 0.434 ms | 0.447 ms | 3.04% | no-copy slower |
  | 30 KiB | 0.414 ms | 0.426 ms | 2.95% | noise |
  | 32 KiB | 0.379 ms | 0.376 ms | -0.84% | noise |
  | 34 KiB | 0.383 ms | 0.376 ms | -1.83% | noise |
  | 36 KiB | 0.383 ms | 0.380 ms | -0.72% | noise |
  | 38 KiB | 0.384 ms | 0.380 ms | -0.98% | noise |
  | 40 KiB | 0.380 ms | 0.376 ms | -1.20% | noise |
  | 42 KiB | 0.382 ms | 0.374 ms | -2.10% | noise |
  | 44 KiB | 0.387 ms | 0.386 ms | -0.26% | noise |
  | 46 KiB | 0.389 ms | 0.384 ms | -1.26% | noise |
  | 48 KiB | 0.392 ms | 0.385 ms | -1.68% | noise |
  | 52 KiB | 0.393 ms | 0.388 ms | -1.41% | noise |
  | 56 KiB | 0.394 ms | 0.386 ms | -1.85% | noise |
  | 60 KiB | 0.400 ms | 0.392 ms | -1.95% | noise |
  | 64 KiB | 0.409 ms | 0.393 ms | -3.82% | no-copy faster |

Legend-Master:
If the regression is mostly noises, what about removing the threshold?

Tunglies:
> If the regression is mostly noises, what about removing the threshold?

Sort of, but really can't tells the real world performance. For stability let's keep the 128KB size for those performance improvements we can actually reproduced.

Legend-Master:
It's more about the readability here, if the performance isn't worse

src/wkwebview/class/url_scheme_handler.rs
```diff
@@ -264,15 +266,22 @@ extern "C" fn start_task(
...
+                  NSData::with_bytes(content.as_ref())
+                } else {
+                  match content {
+                    Cow::Owned(content) => NSData::from_vec(content),
+                    Cow::Borrowed(content) => {
+                      // Copy borrowed responses because NSData cannot take ownership.
+                      NSData::with_bytes(content)
+                    }
+                  }
+                };
```
Legend-Master:
```suggestion
                let data = match sent_response.into_body() {
                    Cow::Owned(content) => NSData::from_vec(content),
                    Cow::Borrowed(content) => NSData::with_bytes(content),
                };
```

let's just keep it simple

.changes/macos-protocol-body-nocopy.md
```diff
@@ -0,0 +1,5 @@
+---
+"wry": patch
+---
+
+On macOS, avoid an extra copy for owned custom protocol response bodies of 128KB or larger by transferring the body buffer into `NSData`.
```
Tunglies:
```suggestion
On macOS, avoid an extra copy for owned custom protocol response bodies by transferring the body buffer into `NSData`.
```

Legend-Master (approved):

Legend-Master:
Thanks! And also for the patience!

PR #1720, opened 2026-04-30
chore: updates for tao 0.36

lucasfernog:
ref https://github.com/tauri-apps/tao/pull/1209

### DIFF
--- a/src/android/kotlin/WryActivity.kt
+++ b/src/android/kotlin/WryActivity.kt
@@ -9,10 +9,8 @@ import android.content.Intent
 import android.os.Build
 import android.os.Bundle
 import android.webkit.WebView
-import android.view.KeyEvent
 import androidx.activity.OnBackPressedCallback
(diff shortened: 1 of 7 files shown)


src/android/kotlin/WryActivity.kt
```diff
@@ -28,26 +28,6 @@ object WryLifecycleObserver : DefaultLifecycleObserver {
         Rust.create()
         Rust.wryCreate()
```
Legend-Master: Do we still want to keep the `WryLifecycleObserver`? I also wonder why we had it in the first place.

Legend-Master: Never mind, I should have read https://developer.android.com/reference/androidx/lifecycle/ProcessLifecycleOwner first

Legend-Master (approved):

FabianLars (approved):

PR #1723, opened 2026-05-05
fix(macos): use modern pasteboard API for drag-drop file collection

vdavid:
On macOS, `collect_paths` reads dragged file paths from the deprecated `NSFilenamesPboardType` only. When the drag source publishes only the modern per-item `public.file-url` type (the standard for `NSDraggingItem.initWithPasteboardWriter:` with `NSPasteboardItem`), `availableTypeFromArray:` reports the legacy type as derivable but `propertyListForType:` returns nil, hitting an `unwrap` and panicking the whole webview process.

This switches `collect_paths` to the modern `NSPasteboard.readObjectsForClasses:options:` API with `NSURL`, which covers both legacy and modern sources in one call. The legacy `NSFilenamesPboardType` branch is kept as a defensive fallback in case `readObjectsForClasses:options:` returns nothing on some configuration. No `unwrap` calls anywhere in the new code — every fallible conversion skips the entry instead of panicking.

## Test plan

I did these:

- Ran `cargo build` and `cargo test` on `aarch64-apple-darwin` and `x86_64-apple-darwin`, with default features, `--all-features`, and `--no-default-features --features os-webview`. All combinations: 8 unit tests + 6–7 doc tests passed, 0 failed.
- Ran `cargo clippy --all-targets --all-features`, it's clean (existing unrelated `examples/gtk_opengl.rs` unused-import warnings only, present on `dev` since #1682).
- Ran `cargo fmt --all -- --check` clean.
- End-to-end validated against [Cmdr](https://getcmdr.com) (my file manager), which is the real-world drag source that triggered this bug. With Cmdr publishing only modern per-item `public.file-url` types (no `NSFilenamesPboardType`), self-drag re-entry into the wry webview no longer panics — the `readObjectsForClasses:` path returns the URLs and `collect_paths` yields the expected `Vec<PathBuf>`.

### DIFF
--- a/.changes/macos-drag-drop-modern-pasteboard.md
+++ b/.changes/macos-drag-drop-modern-pasteboard.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+On macOS, use the modern `readObjectsForClasses:options:` pasteboard API to collect dragged file paths. This fixes a panic when the drag source publishes only the modern per-item `public.file-url` type (as `NSDraggingItem.initWithPasteboardWriter:` does) instead of the legacy `NSFilenamesPboardType`, and removes the deprecated API as the primary path. The legacy type is kept as a defensive fallback.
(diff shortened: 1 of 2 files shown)


vdavid:
@amrbashir, you [seem to be](https://gitstrata.com/github.com/tauri-apps/wry) the most active contributor these days; any chance you could take a look at my fix and get it merged if you like it?

PR #955, opened 2023-05-20
fix(linux):fix fcitx input method wrong positioning

33671:
### What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

### Checklist
- [x] This PR will resolve  [IME window is not positioning](https://github.com/tauri-apps/tauri/issues/5986)
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/wry/blob/dev/.changes/readme.md).
- [x] I have added a convincing reason for adding this feature, if necessary
- [x] It can be built on all targets and pass CI/CD.

PR #1724, opened 2026-05-06
feat(webkitgtk): make input method preedit configurable

eldenpark:
## Background

I ran into this while experimenting with a Linux desktop app on **wry 0.53.5** against **webkit2gtk-4.1 2.50.x** on Ubuntu 24.04, and trying to use a plain `<input>` for Korean text. With a default wry build, typing Hangul felt broken — each keystroke either dropped the in-progress composition or briefly flickered into a detached fcitx popup that didn't track the cursor.

**The root cause is that wry currently disables IME preedit unconditionally on Linux, so there is no way for any wry-based app to display preedit text inline.** `webkitgtk::InnerWebView::set_webview_settings` calls `set_enable_preedit(false)` with no opt-out, which means CJK users never see what they're composing inside `<input>` / `<textarea>` / `contenteditable` elements — they only see characters appear after the IME commits, with no underlined preedit in between.

That call landed in wry 0.29.0 to work around fcitx popup positioning on older webkit2gtk (tauri-apps/tauri#5986). On modern webkit2gtk-4.1 (≥ 2.50) the popup positioning issue is gone, but the unconditional `false` is now actively preventing the standard IME UX that GTK, Firefox, and Chromium-based apps all render correctly on the same systems.

## What this PR does

Adds `WebViewBuilderExtUnix::with_input_method_preedit(bool)` so apps can opt in to inline preedit:

```rust
let builder = WebViewBuilder::new()
    .with_input_method_preedit(true);
```

The default stays false, so this is non-breaking — existing apps that rely on the popup behavior keep getting it.

When enabled, webkit's GtkIMContext reports compositionstart / compositionupdate / compositionend to the DOM, and the IME renders its preedit string (with the standard underline) inline at the cursor.

## Testing

Manually tested on Ubuntu 24.04 (X11) + webkit2gtk-4.1 2.50.x with two IMEs:

- fcitx5-hangul (Korean): typing 안녕 shows inline preedit underlined at the cursor; commit lands cleanly; repeated send-and-clear works in a loop.
- fcitx5-anthy (Japanese): typing romaji like nihongo shows にほんご inline; Space opens the candidate window; Enter commits.

I haven't tested Wayland or ibus — happy to add those data points if a maintainer wants them before merging.

## Related

- tauri-apps/tauri#5986 — original report behind unconditional set_enable_preedit(false)
- tauri-apps/tauri#11412 — recent user request for inline preedit
- tauri-apps/tauri#8264 — related IME report
- tauri-apps/wry#955 — earlier discussion

### DIFF
--- a/.changes/configurable-input-method-preedit.md
+++ b/.changes/configurable-input-method-preedit.md
@@ -1,0 +1,5 @@
+---
+"wry": minor
+---
+
+On Linux, add `WebViewBuilderExtUnix::with_input_method_preedit(bool)` to make the input method preedit configurable. Defaults to `false`, preserving the behavior introduced in wry 0.29.0 (see tauri-apps/tauri#5986). Apps targeting modern webkit2gtk (≥ 2.50) can opt in to inline preedit, which CJK IMEs use to render composition state inside editable elements.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1920,6 +1920,7 @@ impl WebViewBuilderExtAndroid for WebViewBuilder<'_> {
 pub(crate) struct PlatformSpecificWebViewAttributes {
   extension_path: Option<PathBuf>,
   related_view: Option<webkit2gtk::WebView>,
+  input_method_preedit: bool,
 }
 
 #[cfg(any(
@@ -1950,6 +1951,26 @@ pub trait WebViewBuilderExtUnix<'a> {
   /// Creates a new webview sharing the same web process with the provided webview.
   /// Useful if you need to link a webview to another, for instance when using the [`WebViewBuilder::with_new_window_req_handler`].
   fn with_related_view(self, webview: webkit2gtk::WebView) -> Self;
+
+  /// Set whether the WebView's input method context should report
+  /// preedit text to the application during composition.
+  ///
+  /// When `true`, IME composition events (`compositionstart`,
+  /// `compositionupdate`, `compositionend`) fire in the DOM and
+  /// CJK input methods can render their preedit inline in editable
+  /// elements. When `false` (the default), preedit is suppressed
+  /// and the IM falls back to its own popup UI; this matches the
+  /// behavior introduced in wry 0.29.0 to mitigate
+  /// popup-positioning issues on older webkit2gtk versions (see
+  /// [tauri-apps/tauri#5986]).
+  ///
+  /// On modern webkit2gtk-4.1 (≥ 2.50) inline preedit works
+  /// correctly with major IMEs (fcitx5, ibus, etc.); apps targeting
+  /// those environments can opt in here to match the IME UX of
+  /// native GTK / Firefox / Chromium-based apps.
+  ///
+  /// [tauri-apps/tauri#5986]: https://github.com/tauri-apps/tauri/issues/5986
+  fn with_input_method_preedit(self, enabled: bool) -> Self;
 }
 
 #[cfg(any(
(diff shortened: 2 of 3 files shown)


eldenpark:
Hi maintainers, would love a review when you have time. Happy to take it in a different direction if there's a better approach.

FabianLars:
since the other pr, wry's minimum webkitgtk increased and distros removed old-ish versions from their repos so i wonder if it's fine to enable for all again? Though that indeed would require more testing, so at least wayland and ideally also older webkitgtk/gtk versions and perhaps in a tiling wm as well since it was first reported there 🤔

eldenpark:
@FabianLars  I'm leaning toward enabling it for everyone too, but I unfortunately couldn't test on Wayland. I did do my testing under i3, so that covers the tiling-WM + X11 at least. 

Is there anyone on Wayland who could give it a spin before we flip the default?

FabianLars:
I have wayland kde and wayland gnome to test but i never used IMEs so idk how to set that up and use it 🙃 

I asked on our discord server, hopefully someone's up to the task there :)

takakix2:
Wayland data point for this PR — enabling preedit fixes inline IME on GNOME Wayland + fcitx5 + Mozc.

**Environment**

- Ubuntu 24.04, GNOME on **Wayland**
- fcitx5 5.1.1 (`fcitx5-frontend-gtk3` / `-gtk4` 5.1.1), **Mozc** (Japanese), `GTK_IM_MODULE=fcitx5`
- `libwebkit2gtk-4.1-0` **2.52.6**
- wry 0.54.4 / tauri 2.10.3
- Two apps: a terminal (xterm.js) and an editor (Monaco 0.55.1)

**Default (`set_enable_preedit(false)`)**

- No inline preedit and no underline; fcitx5 draws its own detached preedit.
- No `compositionstart` / `compositionupdate` / `compositionend` reach the page during conversion.

**With preedit re-enabled**

Note: this is not a build of this PR. I flipped the same setting after the webview was created, which should be equivalent to `with_input_method_preedit(true)`:

```rust
#[cfg(target_os = "linux")]
if let Some(win) = app.get_webview_window("main") {
    let _ = win.with_webview(|webview| {
        use webkit2gtk::{InputMethodContextExt, WebViewExt};
        if let Some(ctx) = webview.inner().input_method_context() {
            ctx.set_enable_preedit(true);
        }
    });
}
```

- ✅ Preedit renders inline at the caret, with the underline.
- ✅ Composition events reach the DOM in the expected order.
- ✅ The candidate window is placed next to the caret and follows the text as it grows (the positioning problem behind tauri-apps/tauri#5986 did not reproduce here).

**If the default is flipped: app-side issues that start to show up (not wry bugs)**

wry-based apps on Linux haven't received composition events since 0.29.0, so some latent bugs appear once preedit is on. Two we hit:

1. **Unpaired `compositionend`.** Mozc commits a full-width space pressed outside a conversion as `input` (`insertFromComposition`) followed by a `compositionend` **with no preceding `compositionstart`**. xterm.js's `CompositionHelper` assumes the events are paired and re-sent the previously committed text. We guard against the unpaired `compositionend` on the app side.
2. **Monaco: composed text jitters vertically on line 2+.** This is microsoft/monaco-editor#4796 (reported for Safari, i.e. WebKit), with a fix proposed in microsoft/vscode#333909 (still open, so not in Monaco 0.55.1). Setting `accessibilitySupport: 'off'` avoids it.

Neither is caused by enabling preedit, but a default flip would surface them for everyone at once, so they may be worth a line in the release notes.

**Not tested**

KDE, ibus, X11, tiling WMs, or older webkitgtk — only the setup above.

Happy to test a build of this PR on the same setup if that helps.

PR #1728, opened 2026-05-13
fix(ci): bench on new rust nightly

Legend-Master:
Reference: https://github.com/johnthagen/min-sized-rust/pull/86

Try run: https://github.com/tauri-apps/wry/actions/runs/26677612545/job/78632249480


Legend-Master:
Hmm, the test seems to be stuck for some reason, the binary works fine when tested locally on my Windows machine 🤔

Legend-Master:
~~Well, maybe just disable it this action for now then...~~

Seems like the hang was from #1532, fixed in fc23510

FabianLars (approved):

Issue #1730, opened 2026-05-17
[bug] macOS 26.3 SIGSEGV in WebKit::WebPageProxy::dispatchSetObscuredContentInsets() after destroying a secondary webview window


ni-byeongju:
**Describe the bug**

On macOS 26.3 (Tahoe), calling `getCurrentWindow().destroy()` (or `.close()`) on
a secondary webview window from within that window's JS context — typically
right after the user saves a form and the window auto-closes — crashes the
entire app with SIGSEGV in WebKit's main-thread RunLoop work, while the main
window is being re-activated as the key window.

Stack trace (Thread 0, main / com.apple.main-thread):

```
0  WebKit          WebKit::WebPageProxy::dispatchSetObscuredContentInsets() + 324
1  JavaScriptCore  WTF::RunLoop::performWork() + 552
2  JavaScriptCore  WTF::RunLoop::performWork(void*) + 36
3  CoreFoundation  __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__
…
20 tao             tao::platform_impl::platform::event_loop::EventLoop::run_return
21 tao             tao::platform_impl::platform::event_loop::EventLoop::run
22 tao             tao::event_loop::EventLoop::run
23 tauri-runtime-wry  <_ as tauri_runtime::Runtime>::run
24 tauri           tauri::app::App::run

Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x0000000000000640
```

The `0x640` offset suggests dereference of a member on an already-released
`WebPageProxy` (the secondary window's webview), happening while the main
window's KVO/inset recalculation runs on the main-thread RunLoop.

**Environment**
- macOS: 26.3 (25D125), Apple Silicon (Mac14,12)
- tauri: 2.11.0
- wry: 0.55.0
- tao: 0.35.0
- webview backend: WKWebView (system WebKit)

**Reproduction**

1. Open a secondary \`WebviewWindow\` from the main window via Tauri API.
2. From the secondary window's JS, after some user action (e.g. form submit),
   call \`getCurrentWindow().destroy()\` synchronously.
3. The main window regains key-window focus and the app SIGSEGVs on the next
   main-thread RunLoop tick.

The crash does **not** reproduce if the secondary window is first hidden and
the destroy is deferred to a later macrotask — see workaround below.

**Workaround**

[...]

PR #1734, opened 2026-05-21
fix(macos): prevent use-after-free in stop_task on macOS 11 during WKWebView dealloc

### LINKED ISSUE #1733: stop_task use-after-free crash on macOS 11 during WKWebView dealloc
**Describe the bug**
  On macOS 11 (Big Sur), the app crashes with `EXC_BAD_ACCESS` (`objc_release` on a freed pointer) during WKWebView deallocation when
  custom URL scheme handlers are registered.

  The root cause is that during WKWebView dealloc, WebKit internally calls `stopAllTasksForPage` → `platformStopTask`, which invokes
  the `stop_task` ObjC callback. On macOS 11, the task and/or webview pointers passed to this callback may already be freed. The
  current `stop_task` function signature uses objc2 reference types (`&WryWebView`, `&ProtocolObject<dyn WKURLSchemeTask>`), which
  trigger implicit `objc_retain` on function entry — calling `objc_retain` on a freed pointer causes the crash.

  There is also a related race condition: when the async protocol response handler (running on a tokio worker thread) drops its
  `Retained<WryWebView>`, this can trigger WKWebView dealloc on that thread. The dealloc path then calls `stop_task` with freed task
  pointers, and the `Retained<ProtocolObject<dyn WKURLSchemeTask>>` in the response closure may still be alive — creating a
  use-after-free if drop ordering is not enforced.

**Steps To Reproduce**
1. Register a custom URL scheme handler via `WebViewBuilder::with_custom_protocol`
  2. Load a page that triggers requests to the custom protocol
  3. Drop/close the WebView while async protocol responses may still be in-flight
  4. Crash occurs during WKWebView dealloc on macOS 11

  Minimal reproduction (conceptual):

  ```rust
  use wry::WebViewBuilder;

  fn main() {
      // ... create window ...

      let webview = WebViewBuilder::new()
          .with_custom_protocol("app".into(), |_webview_id, request, responder| {
              // Simulate async response (e.g., via tokio::spawn)
              std::thread::spawn(move || {
                  let response = http::Response::builder()
                      .status(200)
                      .body(std::borrow::Cow::from("hello".as_bytes()))
                      .unwrap();
                  responder.respond(response);
              });
          })
          .with_url("app://localhost/index.html")
          .build(&window)
          .unwrap();

      // Drop webview shortly after — on macOS 11, this triggers the crash
      // in stop_task during WKWebView dealloc
      drop(webview);
  }

  Note: The crash is timing-dependent and more likely when async responses are pending during webview destruction.

**Expected behavior**
 Dropping the WebView should not crash. The stop_task callback should safely handle the case where WebKit passes freed pointers during
   dealloc on macOS 11.


**Screenshots**
None

  - OS: macOS 11.7.10 (Big Sur)
  - Rustc: 1.88
  - wry: 0.51 and 0.48


**Additional context**
Proposed fix:

    _this: &ProtocolObject<dyn WKURLSchemeHandler>,
    _sel: objc2::runtime::Sel,
    _webview: *mut AnyObject,  // raw pointer — no implicit objc_retain
    _task: *mut AnyObject,     // raw pointer — no implicit objc_retain
  ) {
    // no-op: avoid accessing task/webview — macOS 11 may pass freed pointers here
  }

  2. Enforce explicit drop ordering in the async response handler:

  // webview must drop before task: if webview drop triggers dealloc →
  // stopAllTasksForPage → platformStopTask, the task must still be alive.
  drop(webview);
  drop(task);
  result

x93008:
macOS 11 WebKit bug: during WKWebView dealloc, `stopAllTasksForPage` calls `stop_task` with already-freed task pointers. Any access (including the implicit `objc_release` from objc2 reference types) causes SIGSEGV.

Fix:
- `stop_task`: use raw pointers (`*mut AnyObject`) instead of objc2 references to skip automatic retain/release. Body is no-op since task is invalid.
- `start_task` response handler: explicit `drop(webview)` before `drop(task)` to ensure correct deallocation order.

closes #1733

## Details

On macOS 11, WebKit's internal `stopAllTasksForPage` (triggered during `[WKWebView dealloc]`) passes already-freed `WKURLSchemeTask` pointers to the `webView:stopURLSchemeTask:` callback. Using objc2 reference types (`&WryWebView`, `&ProtocolObject<dyn WKURLSchemeTask>`) triggers implicit `objc_retain` on entry, which crashes on the freed pointer.

The explicit drop ordering (`drop(webview)` before `drop(task)`) prevents a scenario where dropping `Retained<WryWebView>` on a worker thread triggers dealloc → `stopAllTasksForPage` → `stop_task`, while the task reference is still alive in the closure.

This is the same class of race condition reported in https://github.com/tauri-apps/tauri/issues/11516 (macOS 15) and https://github.com/tauri-apps/wry/issues/1730.

### DIFF
--- a/src/wkwebview/class/url_scheme_handler.rs
+++ b/src/wkwebview/class/url_scheme_handler.rs
@@ -286,12 +286,18 @@ extern "C" fn start_task(
                 }))
                 .map_err(|_e| crate::Error::CustomProtocolTaskInvalid)?;
 
-                if WEBVIEW_STATE.read().unwrap().contains_key(webview_id) {
+                let result = if WEBVIEW_STATE.read().unwrap().contains_key(webview_id) {
                   webview.remove_custom_task_key(task_key);
                   Ok(())
                 } else {
                   Err(crate::Error::CustomProtocolTaskInvalid)
-                }
+                };
+
+                // webview must drop before task: if webview drop triggers dealloc →
+                // stopAllTasksForPage → platformStopTask, the task must still be alive.
+                drop(webview);
+                drop(task);
+                result
               }
 
               #[cfg(feature = "tracing")]
@@ -334,8 +340,8 @@ extern "C" fn start_task(
 extern "C" fn stop_task(
   _this: &ProtocolObject<dyn WKURLSchemeHandler>,
   _sel: objc2::runtime::Sel,
-  webview: &WryWebView,
-  task: &ProtocolObject<dyn WKURLSchemeTask>,
+  _webview: *mut AnyObject,
+  _task: *mut AnyObject,
 ) {
-  webview.remove_custom_task_key(task.hash());
+  // no-op: avoid accessing task/webview — macOS 11 may pass freed pointers here
 }
--- a/.changes/fix-macos11-stop-task-uaf.md
+++ b/.changes/fix-macos11-stop-task-uaf.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+On macOS 11, fix use-after-free crash in custom protocol `stop_task` during WKWebView dealloc by using raw pointers instead of objc2 references and enforcing correct drop ordering in the async response handler.


x93008:
## Tested Platforms

| OS | Version | Result |
|---|---|---|
| macOS | 11.7.10 (Big Sur) | ✅ |
| macOS | 12.7.3 (Monterey) | ✅ |
| macOS | 13.6 (Ventura) | ✅ |
| macOS | 14.2.1 (Sonoma) | ✅ |
| macOS | 15.7.2 (Sequoia) | ✅ |
| macOS | 26.4.1 (Tahoe) | ✅ |
| Windows 10 | 22H2 (19045.5011) | ✅ |
| Windows 11 | 25H2 (26200.8457) | ✅ |

All platforms pass without crash.

sftse:
Without knowing a lot about the platform, it would be very odd for pointers passed from the platform to user code to be invalid.

The manual drop order looks really wrong as well, a mixture between manual memory management and refcounting.
Probably worth investigating those arguments of the function on which somebody put `FIXME`.

x93008:
Thanks for the feedback. Let me address both points:

**Re: "it would be very odd for pointers passed from the platform to user code to be invalid"**

You're right that this *shouldn't* happen, but it's a known WebKit bug that was fixed upstream. The specific issue is in `WebURLSchemeHandlerCocoa::platformStopTask` — prior to the fix in [WebKit changeset 277235](https://www.mail-archive.com/webkit-changes@lists.webkit.org/msg171685.html) (bug [#225373](https://bugs.webkit.org/show_bug.cgi?id=225373)):

> *"We weren't explicitly keeping a strong reference to the task given to the API client in startURLSchemeTask: and stopURLSchemeTask: which could cause all our internal maps to release their references to the task"*

> *"We were keeping a HashSet of raw WebURLSchemeHandler pointers, then using each of them without keeping it alive."*

The fix added `auto strongTask = retainPtr(wrapper(task))` before calling into `stopURLSchemeTask:`. On macOS 11, which ships an older WebKit without this fix, the task pointer passed to `stopURLSchemeTask:` during dealloc can already be freed. This isn't theoretical — it's reproducible on macOS 11.7.10 and matches the crash in #1733.

See also #1484 which fixed a closely related issue on the `start_task` side: *"The `webview` and `task` objects passed to `start_task` were not retained, leading to use-after-free issues in the responder."* My PR addresses the same class of problem on the `stop_task` path — which #1484 did not cover.

**Re: "The manual drop order looks really wrong"**

I agree. The explicit drop ordering is a hack and I'm not confident it's the right approach. To be honest, I'm not deeply familiar with objc2's reference semantics in Rust or the idiomatic way to handle this kind of platform-level lifetime issue. My fix was pragmatic (it stops the crash across macOS 11–26) but clearly not elegant.

If someone more experienced with Rust + macOS/objc2 could suggest a cleaner solution, I'd be happy to rework the PR accordingly. The core constraint is: on macOS 11, `stop_task` *will* receive freed pointers — we need to handle that somehow without triggering implicit retain/release.

src/wkwebview/class/url_scheme_handler.rs
```diff
@@ -286,12 +286,18 @@ extern "C" fn start_task(
...
-                if WEBVIEW_STATE.read().unwrap().contains_key(webview_id) {
+                let result = if WEBVIEW_STATE.read().unwrap().contains_key(webview_id) {
                   webview.remove_custom_task_key(task_key);
                   Ok(())
                 } else {
                   Err(crate::Error::CustomProtocolTaskInvalid)
-                }
+                };
+
+                // webview must drop before task: if webview drop triggers dealloc →
```
Legend-Master, line 296: Does this alone fix the problem?

src/wkwebview/class/url_scheme_handler.rs
```diff
@@ -334,8 +340,8 @@ extern "C" fn start_task(
 extern "C" fn stop_task(
   _this: &ProtocolObject<dyn WKURLSchemeHandler>,
   _sel: objc2::runtime::Sel,
-  webview: &WryWebView,
-  task: &ProtocolObject<dyn WKURLSchemeTask>,
+  _webview: *mut AnyObject,
+  _task: *mut AnyObject,
 ) {
-  webview.remove_custom_task_key(task.hash());
+  // no-op: avoid accessing task/webview — macOS 11 may pass freed pointers here
```
Legend-Master, line 346: I don't think we can simply removing this?

Legend-Master (commented):
Also thanks for the context, 

> FIXME: though we give it a static lifetime, it's not guaranteed to be valid.

is a leftover from #1484 and can be removed now

src/wkwebview/class/url_scheme_handler.rs
sftse: Is this change still necessary once we figure out what to do with `stop_task`?

I'm reading the explanation as, dropping webview is necessary because we call `stop_task` on the freed task, which the other hunk is supposed to address.

PR #1736, opened 2026-05-24
refactor!: remove unsafe impl Send and Sync on `NewWindowOpener`

sftse:
Partially superseding #1571, I'll take another look at that PR shortly, hadn't seen it no longer applies cleanly.

When we backtracked on the handler needing to be Send/Sync, `NewWindowOpener` can now also lose its unsafe override.

### DIFF
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -252,7 +252,10 @@ fn handle_request(
 
 #[allow(non_snake_case)]
 pub unsafe fn wryCreate(env: JNIEnv, _: JClass) {
-  let mut main_pipe = MainPipe { env };
+  let mut main_pipe = MainPipe {
+    env,
+    package: super::PACKAGE.get().unwrap(),
+  };
 
   let looper = ThreadLooper::for_thread().unwrap();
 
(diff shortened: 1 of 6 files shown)


src/android/binding.rs
```diff
@@ -252,15 +252,19 @@ fn handle_request(
...
+  };
 
   let looper = ThreadLooper::for_thread().unwrap();
 
   looper
     .add_fd_with_callback(MAIN_PIPE[0].as_fd(), FdEvent::INPUT, move |fd, _event| {
-      let size = std::mem::size_of::<bool>();
-      let mut wake = false;
-      if libc::read(fd.as_raw_fd(), &mut wake as *mut _ as *mut _, size) == size as libc::ssize_t {
+      let mut buf = [0];
```
Legend-Master: We should use an explicit type here, since all the casting below and can be a problem

```suggestion
      let mut buf = [0u8];
```

Legend-Master (dismissed):
Looks good in general, just a small nitpick

And you need to sign your commits

src/lib.rs
```diff
@@ -520,9 +520,6 @@ pub struct NewWindowOpener {
   pub target_configuration: Retained<objc2_web_kit::WKWebViewConfiguration>,
 }
 
-unsafe impl Send for NewWindowOpener {}
```
Legend-Master, line 523: Actually, this is a breaking change, let's add a change file and this can be scheduled to merge when we finish all the patch releases

Legend-Master (approved):
Will wait a bit before merging at the next minor window

PR #1737, opened 2026-05-25
Rescope unsafe fn blocks

sftse:
To narrow down the causes of the observed crashes, let's rescope the unsafe use to a smaller area.

Objc2-foundation made ~~most~~some of its bindings safe, so here is a first batch of changes.

Most of the diff is dumb indentation changes, split out as a fmt commit.

### DIFF
--- a/.changes/objc2-0.3.2.md
+++ b/.changes/objc2-0.3.2.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+On macOS, updated `objc2-*` dependencies to 0.3.2
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1776,9 +1776,9 @@ dependencies = [
 
 [[package]]
 name = "objc2-app-kit"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


FabianLars:
> Objc2-foundation made most of its bindings safe

iirc at least some of that (thinking of all objc2 crates) wasn't in 0.6.0 / 0.3.0 so we should probably increase the minimum patch version of the objc2 crates or check if `=0.3.0` still compiles

sftse:
`=0.3.0` would not compile. I think the logic of the change is that making methods safe is purely backwards-compatible.

I did a rough smoke-test to check what cargo does. If we do note the patch version explicitly in the `Cargo.toml`, as done in this PR, it seems to treat it as a hard `>=0.3.2` constraint, so dependency alignment means any downstream consumer of wry will update their locked version to `>=0.3.2`.

I think this update could only cause problems if we have a floating patch number like `0.3` in wry's `Cargo.toml`.

FabianLars:
ay sorry, the cargo.toml change didn't load on my phone. didn't notice that you already upped the version. Do any of the other objc2 crates need a bump as well?

sftse:
Let's try.

Legend-Master (approved):

PR #1738, opened 2026-05-25
feat(windows): add `with_profile_name` for WebView2 multi-profile support

### LINKED ISSUE #1321: Support for Browser Profiles (extension of Tauri issue #9285)
I would ideally like to be able to create separate browsing profiles to segment Cookies, localStorage, and IndexedDB data into different profiles. Here are some use cases which could benefit from this feature:

- Supporting multiple account login without any further server-side development
- Local-first applications like Excalidraw which typically stores user data within either localStorage or an IndexedDB table
- To limit third party tracking mechanisms across different webviews

The above use cases are very similar to the use cases for the `data_directory()` api. However, as I noted in the Alternatives considered section, the `data_directory()` api is not supported on MacOS.

## Platform-Specific Implementation Details

In macOS 14 and iOS 17 Apple has released new [DataStore APIs](https://webkit.org/blog/14423/building-profiles-with-new-webkit-api/) which should enable this feature.

In Windows Microsoft has released new [Multiple Profile Support](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/multi-profile-support?tabs=dotnetcsharp) which should enable this feature.

I assume we can polyfill this feature to WebkitGTK (and Android) using the existing implementation used for setting the `data_directory`. Ideally the polyfill would create additional subdirectories within the default data directory to support the different profile names. For private profiles (i.e. incognito windows) which shouldn't be persisted I propose the use of the tmp directory to store those temporary profiles. While we won't get the performance advantages listed below in the `Alternatives Considered`, a polyfill will at least allow some support in the interim before GTK and Android supports browser profiles.

### `WebviewProfile`

I propose the addition of a new struct called `WebviewProfile` which will store all of the necessary references and metadata needed for the profiles. It should have the following functions:

- `WebviewProfile::default() -> WebviewProfile` returns the default profile.
- `WebviewProfile::private() -> WebviewProfile` returns a temporary profile.
- `WebviewProfile::new(name: String) -> WebviewProfile` returns a new profile, storing the display name as metadata. It will not check whether the name has been used before as that would require access the the `App` state. 
- `WebviewProfile::filePath(&self) -> Option<Path>` returns `Some(directory)` where all of the user data is stored if it isn't private, otherwise returns `None`. Note that this option resolves the complaint (#8635) around trying to locate where the browser persists user data when using the default data_directory(), both in the default case, and in the case of a custom profile. I assume that we can hard-code platform-dependent paths manually to support this api.
- `WebviewProfile::display() -> String` returns the configured display name, `"private"`, or `"default"` depending on the type of profile. 
  - Should also implement `std::fmt` which should call the above function.

### Adding `WebviewProfile` to `WebContext`

I propose the addition of a `WebviewProfile` field to be passed into and stored within WebContext (in addition to data_directory), with various platform-specific handlings of what to do with that profile. That should be sufficient to provide the above struct to Tauri and by extension to the end user as an additional parameter for each Webview.

I initially considered using the `data_directory()` api to emulate different profiles across all platforms, similar to what I proposed for the WebkitGTK polyfill. Unfortunately, as @FabianLars points out in [Tauri#8637](https://github.com/tauri-apps/tauri/discussions/8637), that "We can't control the data dir on macos." 

Additionally, Microsoft justifies the addition of multiple profiles (over the existing data folder api) as follows:
> Previously, without multi-profile support, to achieve data separation, a WebView2 app could use different user data folders for different WebView2 controls. However, in that approach, you must run multiple WebView2 runtime instances (each including a browser process and a bunch of child processes), which consumed more system resources including memory, CPU footprint, and disk space. [source](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/multi-profile-support?tabs=dotnetcsharp#previous-approach-using-a-different-user-data-folder-for-each-webview2-control)

**Additional context**

This is an extension to my [Tauri issue #9285](https://github.com/tauri-apps/tauri/issues/9285), as @amrbashir suggested to first implement it here first.

jtbourke:
Adds `WebViewBuilderExtWindows::with_profile_name` so embedded webviews can opt into a named WebView2 profile, isolating cookies, storage, IndexedDB, and cache from the host app's default profile while sharing the same WebView2 environment/runtime.

## Motivation

A common use case for `wry` (and Tauri downstream) is hosting an in-app browser tab alongside the main app webview. Without profile isolation, both share the unnamed default profile under the same `user_data_folder`, which means anything a user does in the embedded browser (signing into github.com, storing cookies, etc.) is mixed with the host app's own `localStorage` state. Profile-wide operations like clearing browsing data become unsafe because they touch the host app's storage as collateral damage.

`with_data_directory` (separate `user_data_folder`) is the existing workaround but it creates an entirely separate WebView2 environment with its own runtime/networking stack — which has historically caused OAuth/login-page regressions in apps that tried it (downstream report: Glint #2665).

Microsoft's recommended multi-profile pattern (https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/multi-profile-support) is named profiles within a shared environment, which `ICoreWebView2ControllerOptions::SetProfileName` enables. This PR exposes it through wry.

## Changes

- New `profile_name: Option<String>` field on the Windows `PlatformSpecificWebViewAttributes`.
- New trait method on `WebViewBuilderExtWindows`: `with_profile_name<S: Into<String>>(self, name: S) -> Self`.
- In `create_controller`, when a profile name is configured, call `ICoreWebView2ControllerOptions::SetProfileName` before `CreateCoreWebView2ControllerWithOptions`.

When `with_profile_name` is not called the behavior is unchanged — the webview uses the unnamed default profile, same as today.

## Platform notes

This PR only touches Windows. macOS / iOS already have `with_data_store_identifier` for the equivalent functionality. A follow-up could add the Linux equivalent (a dedicated `WebsiteDataManager` directory) under a unified API name if there's appetite for that.

## Naming-rule notes

WebView2 enforces profile-name rules (alphanumeric + `.`, `_`, `-`, space; up to 64 chars; cannot start or end with `.` or space). The doc-comment on `with_profile_name` mentions these constraints. The patch doesn't validate the string client-side — `SetProfileName` returns an error for malformed names, which is surfaced as a `wry::Error` at build time.

## Related

- Closes #1321 (Support for Browser Profiles) — at least the Windows half.
- Tauri tracking issue: tauri-apps/tauri#9285.

### DIFF
--- a/src/webview2/mod.rs
+++ b/src/webview2/mod.rs
@@ -135,7 +135,13 @@ impl InnerWebView {
     } else {
       Self::create_environment(&attributes, pl_attrs.clone())?
     };
-    let controller = Self::create_controller(hwnd, &env, attributes.incognito, background_color)?;
+    let controller = Self::create_controller(
+      hwnd,
+      &env,
+      attributes.incognito,
+      background_color,
+      pl_attrs.profile_name.as_deref(),
+    )?;
     let webview = Self::init_webview(
       parent,
       hwnd,
@@ -369,6 +375,7 @@ impl InnerWebView {
     env: &ICoreWebView2Environment,
     incognito: bool,
     background_color: Option<(u8, u8, u8, u8)>,
+    profile_name: Option<&str>,
   ) -> Result<ICoreWebView2Controller> {
     let (tx, rx) = mpsc::channel();
 
@@ -405,6 +412,11 @@ impl InnerWebView {
         }
 
         controller_opts.SetIsInPrivateModeEnabled(incognito)?;
+
+        if let Some(name) = profile_name {
+          controller_opts.SetProfileName(&HSTRING::from(name))?;
+        }
+
         env10.CreateCoreWebView2ControllerWithOptions(hwnd, &controller_opts, &handler)?;
       } else {
         env.CreateCoreWebView2Controller(hwnd, &handler)?
--- a/.changes/windows-with-profile-name.md
+++ b/.changes/windows-with-profile-name.md
@@ -1,0 +1,5 @@
+---
+"wry": minor
+---
+
+On Windows, add `WebViewBuilderExtWindows::with_profile_name` to opt the webview into a named WebView2 profile. Webviews with different profile names within the same environment have isolated cookies, storage, IndexedDB, and cache while sharing the runtime — matching WebView2's documented multi-profile pattern.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1677,6 +1677,7 @@ pub(crate) struct PlatformSpecificWebViewAttributes {
   extension_path: Option<PathBuf>,
   default_context_menus: bool,
   environment: Option<ICoreWebView2Environment>,
+  profile_name: Option<String>,
 }
 
 #[cfg(windows)]
@@ -1692,6 +1693,7 @@ impl Default for PlatformSpecificWebViewAttributes {
       browser_extensions_enabled: false,
       extension_path: None,
       environment: None,
+      profile_name: None,
     }
   }
 }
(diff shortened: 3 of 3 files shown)


src/webview2/mod.rs
```diff
@@ -405,6 +412,12 @@ impl InnerWebView {
         }
 
         controller_opts.SetIsInPrivateModeEnabled(incognito)?;
+
+        if let Some(name) = profile_name {
+          let wide: Vec<u16> = name.encode_utf16().chain(std::iter::once(0)).collect();
```
Legend-Master: Let's use `HSTRING` here instead

Legend-Master (approved):
Thanks! Could you also add a change file?

> https://github.com/tauri-apps/wry/blob/dev/.changes/readme.md

Legend-Master (approved):
Thanks!

PR #1739, opened 2026-05-27
Revert moved package to MainPipe in #1736

Legend-Master:
This partially reverts commit 145b36fb5e3551851a9afbdf9c34154e2388f85d (#1736).

`wryCreate` is called before `android_setup` right now and the change was causing the app to crash


Legend-Master:
cc @sftse

sftse:
Unfortunate. Is there a way we can refactor the statics so the initialization order is clearer, and how best to cover this with a test?

Legend-Master:
I'm not entirely sure, it seems like the order will depends on the user of wry. I do really want to setup tests though just not sure how

PR #1746, opened 2026-06-05
ci: benchmark on all platforms

Legend-Master:
The execution time don't seem to be that stable across multiple runs, this would probably mainly just serve as an integration test and a binary size regression tracker

Companion PR in the `benchmark_results` repo: https://github.com/tauri-apps/benchmark_results/pull/5


bench/src/build_benchmark_jsons.rs
```diff
@@ -37,16 +53,30 @@ fn main() {
...
-      .to_str()
-      .expect("Something wrong with wry_recent"),
+    wry_recent,
     &serde_json::to_value(recent).expect("Unable to build final json (recent)"),
   )
-  .unwrap_or_else(|_| panic!("Unable to write {:?}", wry_recent));
+  .unwrap_or_else(|_| panic!("Unable to write {}", wry_recent.display()));
+}
+
+fn migrate_old_benchmark_data(new_wry_data: &Path, new_wry_recent: &Path) {
```
Legend-Master: We can also remove this code and rename the files in the https://github.com/tauri-apps/benchmark_results directly

FabianLars: whatever you prefer

FabianLars: actually yeah, renaming sounds cleaner if we can remove the code then

Legend-Master: Pushed the rename in https://github.com/tauri-apps/benchmark_results/pull/5/commits/8fa25ebf861162a2f2b111e39b0be54d72841483

FabianLars (dismissed):

FabianLars (approved):

PR #1753, opened 2026-06-17
fix(macos): avoid panics on missing URL (fix #1752)

### LINKED ISSUE #1752: macOS: avoid panic when WKWebView or navigation request URL is nil
### Description

On macOS, there are still a few WKWebView URL access paths that can panic if WebKit returns a nil URL or nil absoluteString. This is especially dangerous for applications built with `panic = "abort"`, where the panic terminates the whole app instead of being recoverable by the caller.

I found this while using `wry 0.55.1` through Tauri. The latest published `wry` version on crates.io is still `0.55.1`, and the affected unwraps are present in the published crate source.

### Affected code paths

- `src/wkwebview/mod.rs`, `url_from_webview()`:

```rust
let url_obj = webview.URL().unwrap();
let url = url_obj.absoluteString().unwrap();
```

- `src/wkwebview/navigation.rs`, `navigation_policy()`:

```rust
let uri = request.URL().unwrap().absoluteString().unwrap();
```

### Why this is a problem

`WKWebView.URL()` and `NSURL.absoluteString()` are Objective-C APIs that can return nil. The current code assumes both values are always present and calls `unwrap()`, so a transient nil value can trigger an `Option::unwrap()` panic.

In a release app configured with `panic = "abort"`, this becomes a process abort. Wrapping the caller side with `catch_unwind` is not a reliable mitigation in that configuration, so the WebKit nil case should be handled inside wry.

### Expected behavior

wry should not panic when WebKit returns a nil URL or nil absoluteString. Possible behavior:

- In `url_from_webview()`, return an error if `WKWebView.URL()` or `absoluteString()` is nil.
- In `navigation_policy()`, cancel or otherwise safely ignore the navigation when `request.URL()` or `absoluteString()` is nil.

### Suggested fix

Replace these `unwrap()` calls with explicit `Option` handling and propagate/cancel appropriately. For example, `url_from_webview()` can use `ok_or_else(...)`, while `navigation_policy()` can match on the optional URL string and cancel the navigation if unavailable.

### Environment

- Platform: macOS
- wry version: `0.55.1`
- Backend: WKWebView

Buncys:
## Summary
This PR removes `unwrap` calls in two code paths, as requested in #1752.

## Affected code paths
- `src/wkwebview/mod.rs`, `url_from_webview()`
- `src/wkwebview/navigation.rs`, `navigation_policy()`

## Changes
- `url_from_webview()` now returns `Ok("")` when URL is unavailable, which is consistent with other platforms
- Navigation will be canceled instead of panicking, when URL is unavailable in `navigation_policy()`

Fixes #1752

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -1346,8 +1346,9 @@ r#"Object.defineProperty(window, 'ipc', {
 }
 
 pub fn url_from_webview(webview: &WKWebView) -> Result<String> {
-  let url_obj = unsafe { webview.URL().unwrap() };
-  let absolute_url = url_obj.absoluteString().unwrap();
+  let absolute_url = unsafe { webview.URL() }
+    .and_then(|u| u.absoluteString())
+    .unwrap_or_default();
 
   let bytes = {
     let bytes: *const c_char = absolute_url.UTF8String();
--- a/src/wkwebview/navigation.rs
+++ b/src/wkwebview/navigation.rs
@@ -63,7 +63,10 @@ pub(crate) fn navigation_policy(
       false
     };
     let request = action.request();
-    let url = request.URL().unwrap().absoluteString().unwrap();
+    let url = request
+      .URL()
+      .and_then(|u| u.absoluteString())
+      .map(|s| s.to_string());
 
     if should_download {
       let has_download_handler = this.ivars().has_download_handler;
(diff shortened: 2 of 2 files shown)


sftse:
Fixes to non-trivial issues are nearly impossible to judge without a reproducer.

Buncys:
Here is the reproducer:
```rust
let window = WindowBuilder::new().build(&event_loop).unwrap();
let web_view = WebViewBuilder::new().build(&window).unwrap();
let _ = dbg!(web_view.url()); // <- panics on macos
```
Running this produces:
```
thread 'main' panicked at .../wry-0.55.1/src/wkwebview/mod.rs:1349:40:
called `Option::unwrap()` on a `None` value
```

PR #1764, opened 2026-07-09
docs: `with_drag_drop_handler` disables HTML drag APIs

Legend-Master:


FabianLars (dismissed):
iirc for some reason this currently happens on other platforms as well even though it doesn't technically have to

Legend-Master:
I could verify the Linux part, could you verify the macOS part?

FabianLars:
ye

Legend-Master:
Linux's HTML drag and drop still works with `with_drag_drop_handler` set to me in WSL.

FabianLars:
not working on macos

FabianLars (approved):

FabianLars:
ah shit it works on macOS - i tested it incorrectly. In Tauri we always return `true` from the drag_drop handler (why?) and i didn't switch that to false in the wry example... sryyy

FabianLars:
pushed docs update

PR #1772, opened 2026-07-17
fix: avoid panic on invalid IPC request URI

### LINKED ISSUE #1255: File uri causes panic in `attach_ipc_handler`
**Describe the bug**
I'm using a `file://` uri for the webview, and attach an IPC handler.  When I run the app it panics at
```
thread 'main' panicked at /home/.../.cargo/registry/src/index.crates.io-6f17d22bba15001f/wry-0.39.4/src/webkitgtk/mod.rs:531:16:
called `Result::unwrap()` on an `Err` value: http::Error(InvalidUri(InvalidFormat))
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/result.rs:1649:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/result.rs:1073:23
   4: wry::webkitgtk::InnerWebView::attach_ipc_handler::{{closure}}
             at /home/.../.cargo/registry/src/index.crates.io-6f17d22bba15001f/wry-0.39.4/src/webkitgtk/mod.rs:528:13
```

which is
```
            Request::builder()
              .uri(webview.uri().unwrap().to_string())
              .body(js.to_string())
              .unwrap(),
```

**Steps To Reproduce**

**Expected behavior**
1. I'd expect the request to be built successfully.  Seems like an `http` issue https://github.com/hyperium/http/issues/323
2. Failing that, I'd be happy with a fake/standin/override url.  I'm only using the body of the IPC request (how common is needing the URL?)

**Screenshots**

OS: Arch
Rustc: 1.76.0

**Additional context**

degant:
## What

The IPC handlers in the `webview2`, `webkitgtk`, and `android` backends build an `http::Request` from the webview's current document URL and `.unwrap()` the result, e.g.:

```rust
ipc_handler(Request::builder().uri(url).body(js).unwrap());
```

`Request::builder().uri(...).body(...)` validates the URI against the strict `http` crate rules. If the document URL is something a browser accepts but `http::Uri` rejects — for example a `file://` URL (whose empty authority yields `InvalidUri(InvalidFormat)`), or a URL containing a raw space or a non-ASCII/emoji character (`InvalidUri(InvalidUriChar)`) — the builder returns an `Err` and the `.unwrap()` aborts the whole process:

```
called `Result::unwrap()` on an `Err` value: http::Error(InvalidUri(InvalidFormat))
```

Because the URL comes from whatever the webview happens to have loaded, an application has no reliable way to prevent this from user-side, so it manifests as an intermittent hard crash.

The WKWebView backend already avoids this — its script-message handler builds the request and returns early when the builder fails instead of unwrapping (see `src/wkwebview/class/wry_web_view_delegate.rs`). This PR makes the other three backends behave the same way.

## How

In each IPC handler, build the request, log the error under the `tracing` feature, and only forward it to the IPC handler when it built successfully:

```rust
let request = Request::builder().uri(url).body(js);
#[cfg(feature = "tracing")]
if let Err(error) = &request {
  tracing::warn!("WebView received invalid IPC request: {error}");
}
if let Ok(request) = request {
  ipc_handler(request);
}
```

For `webkitgtk`, `webview.uri()` was also unwrapped; it is now `webview.uri().map(|u| u.to_string()).unwrap_or_default()`, so a missing URI yields an empty string (which fails URI validation and drops the message) instead of panicking.

## Scope / behavior

This targets the crash only. An IPC message whose document URL is valid in the browser but not a valid `http::Uri` is now dropped (and logged under `tracing`) instead of aborting the process — the same trade-off the WKWebView backend already makes. Behavior for valid URLs is unchanged.

I don't have a Rust toolchain on hand to run `cargo build`/`clippy` locally, so I've leaned on matching the existing WKWebView pattern and keeping the change minimal; happy to adjust to whatever CI/clippy prefers.

Closes #1255

### DIFF
--- a/.changes/fix-ipc-invalid-uri-panic.md
+++ b/.changes/fix-ipc-invalid-uri-panic.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Don't panic in the IPC handler when the webview's current document URL is not a valid `http::Uri` (for example a `file://` URL, or one containing a raw space or non-ASCII characters). Such invalid IPC requests are now logged (under the `tracing` feature) and dropped instead of aborting the process, matching the existing behavior of the WKWebView backend.
--- a/src/android/binding.rs
+++ b/src/android/binding.rs
@@ -418,7 +418,13 @@ pub unsafe fn ipc(mut env: JNIEnv, _: JClass, webview_id: JString, url: JString,
       let body = body.to_string_lossy().to_string();
       let webview_id = webview_id.to_string_lossy().to_string();
       if let Some(ipc) = IPC.lock().unwrap().get(&webview_id) {
-        (ipc.handler)(Request::builder().uri(url).body(body).unwrap())
+        match Request::builder().uri(url).body(body) {
+          Ok(request) => (ipc.handler)(request),
+          Err(_error) => {
+            #[cfg(feature = "tracing")]
+            tracing::warn!("WebView received invalid IPC request: {_error}")
+          }
+        }
       }
     }
     (Err(_e), _, _) | (_, Err(_e), _) | (_, _, Err(_e)) => {
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -732,12 +732,14 @@ impl InnerWebView {
 
       if let Some(js) = msg.js_value() {
         if let Some(ipc_handler) = &ipc_handler {
-          ipc_handler(
-            Request::builder()
-              .uri(webview.uri().unwrap().to_string())
-              .body(js.to_string())
-              .unwrap(),
-          );
+          let uri = webview.uri().map(|u| u.to_string()).unwrap_or_default();
+          match Request::builder().uri(uri).body(js.to_string()) {
+            Ok(request) => ipc_handler(request),
+            Err(_error) => {
+              #[cfg(feature = "tracing")]
+              tracing::warn!("WebView received invalid IPC request: {_error}")
+            }
+          }
         }
       }
     });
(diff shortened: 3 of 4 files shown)


degant:
Quick note on CI in case it's useful when reviewing: the one failing check (`test (x86_64-unknown-linux-gnu, --all-features)`) looks like an infra flake rather than anything in this PR. The job ran for ~46 min and then the runner dropped with "The hosted runner lost communication with the server…" partway through.

A few things that point to it not being the change itself:
- the same target passes on `--no-default-features`
- every other platform in the matrix is green
- `dev` is green, and the diff doesn't touch anything Linux-specific

Should be safe to re-run whenever someone gets a chance. Thanks! 🙏

Legend-Master (dismissed):
Thanks, just a bit of nitpicks, could we use a `match` instead of 2 `if`s?

degant:
Hey @Legend-Master — feedback's all addressed (`match` + signed commits) and CI's green. Your approval got auto-dismissed by the new commit, so it needs another look when you get a sec. Thanks!

Legend-Master (approved):

Legend-Master:
thanks

PR #1777, opened 2026-07-21
fix(webview2): don't panic on IPC from non-http(s) documents

gimenete:

- [x] Bug fix

## Summary

On Windows, the WebView2 `WebMessageReceived` handler builds an `http::Request` from the posting document's `Source` URL and unwraps the result:

```rust
ipc_handler(Request::builder().uri(url).body(js).unwrap());
```

When a message is posted from a document whose URL is not a valid `http::Uri` — for example a `file:`, `data:` or `blob:` document — `Request::builder().uri(url)` fails with `http::Error(InvalidUri(InvalidFormat))`. Because this runs inside a WebView2 callback that cannot unwind, the `.unwrap()` panic turns into a "panic in a function that cannot unwind" abort and takes the whole process down.

This is easy to hit for any app that hosts arbitrary or local content in a webview that also injects the wry IPC bootstrap: navigate to a `file://` page (or let an `http(s)` page open a `data:`/`blob:` document) and have it call `window.ipc.postMessage(...)`.

## Fix

Match on the `Request::builder(...).body(...)` result instead of unwrapping. On success the handler runs as before; on failure the message is dropped (and logged under the `tracing` feature) rather than aborting the process.

Reproduced: `Request::builder().uri("file:///C:/Users/me/index.html").body(())` returns `Err(InvalidUri(InvalidFormat))`, matching the observed crash.

### DIFF
--- a/.changes/webview2-ipc-invalid-uri.md
+++ b/.changes/webview2-ipc-invalid-uri.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+On Windows, fixed a panic in the WebView2 IPC handler when the posting document's `Source` URL is not a valid `http::Uri` (e.g. `file:`, `data:` or `blob:` documents). The failed `Request` build was unwrapped inside a callback that cannot unwind, aborting the process; the message is now dropped instead.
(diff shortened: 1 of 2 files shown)


Legend-Master:
There's already a PR #1772 about, thanks for contributing anyways.

PR #1779, opened 2026-07-22
feat(macos/ios): enable fullscreen without private api

FabianLars:
ref https://github.com/tauri-apps/tauri/pull/15640

tested on macos 26

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,12 +282,11 @@ Wry uses a set of feature flags to toggle several advanced features.
 - `drag-drop` (default): Enables [`WebViewBuilder::with_drag_drop_handler`] to control the behavior when there are files
   interacting with the window.
 - `devtools`: Enables devtools on release builds. Devtools are always enabled in debug builds.
-  On **macOS**, enabling devtools, requires calling private APIs so you should not enable this flag in release
-  build if your app needs to publish to App Store.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -282,12 +282,11 @@ Wry uses a set of feature flags to toggle several advanced features.
 - `drag-drop` (default): Enables [`WebViewBuilder::with_drag_drop_handler`] to control the behavior when there are files
   interacting with the window.
 - `devtools`: Enables devtools on release builds. Devtools are always enabled in debug builds.
-  On **macOS**, enabling devtools, requires calling private APIs so you should not enable this flag in release
-  build if your app needs to publish to App Store.
+  On **macOS**, enabling devtools, requires calling private functions, so avoid this in release builds if you publish your app on the App Store.
 - `transparent`: Transparent background on **macOS** requires calling private functions.
```
Legend-Master, line 286: Since we're at it, might as well rename those feature to something like `macos-private-api-transparent`

Legend-Master (dismissed):

Legend-Master (approved):

README.md
FabianLars: still considering to remove both flags - always using the private transparent api and removing the private fullscreen api...

will do a little bit more research and rename or remove in another pr

PR #1668, opened 2026-02-10
fix(macOS）：Fix the crash issue on macOS 10.13 caused by the missing NSHTTPCookie symbol and the invocation of drawsBackground in WKWebViewConfiguration.

zeta-gundam:
related issue：https://github.com/tauri-apps/tauri/issues/14201

The branch pending for merge previously: https://github.com/tauri-apps/wry/tree/fix/macos/symbol-not-found-1013
As this branch dates back too long, I have re-implemented the corresponding modifications on a new branch, and also fixed additional new issues that were identified subsequently.
To date, all changes have passed testing on my macOS 10.13.6 environment.

closes #1618

PR #1780, opened 2026-07-22
refactor(macos/ios): remove fullscreen and transparent feature flags

FabianLars:
Sooo, looking around on the internet (eg stackoverflow), people have been using these on app store builds for ages, including a few in tauri land.
Speaking of Tauri land, `allowsPictureInPictureMediaPlayback` has been unguarded since it was added and it did not cause issues.

Note, that we can only assume app store compatibility for when we set property values like you can see in the diff (eg via `setValue_forKey` ) - actually using private objc apis **will** cause issues!!

~~draft because done on windows and not tried on macos yet~~ tested on macos 12.4 and 26


src/wkwebview/mod.rs
```diff
@@ -367,16 +368,10 @@ impl InnerWebView {
 
       let version = util::operating_system_version();
 
-      #[cfg(feature = "transparent")]
       if attributes.transparent || attributes.background_color.is_some() {
         let no = NSNumber::numberWithBool(false);
-        {
-          if cfg!(target_os = "ios") || version.0 > 10 || (version.0 == 10 && version.1 >= 14) {
```
FabianLars, line 374: I never saw a crashlog ([from that single user](https://github.com/tauri-apps/wry/pull/1668)) that made the version check required here for macOS. None of the other kvc uses ever caused version specific crashes.

README.md
```diff
@@ -283,10 +283,6 @@ Wry uses a set of feature flags to toggle several advanced features.
   interacting with the window.
 - `devtools`: Enables devtools on release builds. Devtools are always enabled in debug builds.
   On **macOS**, enabling devtools, requires calling private functions, so avoid this in release builds if you publish your app on the App Store.
-- `transparent`: Transparent background on **macOS** requires calling private functions.
```
pewsheen, line 286: Should also remove from `src/lib.rs` and `examples/transparent.rs:35`

src/wkwebview/mod.rs
pewsheen: hmm.. let's give it a shot

README.md
FabianLars: whoops, good catch

pewsheen (dismissed):

Legend-Master (approved):
Have no means to test this, just an approval on the idea

PR #1783, opened 2026-07-23
fix(webview2): honor WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS via the API on Windows (survives Runtime 150 elevated-host hardening)

### LINKED ISSUE #1782: [bug] WebView2 Runtime 150 hardening breaks WebDriver automation on elevated hosts (--remote-debugging-port dropped)
### Summary

WebView2 **Runtime 150** added a security hardening: when the host process is **elevated** (high-integrity / admin), WebView2 ignores the `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS` environment variable, HKCU policy, and command-line switch overrides. Only **HKLM policy** and arguments passed via the **WebView2 API** (`ICoreWebView2EnvironmentOptions::AdditionalBrowserArguments`) are honored. This is by-design on Microsoft's side — see [MicrosoftEdge/WebView2Feedback#5645](https://github.com/MicrosoftEdge/WebView2Feedback/issues/5645) (also #5640, #5644).

### Impact on wry

WebDriver automation of wry/Tauri apps on Windows goes through **msedgedriver**, which enables CDP by injecting `--remote-debugging-port=<port>` via the `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS` environment variable. On an **elevated** host running Runtime **≥150**, that env var is dropped, so the DevTools/CDP endpoint never opens and **session creation times out**:

```
WebDriverError: The operation was aborted due to timeout when running ".../session" with method "POST"
```

This reproduces **100% on GitHub-hosted Windows runners** (which run as administrator) and affects the official `tauri-driver` provider (and any msedgedriver-based automation). It does not reproduce for a normal, non-elevated user.

`wry` never reads `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS` itself — it relies on the runtime to apply it — and on Windows `set_allows_automation` is a no-op (only the WebKitGTK backend consumes it), so nothing on the wry side currently survives the hardening.

### Evidence

- Runtime **149** works, **150** fails — at the exact 149→150 boundary.
- Pinning a fixed-version **149** runtime via `WEBVIEW2_BROWSER_EXECUTABLE_FOLDER` restores it: the app loads 149, the port opens, and msedgedriver connects. This confirms (a) the hardening is specific to 150+, and (b) `WEBVIEW2_BROWSER_EXECUTABLE_FOLDER` itself *does* survive elevation. But pinning is only a stopgap — fixed-version CABs are published for the latest + second-latest major only, and it forces a version-matched msedgedriver.

Since the runtime behavior is by-design, no upstream runtime fix is coming.

### Environment

- wry `dev`
- Windows 11, WebView2 Runtime 150.0.4078.x
- Host process elevated (admin) — e.g. GitHub-hosted `windows-latest` runners

goosewobbler:
Closes #1782.

## Problem

WebView2 Runtime 150 hardened elevated (high-integrity) host processes to ignore `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS`, HKCU policy, and command-line switch overrides — only HKLM policy and args passed via the WebView2 **API** (`ICoreWebView2EnvironmentOptions::AdditionalBrowserArguments`) survive. This is by-design, see [MicrosoftEdge/WebView2Feedback#5645](https://github.com/MicrosoftEdge/WebView2Feedback/issues/5645).

WebDriver automation of wry/Tauri apps relies on msedgedriver injecting `--remote-debugging-port` through that env var. On an elevated host — e.g. GitHub-hosted Windows runners, which run as admin — with Runtime ≥150, the env var is dropped, the CDP/DevTools endpoint never opens, and session creation times out. `wry` never read the env var itself, and `set_allows_automation` is a no-op on Windows (only the WebKitGTK backend consumes it), so nothing on the wry side currently survives the hardening.

## Fix

`create_environment` already passes `AdditionalBrowserArguments` through the API. Fold `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS` into that call so the args survive elevation.

## Design note — unconditional vs. gated

This is the **unconditional** variant: it folds the env var in whenever it's set (effectively gated by the env var's *presence*, which is rare outside automation). It's minimal and self-contained — no new plumbing — and it fixes the regression for *any* elevated use of the env var, not only automation.

Trade-off: when the host is **not** elevated the runtime also reads the env var, so the args may be applied twice — harmless for idempotent flags like `--remote-debugging-port`, and moot if the runtime treats the env var as overriding the API property.

Happy to switch to a variant **gated on `allows_automation`** if you'd prefer to avoid any behavior change for non-automation apps. That needs the automation flag plumbed from `WebContext` into `create_environment`, which is why I defaulted to the smaller change.

## Testing

`src/webview2/mod.rs` is `#[cfg(windows)]`, so this needs Windows CI to compile-check. Behaviorally it's exercised end-to-end by tauri-driver-based WebDriver E2E on an elevated Windows host (where the regression surfaced). Included a covector `patch` changefile.

### DIFF
--- a/.changes/webview2-elevated-browser-args.md
+++ b/.changes/webview2-elevated-browser-args.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+On Windows, fold the `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS` environment variable into the args passed through the WebView2 API so it survives WebView2 Runtime 150's elevated-host hardening. This restores WebDriver automation (e.g. msedgedriver's `--remote-debugging-port`) for apps launched by an elevated host such as CI runners.
--- a/src/webview2/mod.rs
+++ b/src/webview2/mod.rs
@@ -297,7 +297,7 @@ impl InnerWebView {
       .map(HSTRING::from);
 
     // additional browser args
-    let additional_browser_args = pl_attrs.additional_browser_args.unwrap_or_else(|| {
+    let mut additional_browser_args = pl_attrs.additional_browser_args.unwrap_or_else(|| {
       // remove "mini menu" - See https://github.com/tauri-apps/wry/issues/535
       // and "smart screen" - See https://github.com/tauri-apps/tauri/issues/1345
       let default_args = "--disable-features=msWebOOUI,msPdfOOUI,msSmartScreenProtection";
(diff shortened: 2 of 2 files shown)


FabianLars:
I'd say upstream's decision to not read the reg / env in an elevated context is absolutely the right call and we should not blindly revert this.

Not sure what the best alternative is, maybe an `automation` cargo feature flag? I think we could change wry's linux impl to switch to the feature flag instead of a setter on the context too - though that probably has to wait until v3 and it's also problematic that only one context can have automation enabled 🤔

goosewobbler:
Thanks for the feedback @FabianLars, I'll close this then. Our embedded driver isn't blocked on any solution here but happy to take a crack at a feature flag (or other) approach when the time is right - I left #1782 open but trimmed off the implementation proposal.

PR #1781, opened 2026-07-22
fix(macos): dont register unknown delegate methods in debug mode

FabianLars:
Alternative to https://github.com/tauri-apps/wry/pull/1776

closes https://github.com/tauri-apps/tauri/issues/11496
closes #1401

PR #1776, opened 2026-07-21
build: conditionally compile media capture permission for macOS deplo…

JakiChen:
## Motivation

When building `wry` with a macOS deployment target lower than 12.0 (e.g. 10.15 or 11.0), compilation fails because the delegate method `requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:` is only available on macOS 12.0+. This forces downstream users to maintain forks or manually patch `wry`.

## Solution

- In `build.rs`, read the `MACOSX_DEPLOYMENT_TARGET` environment variable and set custom `cfg` flags `macos_12_or_later` / `macos_11_or_earlier`.
- Guard the entire `requestMediaCapturePermissionForOrigin` method with `#[cfg(macos_12_or_later)]`.
- On older deployment targets the method is excluded, so the system default permission flow takes over (no functional loss).

The recently added custom `permission_handler` logic is fully preserved when compiling for macOS ≥ 12.

## Testing

- Built a Tauri 2 project using this branch with `MACOSX_DEPLOYMENT_TARGET=11.0`: **compilation succeeds** (previously failed).
- Built with `MACOSX_DEPLOYMENT_TARGET=12.0`: **compilation succeeds** and the permission handler works as expected.

## Impact

- No behavioral change for macOS 12+ builds.
- Removes the need for downstream patches when targeting older macOS versions.

PR #1202, opened 2024-03-27
refactor: migrate to `dpi` crate

amrbashir:
closes #1172
closes tauri-apps/tauri#9264

PR #1786, opened 2026-07-27
docs: all features

Legend-Master:


src/lib.rs
```diff
@@ -301,13 +301,16 @@
 //!   for the crate to work. This feature was added in preparation of other ports like cef and servo.
 //! - `protocol` (default): Enables [`WebViewBuilder::with_custom_protocol`] to define custom URL scheme for handling tasks like
 //!   loading assets.
+//! - `x11` (default): Enables x11 support and dependencies on Linux.
+//! - `serde`: Enables `dpi`'s `serde` feature.
```
Legend-Master, line 305: This was from #1202 but I'm not really sure the use of it, maybe for enabling `serde` on the returned positions for the caller?

src/lib.rs
```diff
@@ -301,13 +301,16 @@
...
 //! - `devtools`: Enables devtools on release builds. Devtools are always enabled in debug builds.
 //!   On **macOS**, enabling devtools, requires calling private APIs so you should not enable this flag in release
 //!   build if your app needs to publish to App Store.
 //! - `transparent`: Transparent background on **macOS** requires calling private functions.
 //!   Avoid this in release build if your app needs to publish to App Store.
 //! - `fullscreen`: Fullscreen video and other media on **macOS** requires calling private functions.
 //!   Avoid this in release build if your app needs to publish to App Store.
+//! - `mac-proxy`: Enables `WebViewBuilder::with_proxy_config` on macOS.
```
Legend-Master, line 309: This is from #1006, it's not super clear to me why we needed this in a feature flag for macOS?

From the description in the original PR, it seems like it's only available on later version of macOS but I think that can be gated in a version check?

src/lib.rs
FabianLars: yeah, i think so. maybe it's even used in tauri to send it to the js side? can't remember. but for tao consumers it's probably rarely helpful to enable.

src/lib.rs
```diff
@@ -301,9 +301,12 @@
 //!   for the crate to work. This feature was added in preparation of other ports like cef and servo.
 //! - `protocol` (default): Enables [`WebViewBuilder::with_custom_protocol`] to define custom URL scheme for handling tasks like
```
Legend-Master, line 302: Actually wonder if we should remove this as well that nothing except the builder methods are still behind the flag

src/lib.rs
FabianLars: https://github.com/tauri-apps/wry/pull/1006#discussion_r1295328097 / https://github.com/tauri-apps/wry/pull/1006/changes#diff-182c0a218443303957d1585b15478976c27894b9104c8caa2527f41f357b7796

I think it could still be an issue nowadays after the objc2 migration because the whole framework was added in 10.14 while we still support 10.13 so i think it could cause build issues (not runtime issues after adding version checks) on 10.13

src/lib.rs
FabianLars: Was there ever a reason for this?

src/lib.rs
Legend-Master: Will something you tried in #1781 work with this as well?

FabianLars: ~~I'm not sure, i don't think so because we do have to link the framework at least - at least if i look at objc2-generated it seems like it.~~

~~I think in this case we'd need https://github.com/tauri-apps/wry/pull/1776 - or we just drop support for compiling wry on 10.13, running apps should work regardless.~~
The fck, i forgot what my PR was. Yes it may work


Or i'll try to get a 10.13 vm running again, which i need a bit of time for, and test how things go.

FabianLars: > The fck, i forgot what my PR was. Yes it may work

But without the 0.7 annotation, this here would stay until we drop 10.13 support.

FabianLars (approved):

PR #1792, opened 2026-07-29
refactor(android): reomve some unused variables

Legend-Master:
This works for me from a quick test in tauri.

I honest don't really get our Android multi-window which is just multiple activities inside a single window and even not visible(stacked) on smaller screens like a phone 😂


src/android/mod.rs
```diff
@@ -134,13 +132,6 @@ pub unsafe fn android_setup(
   let window_manager = env.new_global_ref(window_manager).unwrap();
 
   register_activity_proxy(vm, activity_id, activity, window_manager);
-
-  if let Some(webview_attributes) = WEBVIEW_ATTRIBUTES.lock().unwrap().get(&activity_id) {
```
Legend-Master, line 142: This doesn't make sense to me, the new window comes from `InnerWebView::new` which already sends the `CreateWebview` message.

(This runs before the mobile entry point in first activity so nothing can be here, the following creations must be from `InnerWebView::new`)

FabianLars:
> I honest don't really get our Android multi-window which is just multiple activities inside a single window and even not visible(stacked) on smaller screens like a phone

The implementation was done more or less according to customer requirements. I don't see much of a use case for it on phones (maybe temporarily load another site with an easy way to go back to the main thing) but afaik on tablets you're able to have multiple activities open side by side with this implementation but tbh never tried it on a tablet.

But yeah, not to be confused with multi webview like behavior

Legend-Master:
Activity embedding is only used for legacy multi-activity apps which we were not.

> Note: Modern android development (MAD) uses a single-activity architecture based on Jetpack APIs, including Jetpack Compose. Activity embedding is designed for multiple-activity, legacy apps that can't be easily updated to MAD. Create new apps using MAD. Update your legacy apps to MAD whenever possible.
> https://developer.android.com/guide/topics/large-screens/activity-embedding

It shows side by side on a tablet but that can be achieved in a single activity/webview so I'm not sure what's the point 😂

FabianLars:
okay so what i had in mind was an iPad thing, thought that'd work on android as well but nope.

> It shows side by side on a tablet but that can be achieved in a single activity/webview so I'm not sure what's the point 😂

maybe to have the same-ish api on android and ios?

Legend-Master:
I feel like it just made everything way harder without doing any good, but we can't remove it now 😂 I tried to search about this but it seems like on Android, true multi-window (like 2 different windows on desktop mode) within a single instance is not possible. (can be done through multi-instances but that's nothing in tauri stops that today I believe)

> maybe to have the same-ish api on android and ios?

It doesn't really work in a similar way, so idk.

FabianLars:
it was a bit time sensitive, maybe we can rethink it in v3 then

Legend-Master:
I see why this is a thing now... We don't always destroy the rust window when `onDestroy` if it's caused by a configuration change. This is used to re-create the webview to match in that case...

PR #1795, opened 2026-07-30
fix(windows): `RemoveWindowSubclass` first before dropping `ICoreWebView2Controller`

### LINKED ISSUE #1794: Windows parent subclass dereferences null refdata during teardown reentrancy
## Affected versions

Confirmed in Wry 0.55.1 and still present in 0.56.0 on Windows with WebView2.

## Crash

A production Tauri 2.11.2 application reported:

```text
EXCEPTION_ACCESS_VIOLATION_READ / 0x0
app                  0x7ff74e82cbf2
ntdll                RtlFreeHeap
comctl32             CallNextSubclassProc
comctl32             MasterSubclassProc
...
comctl32             DefSubclassProc
app                  0x7ff74e82cc39
```

The ASLR load base was `0x7ff74e010000`. Disassembly maps:

- `0x7ff74e82cbf2` to RVA `0x81cbf2`, the first controller pointer dereference in `parent_subclass_proc` for `WM_SETFOCUS | WM_ENTERSIZEMOVE`.
- `0x7ff74e82cc39` to RVA `0x81cc39`, immediately after the callback calls `DefSubclassProc`.

The fault address is zero, so the callback was re-entered with `dwrefdata == 0` while the subclass was still installed.

## Root cause

The destroy branch drops the boxed controller, then updates subclass refdata to null, and finally calls `DefSubclassProc`. `dettach_parent_subclass` removes the subclass only after the synchronous destroy message returns. Nested window dispatch during that interval can therefore enter controller-using branches with null refdata. Those branches cast and dereference it without a guard.

The current order also leaves a reentrancy window during controller drop before null is published.

## Proposed fix

1. At callback entry, delegate directly to `DefSubclassProc` when `dwrefdata == 0`.
2. In the destroy branch, update subclass refdata to null before dropping the controller.
3. Drop only when the null update succeeds; retaining one controller clone is safer than publishing a dangling pointer if the Win32 update fails.

## Regression test

A deterministic Windows test can create a hidden `STATIC` window, install an observer subclass, install the Wry parent subclass with null refdata, and synchronously send `WM_SIZE`, `WM_SETFOCUS`, `WM_ENTERSIZEMOVE`, `WM_MOVE`, and `WM_MOVING`. Each message must reach the observer through `DefSubclassProc` without terminating the process.

I am preparing a PR with this patch and the native Win32 regression test.

qwerfunch:
Closes #1794.

During parent-window teardown, Wry drops the WebView2 controller and only then clears the subclass reference data. Releasing the COM controller can synchronously dispatch a nested window message, which re-enters `parent_subclass_proc` and dereferences stale or null refdata.

This change:
- clears the subclass refdata before dropping the controller;
- delegates immediately through `DefSubclassProc` when refdata is already cleared;
- retains the controller clone if `SetWindowSubclass` fails, avoiding a dangling pointer; and
- adds a native hidden-window regression test covering `WM_SIZE`, `WM_SETFOCUS`, `WM_ENTERSIZEMOVE`, `WM_MOVE`, and `WM_MOVING`.

Validation:
- `cargo fmt -- --check`
- `cargo check --tests --target x86_64-pc-windows-msvc --locked --all-features`


qwerfunch:
The initial Windows `cargo test` and a separate exact native Windows run both passed. The first upstream Miri step failed only because Miri cannot call `CreateWindowExW`; commit d374efa adds `cfg_attr(miri, ignore)` while keeping the test active in normal Windows runs. Could the updated Actions workflows be approved when convenient?

src/webview2/mod.rs
```diff
@@ -1326,17 +1336,19 @@ impl InnerWebView {
       }
 
       msg if msg == WM_DESTROY || msg == PARENT_DESTROY_MESSAGE => {
-        // check if `dwrefdata` is null to avoid double-freeing the controller
-        if !(dwrefdata as *mut ()).is_null() {
+        // Publish the cleared pointer before dropping the controller. Dropping
+        // COM objects can synchronously dispatch another parent-window message.
+        // If updating the subclass fails, retain the clone rather than leave a
+        // dangling pointer that a later dispatch could dereference.
+        if SetWindowSubclass(
```
Legend-Master: We should drop the controller not matter if this succeed

src/webview2/mod.rs
```diff
@@ -1284,6 +1287,13 @@ impl InnerWebView {
     _uidsubclass: usize,
     dwrefdata: usize,
   ) -> LRESULT {
+    // The controller refdata is cleared before it is dropped during teardown.
+    // A nested window message can re-enter this callback in that interval, so
+    // delegate immediately instead of dereferencing the cleared pointer.
+    if dwrefdata == 0 {
+      return DefSubclassProc(hwnd, msg, wparam, lparam);
```
Legend-Master: This should not be able to happen with the change below right?

src/webview2/tests.rs
```diff
@@ -0,0 +1,156 @@
+// Copyright 2020-2023 Tauri Programme within The Commons Conservancy
```
Legend-Master: Remove this, the test looks very non-sensible

Legend-Master (changes requested):

src/webview2/mod.rs
Legend-Master: Actually, I thought it was `RemoveWindowSubclass`, why not `RemoveWindowSubclass` here that we will no longer be doing anything here?

Legend-Master (dismissed):

Legend-Master (approved):

Issue #1153, opened 2024-01-26
Community Verification of Tauri & Servo Integration


wusyong:
We currently have a [branch](https://github.com/tauri-apps/wry/tree/servo-wry-demo) for experimenting servo integration. Servo also has a [blog post](https://servo.org/blog/2024/01/19/embedding-update/) about it.
If anyone is interested, please help us test if it works well on each platform.

Here are prerequisites for each platform:

### Windows

- Install [scoop](https://scoop.sh/) and then install other tools from it. You can install them manually but this is easier:

```sh
scoop install git python llvm cmake curl
```

### MacOS

- Install [Xcode](https://developer.apple.com/xcode/)
- Install [Homebrew](https://brew.sh/) and then run:

```sh
brew install cmake pkg-config
```

### Linux

For now, please follow the instructions in [Servo - Build Setup](https://github.com/servo/servo) to bootstrap first.
And it only works on x11 at the moment. Wayland or Xwayland won't work.

### Run Demo

```
cargo r --example servo
```

tensor-programming:
Builds and runs on Windows 11 build 2263

![image](https://github.com/tauri-apps/wry/assets/22513880/444cc7c5-7c92-4f55-a375-f0817a14ae2b)

Will also check WSL2 ubuntu tomorrow.

edit: the servo instructions do not mention installing cmake and pkconfig. Tried WSL but am still having errors with pkg-config not finding the font lib.  Will mess with it some more and get back.

daniel-mader:
I'm a developer [@impierce](https://github.com/impierce) and I can confirm it works on macOS 14, `rustc 1.75.0`

Norlock:
It doesn't work on my machine, I use Arch btw ;). 

Running on Wayland (sway).

```shell
Finished dev [unoptimized + debuginfo] target(s) in 9m 21s
     Running `target/debug/examples/servo`
servo: ../wayland-1.22.0/src/wayland-client.c:2377: wl_proxy_set_queue: Assertion `proxy->display == queue->display' failed.
fish: Job 1, 'cargo r --example servo' terminated by signal SIGABRT (Abort)
```

adriantombu:
![Screenshot 2024-01-26 at 11 56 25](https://github.com/tauri-apps/wry/assets/383297/943b04dc-a823-4943-a0c9-b2e0eb9165d8)

The branch `servo-wry-demo` Works on my machine™ running on `macOs 13.6` with `rustc 1.74.1`, nice work!

[...]

PR #1797, opened 2026-08-02
Feat: Servo wry demo

jonathanKingston:
This is a refactoring of the work from: https://servo.org/blog/2024/01/19/embedding-update/ this brings it up to date with the many changes in Servo and worked on in #1153. The original Servo backend work by: @wusyong, @pewsheen, @delan, @rhysd

It rebases Wry onto current dev, updates the backend to current Servo main, adopts the latest embedding/webview-handle APIs, integrates with Tao-owned windows, and fixes input, HiDPI, Tauri hooks, tests, and CI.

I've validated this against a large UI I'm building and the only issues were in servo support of unrelated things: contenteditable, top level await in module workers, lack of has() in CSS. The performance seems to come close to Electron although I didn't look to hard at this. I've marked it as draft as I'm unsure you'd accept this work, I have tested it enough that it feels good enough to land. A corresponding Tauri [branch is here](https://github.com/tauri-apps/tauri/compare/dev...jonathanKingston:tauri:codex/servo-backend?expand=1) which is what I was using to validate.


FabianLars:
very interesting PR! So we're not 100% set on a decision yet but lately we followed the plan of keeping Wry a _system_ webview thing and have the different runtimes on the tauri layer instead (see tauri-runtime-wry, tauri-runtime-cef, tauri-runtime-verso)

If my assumption is correct and servo does not need external runtime files like the others then i wouldn't mind rethinking that approach for servo in particular but would probably start with tauri-runtime-servo anyway. btw, not asking _you_ to do it, i think this PR is already a huge help for anyone picking that up.

jonathanKingston:
@FabianLars I tried moving this over to tauri as a crate of it's own: https://github.com/tauri-apps/tauri/compare/dev...jonathanKingston:tauri:copse/servo-backend?expand=1 this does load the hello world, I've not verified it much.

`crates/tauri-runtime-servo/src/lib.rs` is essentially the other crate rewired; I think there's perhaps 200 lines of code that can be removed but that's still perhaps 5k that is copied almost verbatim from the wry runtime.

Which direction should I continue in?

eveeifyeve:
> very interesting PR! So we're not 100% set on a decision yet but lately we followed the plan of keeping Wry a _system_ webview thing and have the different runtimes on the tauri layer instead (see tauri-runtime-wry, tauri-runtime-cef, tauri-runtime-verso)
> 
> If my assumption is correct and servo does not need external runtime files like the others then i wouldn't mind rethinking that approach for servo in particular but would probably start with tauri-runtime-servo anyway. btw, not asking _you_ to do it, i think this PR is already a huge help for anyone picking that up.

I think using servo instead of verso (which is now officially deprecated) is the right start in the direction to getting servo support. As verso was implemented as a hacky solution for now*, but now we have that mostly upstream in servo..

FabianLars:
> Which direction should I continue in?

Well, i think tauri-runtime-servo is the best direction for now as it's also in line with tauri-runtime-cef and the outdated tauri-runtime-verso.

wdyt @Legend-Master @amrbashir @lucasfernog

Legend-Master:
From just the structure point of view, either one should be fine to me, but speaking from the maintenance point of view, I would say an out-of-tree `tauri-runtime-servo` since we're so short staffed that maintaining a fast moving backend like Servo is infeasible for us right now.

eveeifyeve:
> From just the structure point of view, either one should be fine to me, but speaking from the maintenance point of view, I would say an out-of-tree `tauri-runtime-servo` since we're so short staffed that maintaining a fast moving backend like Servo is infeasible for us right now.

The opposing argument is, should we do it tauri-runtime-cef as well and move it to the cef-rs repo? I don't know where you draw the line of it can live out of tree or in-tree, because some might include other browser engines like ladybird.

FabianLars:
Even if the initial integration is contributed there's still the fact that we actually have no capacity to maintain it.
Together with the insane lock file diff a separate repo just for the runtime, similar to how we had it with tauri-runtime-verso makes sense to me. Would also allow independent updates from tauri's release process.
Edit: oh and another important point, access management. We could give eg you permissions to the servo runtime repo without giving access to main tauri, which lowers the bar of entry in that sense.
Doesn't have to be permanent, we could move it in and out as we see fit (i'm not taking positions here, don't really care either way)

Either way, the main cef-rs repo is not the right place for this, that's supposed to be uncoupled from Tauri, similar to the webkit2gtk-rs repo for example.

jonathanKingston:
I moved this to: https://github.com/copse-dev/tauri-runtime-servo/ for now. Happy to migrate it over to this org, whatever works :).

The repo includes a bunch of fixes that has made Servo pretty comparable to my Electron build, they're opt in. I'll try and get them upstreamed somehow.

eveeifyeve:
> I moved this to: [copse-dev/tauri-runtime-servo](https://github.com/copse-dev/tauri-runtime-servo/?rgh-link-date=2026-08-24T23%3A39%3A14Z) for now. Happy to migrate it over to this org, whatever works :).

I wonder if the servo org would like to have the servo runtime. 🤔

jonathanKingston:
Not sure, they'd be welcome to have it. I don't intend to typosquat or anything.

PR #1801, opened 2026-08-03
docs(android): why `WEBVIEW_ATTRIBUTES` exists

Legend-Master:
Closes #1792

To be honest, I don't even think we should support this use case, modern Android apps should declare `android:configChanges` so the activity is never re-created.

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -139,6 +139,8 @@ pub unsafe fn android_setup(
 
   register_activity_proxy(vm, activity_id, activity, window_manager);
 
+  // We don't always destroy the rust window when `onDestroy` if it's caused by a configuration change.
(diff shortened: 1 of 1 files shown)


FabianLars:
What kind of config are we talking about here?

Legend-Master:
> What kind of config are we talking about here?

> - App display size
> - Screen orientation
> - Font size and weight
> - Locale
> - Dark mode versus light mode
> - Keyboard availability
> 
> https://developer.android.com/guide/topics/resources/runtime-changes

Seems like maybe we still need this even with all the correct `android:configChanges` declared, that the docs is saying `but aren't limited to` 😂

FabianLars (approved):

PR #1804, opened 2026-08-05
fix(android): override getDefaultVideoPoster to suppress placeholder

### LINKED ISSUE #1802: [Android] Gray play-button overlay on `<video>`, and no way to override `getDefaultVideoPoster`
On Android, a `<video>` with no `poster` attribute gets Chromium's built-in gray placeholder painted over it until the first frame decodes: the circle-with-a-play-triangle. Chromium scales it to the element box, so on anything that isn't small and square it looks stretched.

Image

The cause is that `WebChromeClient.getDefaultVideoPoster()` returns null by default and `RustWebChromeClient` doesn't override it. Returning any non-null bitmap makes it go away.

The reason I'm filing rather than just fixing it in my own app is that there's no supported way to do that:

- `RustWebChromeClient` isn't `open`, so it can't be subclassed.
- It has no `{{class-extension}}` placeholder, unlike `WryActivity`, `RustWebView`, `RustWebViewClient` and `Ipc`, so `WRY_RUSTWEBCHROMECLIENT_CLASS_EXTENSION` is silently a no-op.
- `onWebViewCreate` runs before `setWebChromeClient` in `main_pipe.rs`, so `webView.webChromeClient` is still null there.

`on_webview_created` does fire late enough and works if you're using wry directly, but it isn't available under Tauri, which claims that slot on Android to drive `PluginManager.onWebViewCreated`.

That leaves swapping in a hand-written `WebChromeClient` that delegates every method back to wry's. It works today, but it fails quietly later: any override added to `RustWebChromeClient` after the fact falls through to the platform default instead of wry's version, so the wrapper has to be re-checked on every upgrade.

## Recommendation

```kotlin
import android.graphics.Bitmap

override fun getDefaultVideoPoster(): Bitmap {
  return Bitmap.createBitmap(50, 50, Bitmap.Config.ARGB_8888)
}
```

react-native-webview does exactly this, unconditionally, on both of its chrome-client paths (`RNCWebViewManager.kt`). Capacitor's `BridgeWebChromeClient`, which this file was originally based on, doesn't.

I'd argue for unconditional rather than a builder flag. What it replaces is a stretched placeholder nobody is picking on purpose, so an opt-in would just mean every app that plays video has to go find the flag to get the only result anyone wants.

Checked against 0.55.1 and `dev`.

velocitysystems:
## What kind of change does this PR introduce?

Bug fix — overrides `getDefaultVideoPoster()` on Android to return a transparent bitmap instead of null, preventing Chromium from painting its default gray play-button placeholder on `<video>` elements.

Closes #1802

## Why is this change needed?

Android's `WebChromeClient.getDefaultVideoPoster()` returns `null` by default. When a `<video>` element has no poster attribute, Chromium paints a stretched gray play-button placeholder bitmap in its place. This is visually undesirable in apps that use video elements without poster images.

## What does this PR implement?

Overrides `getDefaultVideoPoster()` in `RustWebChromeClient` to return a transparent `ARGB_8888` bitmap (`Bitmap.createBitmap(50, 50, Bitmap.Config.ARGB_8888)`), matching the approach used by [react-native-webview](https://github.com/react-native-webview/react-native-webview/blob/58daac9e2e532ca1dcb49003b86568218b9b2b1d/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.kt).

## Verification

- `cargo test` passes
- Verified on Android device — gray placeholder no longer appears on `<video>` elements

### DIFF
--- a/.changes/video-poster-placeholder.md
+++ b/.changes/video-poster-placeholder.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
+Override `getDefaultVideoPoster()` on Android to return a transparent bitmap instead of null, preventing Chromium from painting its default gray play-button placeholder on `<video>` elements.
--- a/src/android/kotlin/RustWebChromeClient.kt
+++ b/src/android/kotlin/RustWebChromeClient.kt
@@ -14,6 +14,7 @@ import android.app.AlertDialog
 import android.content.ActivityNotFoundException
 import android.content.DialogInterface
 import android.content.Intent
+import android.graphics.Bitmap
 import android.net.Uri
 import android.os.Build
 import android.os.Environment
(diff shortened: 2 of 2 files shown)


FabianLars (approved):
that's kinda crazy ngl. can't test this rn but lgtm 🤷

velocitysystems:
@FabianLars One of Android's crazy quirks. Verified on a device, and everything looks good.

PR #1806, opened 2026-08-08
fix: avoid ipc global lexical collisions

### LINKED ISSUE #11612: [bug] Error loading docs.google.com in a webview
### Describe the bug

I'm trying to use the [Google Docs creation page](https://docs.google.com/document/u/0/create) within a Tauri webview. While the Google login challenge completes successfully, the Google Docs page fails to load afterward, showing this error message:

image

We believe the issue might be due to Google Docs attempting to download and install an extension to enable collaborative editing which may not be permitted by default in Tauri's webview environment.

Is there a configuration or workaround we can add to `tauri::WebviewBuilder` to address this loading issue?

### Reproduction

[This repository](https://github.com/sebhernandezr/googledocs-tauri) contains a basic Tauri app that opens a single webview pointing to `https://docs.google.com/document/u/0/create`. Upon launching the app, you’ll be prompted to log in with your Google account. After logging in, the webview attempts to load the Google Docs editor but fails to fully display the page.

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 14.6.0 arm64 (X64)
    ✔ rustc: 1.82.0 (f6e511eec 2024-10-15)
    ✔ cargo: 1.82.0 (8f40fc59f 2024-08-21)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 20.13.1
    - pnpm: 9.1.2
    - yarn: 1.22.22
    - npm: 10.5.2

[-] Packages
    - tauri 🦀: 2.0.6
    - tauri-build 🦀: 2.0.2
    - wry 🦀: 0.46.3
    - tao 🦀: 0.30.3
    - tauri-cli 🦀: 2.0.2
    - @tauri-apps/api : not installed!
    - @tauri-apps/cli : 2.0.2 (outdated, latest: 2.0.4)

[-] Plugins
    - tauri-plugin-os 🦀: 2.0.1
    - @tauri-apps/plugin-os : not installed!
    - tauri-plugin-store 🦀: 2.0.1
    - @tauri-apps/plugin-store : not installed!
    - tauri-plugin-updater 🦀: 2.0.2
    - @tauri-apps/plugin-updater : not installed!

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../../dist/apps/exam-web
    - framework: React
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

[Discord post](https://discord.com/channels/616186924390023171/1304008335883763712)

JasonColapietro:
## Summary

- make Wry's injected `window.ipc` property configurable on WKWebView, WebView2, and WebKitGTK
- preserve the existing frozen `window.ipc.postMessage` bridge for API and version-skew compatibility
- add the required patch release note

## Root cause

Wry currently installs `window.ipc` as a non-configurable own property. ECMAScript global declaration instantiation then rejects a page script that declares a global lexical binding with the same name, such as `const ipc`, before that script can execute. Google Workspace bundles hit this collision and fail with an `ipc` redeclaration syntax error.

Making the property configurable removes that declaration conflict without renaming or removing `window.ipc`. Existing Wry callers and Tauri's v2 postMessage fallback continue to use the same frozen bridge, including when `tauri` and `tauri-runtime-wry` are updated independently.

Closes tauri-apps/tauri#11612.

## Validation

- `cargo fmt --all -- --check`
- `cargo test --verbose` (8 unit tests and 9 doc tests passed; 2 doc tests ignored)
- `cargo clippy --all-targets`
- `git diff --check`
- JavaScriptCore regression check: the current descriptor fails with `Can't create duplicate variable that shadows a global property: 'ipc'`; the configurable descriptor accepts `const ipc` and leaves `globalThis.ipc.postMessage` callable
- Node/V8 regression check for the same declaration and bridge behavior

### DIFF
--- a/.changes/configurable-ipc-global.md
+++ b/.changes/configurable-ipc-global.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
+Make the injected `window.ipc` property configurable so websites can declare their own global `ipc` binding without a syntax error while preserving the existing Wry IPC API.
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -345,7 +345,8 @@ impl InnerWebView {
     };
 
     // Initialize message handler
-    w.init("Object.defineProperty(window, 'ipc', { value: Object.freeze({ postMessage: function(x) { window.webkit.messageHandlers['ipc'].postMessage(x) } }) })", true)?;
+    // Keep the bridge configurable so it does not block a page's global lexical `ipc` binding.
+    w.init("Object.defineProperty(window, 'ipc', { configurable: true, value: Object.freeze({ postMessage: function(x) { window.webkit.messageHandlers['ipc'].postMessage(x) } }) })", true)?;
 
     // Initialize scripts
     for init_script in attributes.initialization_scripts {
--- a/src/webview2/mod.rs
+++ b/src/webview2/mod.rs
@@ -944,10 +944,11 @@ impl InnerWebView {
     attributes: &mut WebViewAttributes,
     token: &mut EventRegistrationToken,
   ) -> Result<()> {
+    // Keep the bridge configurable so it does not block a page's global lexical `ipc` binding.
     Self::add_script_to_execute_on_document_created(
       webview,
       String::from(
-        r#"Object.defineProperty(window, 'ipc', { value: Object.freeze({ postMessage: s=> window.chrome.webview.postMessage(s) }) });"#,
+        r#"Object.defineProperty(window, 'ipc', { configurable: true, value: Object.freeze({ postMessage: s=> window.chrome.webview.postMessage(s) }) });"#,
       ),
     )?;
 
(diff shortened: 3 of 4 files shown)


FabianLars:
This makes no sense. Not only would this silently break the IPC if it's overwritten it would also be an obvious security issue.

I'd much rather rename it to something more specific like `window.wryIpc` in time for tauri v3.

JasonColapietro:
"Makes no sense" is a strange call for a patch with a reproduced Workspace collision, regression checks in two engines, and a bridge that survives init. The descriptor's the real problem: configurable: true makes window.ipc hijackable, so wryIpc in v3 wins. The repro's here whenever that lands. And slop doesn't come with a validation section.

PR #1816, opened 2026-08-17
perf: only register `window.ipc` if IPC handler is set

### LINKED ISSUE #1813: webview2: attach_ipc_handler blocks the UI thread ~100 ms injecting the ipc bridge even when no ipc_handler is set
## Describe the bug

On Windows, `InnerWebView::attach_ipc_handler` unconditionally injects the `window.ipc` bridge script via `Self::add_script_to_execute_on_document_created`, which waits synchronously (`wait_for_async_operation`) for the cross-process `AddScriptToExecuteOnDocumentCreated` completion — even when `attributes.ipc_handler` is `None`:

https://github.com/tauri-apps/wry/blob/dev/src/webview2/mod.rs (see `attach_ipc_handler`, the `add_script_to_execute_on_document_created` call before `ipc_handler.take()`)

Because this runs right after controller creation — while the renderer process is still booting — the round-trip is expensive. Phase-instrumented on wry 0.56.1 (timestamps around each step of `new_in_hwnd`, release build, Windows 11, WebView2 Evergreen 151, warm profile): `attach_ipc_handler` takes **96–109 ms** on the UI thread across samples, while every other settings/handler registration step is ~0.3 ms combined.

For embedders that never use `window.ipc`, that is ~100 ms of unnecessary UI-thread blocking during every webview build. (For paint-bound hello-window benchmarks it largely overlaps renderer startup, so it may not show up in time-to-first-paint — but it does block the UI thread and delays anything else the embedder wants to do after `build()`.)

## Possible fixes

- Skip the bridge injection when `attributes.ipc_handler` is `None` (matches the docs' contract that `window.ipc.postMessage` exists for the ipc handler), or
- inject it with the async completion-handler form instead of `wait_for_async_operation`, so the wait leaves the UI thread.

Happy to provide the raw phase timestamps if useful.

## Environment

- wry 0.56.1, tao 0.35.3
- Windows 11 (10.0.26200), WebView2 Evergreen 151.0.4129.86
- release build, warm user-data folder

Legend-Master:
Closes #1813

I'm not sure if this can be done for Android or simply not worth the time if it doesn't take much time anyways.

### DIFF
--- a/.changes/no-window-ipc.if-no-handler.md
+++ b/.changes/no-window-ipc.if-no-handler.md
@@ -1,0 +1,5 @@
+---
+wry: patch
+---
+
+Only inject `window.ipc` if the webview is built with `WebViewBuilder::with_ipc_handler`
--- a/src/webkitgtk/mod.rs
+++ b/src/webkitgtk/mod.rs
@@ -311,7 +311,9 @@ impl InnerWebView {
     Self::attach_handlers(&webview, web_context, &mut attributes);
 
     // IPC handler
-    Self::attach_ipc_handler(webview.clone(), &mut attributes);
+    if let Some(ipc_handler) = attributes.ipc_handler.take() {
+      Self::attach_ipc_handler(webview.clone(), ipc_handler);
+    }
 
     // Drag drop handler
     if let Some(drag_drop_handler) = attributes.drag_drop_handler.take() {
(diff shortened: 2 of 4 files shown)


FabianLars (approved):
untested

PR #1818, opened 2026-08-18
feat(android): launch new task on `startActivity`

Legend-Master:
Reference https://github.com/tauri-apps/tauri/issues/15857

image

This will also require us to remove the `android:launchMode="singleTask"` from our cli template, but I still don't fully understand why it was added in the first place (oh, maybe it's so sharing to the app don't create new activities? I wonder how this should/can be done now...)


Legend-Master:
cc @FabianLars @lucasfernog do you think this is acceptable change or should we put it behind another tao platform specific flag? (or if you don't want to see this at all)

PR #1820, opened 2026-08-21
fix(macos): domain-match parent cookies in cookies_for_url

### LINKED ISSUE #1749: macOS cookies_for_url appears to miss domain cookies for subdomains
### Describe the bug

On macOS/WKWebView, `WebView::cookies_for_url(url)` appears to filter cookies by exact domain equality. This can omit valid domain cookies when the request URL is a subdomain.

For example, if the cookie store contains a secure HTTP-only cookie scoped to `example.com`, then querying cookies for `https://www.example.com/path` can return no such cookie, even though a domain cookie for `example.com` should be sent to `www.example.com` according to RFC 6265 domain-match rules.

The current macOS implementation in Wry 0.54.2 and 0.55.1 appears to use exact equality:

```rust
cookie.domain() == url.domain()
```

That means a cookie whose domain is `example.com` will not be returned for `https://www.example.com/`.

### Reproduction

I observed this through a desktop app using Wry-backed WebViews and a shared persistent cookie store. Details below are anonymized.

1. Load `https://www.example.com/some/path` in a WKWebView.
2. The site sets or updates cookies during a browser challenge / normal navigation.
3. Inspecting cookies by URL shows:
   - `cookies_for_url("https://www.example.com/some/path")` returns regular cookies scoped to `www.example.com`.
   - `cookies_for_url("https://example.com/")` returns a secure HTTP-only clearance/session cookie scoped to `example.com`.
4. A non-WebView HTTP client using `cookies_for_url("https://www.example.com/some/path")` therefore does not receive the `example.com` cookie and the request fails authentication/challenge checks.

Anonymized shape of the cookie store:

```text
Query URL: https://www.example.com/some/path
Returned cookies: [regular_cookie(domain=www.example.com, path=/)]

Query URL: https://example.com/
Returned cookies: [clearance_cookie(domain=example.com, path=/, secure=true, httpOnly=true)]
```

### Expected behavior

For domain cookies, `cookies_for_url("https://www.example.com/path")` should include cookies scoped to `example.com` when the request host domain-matches the cookie domain.

For host-only cookies, exact host matching should still be preserved.

In RFC 6265 terms, retrieval should distinguish:

- host-only cookie: `request-host == cookie.domain`
- domain cookie: request host domain-matches `cookie.domain`

### Actual behavior

On macOS/WKWebView, the current exact equality check appears to exclude the parent-domain cookie for subdomain URLs.

### Notes

I realize this may need care because `cookie::Cookie::domain()` does not by itself expose whether the original cookie was host-only or set with a `Domain` attribute. A simple `ends_with` check would be unsafe for host-only cookies.

Could Wry preserve enough host-only/domain-cookie metadata from WKHTTPCookie/NSHTTPCookie, or delegate URL filtering to platform cookie APIs where possible, so `cookies_for_url` follows browser cookie retrieval semantics?

### Environment

- Wry versions checked: 0.54.2 and 0.55.1
- Platform: macOS / WKWebView backend
- API: `WebView::cookies_for_url`

### Relevant source

In `src/wkwebview/mod.rs`:

```rust
pub fn cookies_for_url(&self, url: &str) -> Result<Vec<cookie::Cookie<'static>>> {
  let url = url::Url::parse(url)?;

  self.cookies().map(|cookies| {
    cookies.into_iter().filter(|cookie: &cookie::Cookie| {
      let secure = cookie.secure().unwrap_or_default();
      cookie.domain() == url.domain()
        && (/* secure checks */)
    }).collect()
  })
}
```

richardmilles:
Use RFC 6265 domain-match in macOS cookies_for_url so parent Domain cookies (e.g. example.com) are returned for subdomain URLs like www.example.com.

Fixes #1749

### DIFF
--- a/.changes/macos-cookies-domain-match.md
+++ b/.changes/macos-cookies-domain-match.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Fix macOS `cookies_for_url` to include parent Domain cookies for subdomain URLs via RFC 6265 domain-match.
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -1201,8 +1201,8 @@ impl InnerWebView {
     self.cookies().map(|cookies| {
       cookies.into_iter().filter(|cookie: &cookie::Cookie| {
         let secure = cookie.secure().unwrap_or_default();
-        // domain is the same
-        cookie.domain() == url.domain()
+        // domain-match (includes parent Domain cookies for subdomain URLs)
+        cookie_domain_matches(cookie.domain(), url.domain())
           // and one of
           && (
             // cookie is secure and url is https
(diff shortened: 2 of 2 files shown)


JJJ:
I tested the current patch on macOS, and I think the host-only cookie handling needs one correction before merge. `NSHTTPCookie` preserves the leading dot in `domain`, and Apple documents that a domain without a leading dot is restricted to the exact host, while a dotted domain may include subdomains: https://developer.apple.com/documentation/foundation/httpcookie/domain?changes=_7&language=objc

Stripping the dot before matching loses that distinction, so a host-only `example.com` cookie currently becomes visible to `www.example.com`. The suffix match should only run when the original cookie domain begins with `.`. The tests should expect:

- `example.com` → `www.example.com`: false
- `.example.com` → `www.example.com`: true
- exact host matches: true

This keeps the intended parent-domain fix without widening host-only cookie scope.

JJJ:
I opened #1851 as an updated version of this change. It keeps @richardmilles's original commit and attribution, then adds the host-only cookie-scope correction discussed above. The full `cargo test --all-features` suite passes on macOS.

PR #1824, opened 2026-08-26
fix(ios): follow the system background when no color is set

IntellyCode:
On iOS a `WebView` created without a `background_color` keeps UIKit's default: opaque and white. During the gap between the launch screen dismissing and the page painting its first frame, that white shows through as a flash — noticeable in dark mode, where the app is dark on both sides of it.

The existing `background_color` path already sets `setOpaque(false)` and paints the given colour to avoid exactly this (see the `Some` branch). The `None` case was left untouched, so it still flashes.

This adds an `else` branch: when no colour is given, set the webview non-opaque and fall back to `UIColor.systemBackgroundColor`, which follows the current light/dark appearance. Apps that pass a `background_color` are unaffected.

Requires the `UIInterface` feature of `objc2-ui-kit` for `systemBackgroundColor`, added to `Cargo.toml`.

Tested on a device (iOS, dark and light): the launch flash is gone in both.

### DIFF
--- a/src/wkwebview/mod.rs
+++ b/src/wkwebview/mod.rs
@@ -466,6 +466,15 @@ impl InnerWebView {
           // This has to be monitored as it may clash with isOpaque = true.
           // The webview background color may also applied too late so actually not that useful.
           webview.setBackgroundColor(Some(&color));
+        } else {
+          // Without a color the webview stays opaque and white, flashing before the page paints.
+          // Fall back to the system background so it follows the current appearance instead.
+          webview.setOpaque(false);
+          let color = objc2_ui_kit::UIColor::systemBackgroundColor();
+          if !is_child {
+            ns_view.setBackgroundColor(Some(&color));
+          }
+          webview.setBackgroundColor(Some(&color));
         }
         webview
       };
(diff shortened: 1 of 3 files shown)


src/wkwebview/mod.rs
```diff
@@ -466,6 +466,15 @@ impl InnerWebView {
...
           webview.setBackgroundColor(Some(&color));
+        } else {
+          // Without a color the webview stays opaque and white, flashing before the page paints.
+          // Fall back to the system background so it follows the current appearance instead.
+          webview.setOpaque(false);
+          let color = objc2_ui_kit::UIColor::systemBackgroundColor();
+          if !is_child {
+            ns_view.setBackgroundColor(Some(&color));
+          }
+          webview.setBackgroundColor(Some(&color));
```
FabianLars: most of this is duplicated code, how about merging both if/else branches by branching on the color variable instead? eg
```rs
let color = if let Some((red, green, blue, alpha)) = attributes.background_color {
  objc2_ui_kit::UIColor::colorWithRed_green_blue_alpha(
    red as f64 / 255.0,
    green as f64 / 255.0,
    blue as f64 / 255.0,
    alpha as f64 / 255.0,
  )
} else {
  objc2_ui_kit::UIColor::systemBackgroundColor()
};
webview.setOpaque(false);
if !is_child {
  ns_view.setBackgroundColor(Some(&color));
}
webview.setBackgroundColor(Some(&color));
```

IntellyCode:
You are absolutely right. I did not notice that duplication. Made that into a new commit.

FabianLars (approved):

FabianLars:
ah sorry but you'll need to sign your commits for me to merge this :/

IntellyCode:
Okay, signed.

PR #1829, opened 2026-09-03
fix(macos): forward non-file drags to WebKit

thatjuan:
## Problem

When Wry's native drag-drop handler accepts an in-page drag on macOS, `WryWebView` does not call its `WKWebView` superclass. HTML5 `dragenter`, `dragover`, and `drop` events therefore never reach the page while native file drag-and-drop is enabled. This prevents apps from supporting both native path-based file drops and in-page HTML5 drag-and-drop. See tauri-apps/tauri#14373.

## Fix

Check the dragging pasteboard for the standard `NSPasteboardTypeFileURL` type before invoking Wry's native file-drop handler. File drags keep the existing Wry behavior. Non-file drags are forwarded to WebKit in `draggingEntered`, `draggingUpdated`, `performDragOperation`, and `draggingExited`.

## Verification

- `cargo fmt --all -- --check`
- `cargo build` on aarch64 macOS

### DIFF
--- a/.changes/macos-forward-non-file-drags.md
+++ b/.changes/macos-forward-non-file-drags.md
@@ -1,0 +1,5 @@
+---
+"wry": patch
+---
+
+Forward non-file drag-and-drop events to WebKit on macOS so HTML5 drag-and-drop can coexist with Wry's native file drop handler.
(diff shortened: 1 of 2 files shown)


JJJ:
I tested this locally on macOS. `cargo test --all-features` passes, and a direct AppKit pasteboard probe confirmed that a legacy `NSFilenamesPboardType` file drag is also advertised as `public.file-url`, while a string/HTML drag is not. The file/non-file split here looks correct to me.

One coordination note: #1844 also touches `src/wkwebview/drag_drop.rs`, so whichever PR lands second may need a rebase.

PR #1839, opened 2026-09-12
fix(webkitgtk): only inject IPC bridge with handler

zackees:
The WebKitGTK backend conditionally registered the native IPC handler, but unconditionally injected the browser-facing `window.ipc` bridge. This leaves a browser-visible host bridge on webviews that intentionally configure no IPC handler.

This makes script injection conditional on the same IPC-handler configuration and adds a focused regression test.

Coordinated with zackees/kernal-api#18 and tauri-apps/tauri#16010.


src/webkitgtk/mod.rs
```diff
@@ -63,6 +63,11 @@ use crate::{
 use self::web_context::WebContextExt;
 
 const WEBVIEW_ID: &str = "webview_id";
+const IPC_INIT_SCRIPT: &str = "Object.defineProperty(window, 'ipc', { value: Object.freeze({ postMessage: function(x) { window.webkit.messageHandlers['ipc'].postMessage(x) } }) })";
+
+fn ipc_init_script(ipc_enabled: bool) -> Option<&'static str> {
+  ipc_enabled.then_some(IPC_INIT_SCRIPT)
+}
```
Legend-Master: This is much more added complexity than needed, let's keep it simple by removing this and the tests. Just keep the `ipc_handler.is_some` check is enough.

Legend-Master (commented):
Thanks, nice catch!

zackees:
It's been signed

zackees:
Thanks for the review! Both requests are addressed:

- Removed the `IPC_INIT_SCRIPT` const, `ipc_init_script()` helper, and the tests — it's now just the `ipc_handler.is_some()` check (763bd6c).
- All commits on the branch are signed.

PR #1846, opened 2026-09-15
fix(android): don't abort when the activity is already gone

### LINKED ISSUE #1845: Android: get_webview aborts the process with SIGABRT after the activity is destroyed
## Bug

`get_webview` in `src/android/main_pipe.rs` unwraps the `ACTIVITY_PROXY` lookup:

```rust
pub fn get_webview(activity_id: ActivityId) -> Option<GlobalRef> {
  ACTIVITY_PROXY
    .lock()
    .unwrap()
    .get(&activity_id)
    .unwrap()
    .webview
    .as_ref()
    .cloned()
}
```

`WebViewMessage::OnDestroy` removes the activity's entry from `ACTIVITY_PROXY` whenever Android destroys that activity — the system back gesture being the ordinary case — while the host process keeps running. Any message handled in the window after `OnDestroy` and before the process actually exits looks the id up, finds nothing, and `.unwrap()` aborts the whole process with `SIGABRT` instead of returning `None`.

The function already returns `Option<GlobalRef>`, and all nine call sites in the crate are already written defensively as `if let Some(webview) = get_webview(activity_id) { ... }` — `None` is the case they were built to handle. The `.unwrap()` inside `get_webview` itself is the one place that denies it to them.

## Repro / evidence

Measured on a Galaxy S23 FE (Android 14) running a Tauri 2 app that polls a webview's cookies on a 5s timer via `cookies_for_url` (the same window this crate's `#1837` / `#1838` cover):

1. Sign in / open a secondary webview.
2. Press the system Back button to destroy the activity while the app process stays alive (e.g. another activity or the main activity is still running).
3. ~26 seconds later — the next timer tick that lands after `OnDestroy` but before the process actually goes away — the app crashes.

Logcat shows a panic at `main_pipe.rs:121` immediately followed by the process dying on signal 6:

```
panicked at .../main_pipe.rs:121
... SIGABRT
```

Intermittent rather than every time, since it depends on a message landing in that specific window between `OnDestroy` and process exit.

## Fix

```diff
 pub fn get_webview(activity_id: ActivityId) -> Option<GlobalRef> {
   ACTIVITY_PROXY
     .lock()
     .unwrap()
-    .get(&activity_id)
-    .unwrap()
+    .get(&activity_id)?
     .webview
     .as_ref()
     .cloned()
 }
```

PR incoming with this change against `dev`.

## Environment

- wry: `dev` (also present in the `0.55.1`/`0.57.0` lines used by tauri 2.11.x)
- Platform: Android

bbmediaco:
Fixes #1845.

`get_webview` unwrapped the `ACTIVITY_PROXY` lookup, but `OnDestroy` removes that entry whenever Android destroys the activity — the back gesture being the ordinary case — while the process keeps running. A message handled in that window finds no entry and the unwrap aborts the process with `SIGABRT`.

Measured on a Galaxy S23 FE: a consumer that harvests cookies on a timer (`cookies_for_url`) crashed ~26s after the back gesture, panicking at `main_pipe.rs:121` and dying on signal 6. Intermittent rather than every time — the trigger is a message landing after `OnDestroy` and before the process goes.

The function already returns `Option`, and all nine call sites are already written `if let Some(webview) = get_webview(activity_id)`, so `None` is the case they were built for; the unwrap was what denied it to them.

Same shape as #1838 — a one-line fix for a null/missing case the crate's own call sites already expect.


FabianLars:
No need for the slop doc comment, should be very clear to everyone not to use unwrap. You can remove all of that comment again really.

If you instead add a change file we can merge it :)

